PR #48, opened 2016-02-09
Round notch corners

### LINKED ISSUE #40: Horizontal Grammar - Notch to spec
The notch as currently rendered does not match the spec from @carljbowman 
#### Current

image
#### Spec

image

rschamp:
Resolves #40

The rounding adds some height to the notch, so this is compensated by removing height from the notch "tab". For this reason the radius should stay relatively small.

### DIFF
--- a/core/block_horizontal_scratch.js
+++ b/core/block_horizontal_scratch.js
@@ -819,44 +819,37 @@ Blockly.BlockSvg.NOTCH_HEIGHT = Blockly.BlockSvg.NOTCH_BASE_HEIGHT + Blockly.Blo
  * @const
  */
 Blockly.BlockSvg.NOTCH_PATH_DOWN = 
-  'a ' + Blockly.BlockSvg.NOTCH_RADIUS + ',' + 2*Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
-          '-45 0 0 ' + Blockly.BlockSvg.NOTCH_RADIUS/2 + ',' + Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
+  'a ' + -Blockly.BlockSvg.NOTCH_RADIUS + ',' + Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
+          '0 0 0 ' + Blockly.BlockSvg.NOTCH_RADIUS/2 + ',' + Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
   'l ' + (Blockly.BlockSvg.NOTCH_BASE_HEIGHT/4 - Blockly.BlockSvg.NOTCH_RADIUS) + ',' + 
          (Blockly.BlockSvg.NOTCH_BASE_HEIGHT/4 - Blockly.BlockSvg.NOTCH_RADIUS) + ' ' +
-  'a ' + Blockly.BlockSvg.NOTCH_RADIUS + ',' + 2*Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
-          '-45 0 1 ' + Blockly.BlockSvg.NOTCH_RADIUS/2 + ',' + Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
+  'a ' + -Blockly.BlockSvg.NOTCH_RADIUS + ',' + Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
+          '0 0 1 ' + Blockly.BlockSvg.NOTCH_RADIUS/2 + ',' + Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
   'v ' + (Blockly.BlockSvg.NOTCH_BASE_HEIGHT/2 - Blockly.BlockSvg.NOTCH_RADIUS) + ' ' +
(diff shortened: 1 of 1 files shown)


core/block_horizontal_scratch.js
```diff
@@ -793,22 +793,54 @@ Blockly.BlockSvg.TAB_WIDTH = 8;
  * Width of vertical tab (inc left margin).
```
tmickel: maybe let's update this comment as well? :) would be worth pointing out what is notch_base_height vs. notch_height

core/block_horizontal_scratch.js
```diff
@@ -793,22 +793,54 @@ Blockly.BlockSvg.TAB_WIDTH = 8;
...
  * @const
  */
 Blockly.BlockSvg.CORNER_RADIUS = 4;
 /**
- * SVG path for drawing next/previous notch from left to right.
+ * Rounded notch radius.
  * @const
  */
-Blockly.BlockSvg.NOTCH_PATH_DOWN = 'l 8,8 0,16 -8,8';
+Blockly.BlockSvg.NOTCH_RADIUS = 1;
```
tmickel: I noticed in my test that the rounding is not really the spec in #40 - it looks closer to me around NOTCH_RADIUS = 2.5. I'm assuming you're working this out with Carl

core/block_horizontal_scratch.js
```diff
@@ -793,22 +793,54 @@ Blockly.BlockSvg.TAB_WIDTH = 8;
...
+Blockly.BlockSvg.NOTCH_HEIGHT = Blockly.BlockSvg.NOTCH_BASE_HEIGHT + Blockly.BlockSvg.NOTCH_RADIUS;
```
tmickel, line 816: also just wondering, is this an approximation or is it the correct height of NOTCH_PATH_DOWN and NOTCH_PATH_UP?

core/block_horizontal_scratch.js
```diff
@@ -793,22 +793,54 @@ Blockly.BlockSvg.TAB_WIDTH = 8;
...
+         (Blockly.BlockSvg.NOTCH_BASE_HEIGHT/4 - Blockly.BlockSvg.NOTCH_RADIUS) + ' ' +
+  'a ' + Blockly.BlockSvg.NOTCH_RADIUS + ',' + 2*Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
+          '-45 0 1 ' + Blockly.BlockSvg.NOTCH_RADIUS/2 + ',' + Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
+  'v ' + (Blockly.BlockSvg.NOTCH_BASE_HEIGHT/2 - Blockly.BlockSvg.NOTCH_RADIUS) + ' ' +
+  'a ' + Blockly.BlockSvg.NOTCH_RADIUS + ',' + 2*Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
+          '45 0 1 ' + -1*Blockly.BlockSvg.NOTCH_RADIUS/2 + ',' + Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
+  'l ' + (-1*Blockly.BlockSvg.NOTCH_BASE_HEIGHT/4 + Blockly.BlockSvg.NOTCH_RADIUS) + ',' + 
+         (1*Blockly.BlockSvg.NOTCH_BASE_HEIGHT/4 - Blockly.BlockSvg.NOTCH_RADIUS) + ' ' +
+  'a ' + Blockly.BlockSvg.NOTCH_RADIUS + ',' + 2*Blockly.BlockSvg.NOTCH_RADIUS + ' ' +
+        '45 0 0 ' + -Blockly.BlockSvg.NOTCH_RADIUS/2 + ',' + Blockly.BlockSvg.NOTCH_RADIUS;
```
tmickel: this looks good to me, especially if we never have to touch it again :) I wouldn't mind also separating out the radius for the top and bottom curves vs. the notch curves (and also x radius vs. y radius) rather than just having a single NOTCH_RADIUS and doing a bunch of division/math. That would probably make it easier to adjust to exactly what we want. But I'm thinking we won't have to adjust it very much, so I'd say just leave it...

core/block_horizontal_scratch.js
rschamp, line 816: This is the correct height of the notch. The dimensions of the components of the notch use `NOTCH_BASE_HEIGHT` because they shouldn't take the radius leading into the notch into account, but everything else that needs to know the notch height should include the radius.

core/block_horizontal_scratch.js
rschamp: Ah whoops, thanks.

core/block_horizontal_scratch.js
rschamp: Looking at the comp again it seems like this math shouldn't even be necessary since those curves are circular not elliptical like mine. I am going to take another pass so that we don't need separate x/y radii so this doesn't look so insane.

core/block_horizontal_scratch.js
rschamp: You're right. Updated.

PR #61, opened 2016-02-18
Merge latest blockly master

rschamp:
Brings in the new events system.  I renamed core/block_horizontal_scratch.js to core/block_svg.js so that in the future, changes to block_svg.js will be applied directly to the horizontal file rather than us needing to manually merge them in.

To resolve the compressed files, I rebuilt and used the resulting files.

I would appreciate if @NeilFraser and @rachel-fenichel could look this over in case there is something glaringly incorrect.  The playground seemed to work ok.


core/block_vertical_scratch.js
```diff
@@ -24,7 +24,7 @@
  */
 'use strict';
 
-goog.provide('Blockly.BlockSvg');
+// goog.provide('Blockly.BlockSvg');
```
rachel-fenichel: For build.py to work this needs to not be commented out.

You also need to change build.py to exclude block_svg.js in the vertical case.  It currently excludes block_horizontal_scratch.js.

I also think that in general the non-rendering parts of block_svg should move into another file--I think we still have a lot duplicated between horizontal and vertical.  That's probably a problem for a different pr though.

PR #67, opened 2016-02-19
Additional playground with multiple instances of blockly

rachel-fenichel:
For at-a-glance testing of basic rendering in various combinations of modes.

Eventually it would be excellent for this to use events to reflect changes in one playground to all others, though that runs the risk of making all failures catastrophic.


core/toolbox.js
```diff
@@ -78,12 +78,42 @@ Blockly.Toolbox = function(workspace) {
...
+    cssTreeRow: 'blocklyTreeSeparator'
+  };
+
   if (this.horizontalLayout_) {
     this.CONFIG_['cssTreeRow'] =
         this.CONFIG_['cssTreeRow'] +
         (workspace.RTL ? ' blocklyHorizontalTreeRtl' : ' blocklyHorizontalTree');
 
-    Blockly.Toolbox.TreeSeparator.CONFIG_['cssTreeRow'] =
+    this.TREE_SEPARATOR_CONFIG_['cssTreeRow'] =
```
rschamp: Will anything complain later if this `@const` is modified here?

rachel-fenichel: No, but I shouldn't leave misleading annotations.  Removed, and names changed to camel case to indicate that they aren't constant anymore.

PR #79, opened 2016-02-23
Match string and number inputs to spec

rschamp:
_Preview:_
image

Also adds a "Say" block to demo string inputs.

I couldn't see a good way to override the looks of inputs just for horizontal blocks so this breaks the vertical style quite a bit.  Suggestions welcome...


blocks_horizontal/looks.js
```diff
@@ -0,0 +1,65 @@
...
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "id": "looks_say",
+      "message0": "%1 %2",
+      "args0": [
+        {
+          "type": "field_image",
+          "src": "../media/icons/looks_say.svg",
```
tmickel: let's make this relative as well!

rschamp: Oh, right — thanks!

core/field.js
```diff
@@ -76,7 +78,7 @@ Blockly.Field.prototype.name = undefined;
  * Maximum characters of text to display before adding an ellipsis.
  * @type {number}
  */
-Blockly.Field.prototype.maxDisplayLength = 50;
+Blockly.Field.prototype.maxDisplayLength = 5;
```
tmickel, line 81: Someday we should think about where to put this. Should it be defined with the rest of the block?

core/field.js
```diff
@@ -137,10 +139,14 @@ Blockly.Field.prototype.init = function(block) {
...
-       'height': 16}, this.fieldGroup_, this.sourceBlock_.workspace);
+       'height': Blockly.BlockSvg.FIELD_HEIGHT}, this.fieldGroup_, this.sourceBlock_.workspace);
   /** @type {!Element} */
   this.textElement_ = Blockly.createSvgElement('text',
-      {'class': 'blocklyText', 'y': this.size_.height - 12.5},
+      {'class': 'blocklyText',
+       'y': this.size_.height/2 + 6.25,
+       'x': Blockly.BlockSvg.FIELD_WIDTH / 2,
+       'width': Blockly.BlockSvg.FIELD_WIDTH - Blockly.BlockSvg.SEP_SPACE_X,
+       'text-anchor': 'middle'},
```
tmickel, line 149: and this section I'm guessing we'll have to reconsider for vertical

PR #91, opened 2016-02-26
Add spec tool playground

tmickel:
This playground has:
-Wire grid on by default and sized to the correct px to match Carl's grid
-Partially transparent workspace overlaid on a background image (i.e. for Carl's specs)
-Default zoom for precision working on paths

@thisandagain @rschamp anything else that would be useful here?


tests/spectool.html
```diff
@@ -0,0 +1,223 @@
...
+  if (text) {
+    document.getElementById('importExport').value = text;
+  }
+  taChange();
+  // Restore event logging state.
+  var state = sessionStorage.getItem('logEvents');
+  logEvents(Boolean(state));
+}
+
+function toXml() {
```
rachel-fenichel: Do you need the textarea, toXml, event logging, etc?

tmickel: Definitely not :P I went ahead and simplified it.

Issue #75, opened 2016-02-22
Horizontal Grammar - Control Loop - Forever


carljbowman:
## Illustrator File

[R4_Loop Forever.ai.zip](https://github.com/LLK/scratch-blockly/files/140835/R4_Loop.Forever.ai.zip)
## General Size / Icon

![r4_loop-forever size](https://cloud.githubusercontent.com/assets/3409578/13222427/d966a5ac-d94d-11e5-9263-e1919262d110.png)
## Radii

![r4_loop-forever radii](https://cloud.githubusercontent.com/assets/3409578/13222437/df8559b0-d94d-11e5-9f6e-bbde4eca55b5.png)

Issue #74, opened 2016-02-22
Horizontal Grammar - Control Loop with Int Input - Repeat


carljbowman:
## Illustrator File

[R4_Loop Int Input Block.ai.zip](https://github.com/LLK/scratch-blockly/files/141342/R4_Loop.Int.Input.Block.ai.zip)
## General Size / Icon

![r4_loop-int-input-block size](https://cloud.githubusercontent.com/assets/3409578/13222338/8e013280-d94d-11e5-95e7-01a92023f7ce.png)
## Radii

![r4_loop-int-input-block radii](https://cloud.githubusercontent.com/assets/3409578/13230988/b536dfda-d975-11e5-81b2-af298d7d6dfe.png)
## Input Alignement

![r4_loop-int-input-block input](https://cloud.githubusercontent.com/assets/3409578/13222351/9c76d9d2-d94d-11e5-836f-afc1779c4ff6.png)

Issue #72, opened 2016-02-22
Horizontal Grammar - Event Cap - Green Flag


carljbowman:
## Illustrator File

[R4_Event Cap.ai.zip](https://github.com/LLK/scratch-blockly/files/140810/R4_Event.Cap.ai.zip)
## General Size / Icon

![r4_event-cap size](https://cloud.githubusercontent.com/assets/3409578/13222013/05f6329c-d94c-11e5-9935-e496999f31f9.png)
## Radii

![r4_event-cap radii](https://cloud.githubusercontent.com/assets/3409578/13222022/0d40fa82-d94c-11e5-8bbf-d26502db7a0b.png)

Issue #71, opened 2016-02-22
Horizontal Grammar - Standard Block with String Input - Say


carljbowman:
## Illustrator File

[R4_Standard String Input Block.ai.zip](https://github.com/LLK/scratch-blockly/files/140795/R4_Standard.String.Input.Block.ai.zip)
## General Size / Icon

![r4_standard-string-input-block size](https://cloud.githubusercontent.com/assets/3409578/13221848/68c96ae8-d94b-11e5-9254-0a5d0712a1ab.png)
## Radii

![r4_standard-string-input-block radii](https://cloud.githubusercontent.com/assets/3409578/13221857/6dd7e3e8-d94b-11e5-8d45-72dabfce1d50.png)
## Input Alignment

![r4_standard-string-input-block input](https://cloud.githubusercontent.com/assets/3409578/13221890/8ad78106-d94b-11e5-9dc0-595965814f51.png)

PR #99, opened 2016-03-01
Implement block shapes to latest spec

tmickel:
-Convert most constants in the code to use multiplications of GRID_UNIT (4px). This lets us be consistent with Carl's specs and helps explain the numbers a little better.
-Implement notches using cubic curves instead of arcs. This version matches the spec pretty much exactly. As part of this, remove NOTCH_BASE_HEIGHT which is no longer relevant.
-Slight simplification of renderCompute_ (which calculates metrics).
-Fixed SEP_SPACE_X and SEP_SPACE_Y to be the correct number of grid units. Our blocks before were slightly too small, now they are the right size.
-Implement the correct corner radius, hat/end hat radius.
-Add fixed ICON_WIDTH and ICON_HEIGHT, i.e., always size the icons to 40x40px as according to the spec.
-Add fixed STATEMENT_BLOCK_SPACE to represent the size of the bar above forever/repeat/pants blocks. Previously the code was assuming this was SEP_SPACE_Y, when in fact the spec has it as larger (4 units).
-Add accurate MIN_BLOCK_X and MIN_BLOCK_Y for simpler calculation of minimum bay size.
-Remove some unused SVG paths (INNER_TOP_LEFT_CORNER, INNER_BOTTOM_LEFT_CORNER).
-Remove extra code in getHeightWidth which caused a bug in rendering forever-in-repeat.
-Fix stroke width around the blocks to be 1px, according to spec.

I've checked that everything lines up with Carl's spec images at 8x zoom.

This should implement #69, #70, #71, #72, #74, and #75 (not 73 as we haven't done that input yet).

### DIFF
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -397,7 +397,7 @@ Blockly.BlockSvg.prototype.renderCompute_ = function() {
     metrics.width = Blockly.BlockSvg.FIELD_WIDTH;
   } else {
     metrics.height = Math.max(
-      Blockly.BlockSvg.SEP_SPACE_Y * 2 + iconSize.height,
+      Blockly.BlockSvg.SEP_SPACE_Y * 3 + iconSize.height,
       Blockly.BlockSvg.NOTCH_HEIGHT + 16 + Blockly.BlockSvg.CORNER_RADIUS * 2,
       metrics.bayHeight + Blockly.BlockSvg.SEP_SPACE_Y
     );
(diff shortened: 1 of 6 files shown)


core/block_render_svg_horizontal.js
```diff
@@ -536,15 +525,16 @@ Blockly.BlockSvg.prototype.renderDrawBottom_ = function(steps,
...
   // Has statement
   if (metrics.statement) {
-    steps.push('h', 8);
+    steps.push('h', 4 * Blockly.BlockSvg.GRID_UNIT);
     steps.push('a', Blockly.BlockSvg.CORNER_RADIUS + ',' +
                Blockly.BlockSvg.CORNER_RADIUS + ' 0 0,0 ' +
                Blockly.BlockSvg.CORNER_RADIUS + ',-' +
                Blockly.BlockSvg.CORNER_RADIUS);
-    steps.push('v', -8);
+    steps.push('v', -10);
```
rachel-fenichel: Should be a constant.

thisandagain: Agreed. Used on line 550 as well.

tmickel: Addressed!

PR #104, opened 2016-03-01
Colour spec implementation, #98

tmickel:
Creates a centralized place for our colours to live, Blockly.colours. Pull all colours for blocks and interface elements from here.

Extends block colours to include getters and setters for secondary and tertiary colours.

Updates updateColour and css.js to use the new colours.

### DIFF
--- a/core/colours.js
+++ b/core/colours.js
@@ -29,5 +29,6 @@ Blockly.Colours = {
     "tertiary": "#CCAA00"
   },
   "text": "#575E75",
-  "workspace": "#F5F8FF"
+  "workspace": "#F5F8FF",
+  "text_field": "#FFFFFF"
 };
--- a/core/colours.js
+++ b/core/colours.js
@@ -1,0 +1,33 @@
+'use strict';
+
+goog.provide('Blockly.Colours');
+
+Blockly.Colours = {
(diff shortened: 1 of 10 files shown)


core/colours.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+goog.provide('Blockly.Colours');
+
+Blockly.Colours = {
```
thisandagain, line 5: :+1:

core/colours.js
```diff
@@ -0,0 +1,34 @@
...
+    "tertiary": "#CF8B17"
+  },
+  "event": {
+    "primary": "#FFD500",
+    "secondary": "#DBC200",
+    "tertiary": "#CCAA00"
+  },
+  "text": "#575E75",
+  "workspace": "#F5F8FF",
+  "text_field": "#FFFFFF"
```
thisandagain: I can't quite work out the naming conventions within blockly: it's a mix of `camelCase` and `under_score` but it seems like most vars / props in `Colour` follow camel case. Thoughts on the use of `text_field` here?

tmickel: Sounds good to have it consistent within Blockly.Colours at least!

PR #112, opened 2016-03-03
Some RTL fixes #34

### LINKED ISSUE #34: L10N - Fix RTL rendering of blocks
In the process of quickly implementing the horizontal grammar we broke RTL in a number of places:
image

Depends on #31

tmickel:
This is a partial implementation of correct positioning in RTL.

-Fixes a bug with the horizontal workspace positioning of blocks
-Add property to image fields to indicate we should flip the icon
-Positioning of icons correctly
-Positioning of connections correctly

Still to do:
-Positioning of the inner input/field text in the shadow block seems off...not sure what's up with that yet. 
screen shot 2016-03-03 at 5 49 41 pm

### DIFF
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -518,9 +518,10 @@ Blockly.BlockSvg.prototype.renderDrawBottom_ = function(steps,
                Blockly.BlockSvg.CORNER_RADIUS);
 
     // Create statement connection.
-    // @todo RTL
-    // var connectionX = connectionsXY.x + (this.RTL ? -cursorX : cursorX + 1);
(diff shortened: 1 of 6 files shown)


core/block_render_svg_horizontal.js
```diff
@@ -380,21 +380,34 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(metrics) {
     var icon = metrics.icon.getSvgRoot();
     var iconSize = metrics.icon.getSize();
     var iconX = metrics.width - iconSize.width - Blockly.BlockSvg.SEP_SPACE_X / 1.5;
```
rachel-fenichel, line 383: Add comment: // icon's position is calculated relative to the "end" edge of the block.

PR #113, opened 2016-03-04
Glows on running blocks and (partially) stacks

tmickel:
This adds a property to the Block SVG to track "glow" state (since "highlight" is already used in the Blockly codebase). Blocks switch their fill color to the secondary color when the state is set by the workspace. Should allow us to easily implement block-by-block glowing (as in ScratchJr).

I've also added a new SVG filter for "stack glow" more like Scratch.
screen shot 2016-03-04 at 12 24 54 pm
It works if applied manually to the dragging group of the stack. I didn't add setters for this yet as I think we need to consider where the filter might actually be applied (and if we can use it).
For example, because of the way we're likely doing ghosts, when we apply the filter to the dragging group, the filter is also applied to the ghost :( so the insertion ghosts appear highlighted if they're dragged in from the right to a "running" stack. It makes Carl pretty sad, but maybe it's actually consistent/reasonable behavior anyway.

Anyway this PR should get us mostly there.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -464,12 +464,10 @@ Blockly.WorkspaceSvg.prototype.glowBlock = function(id, isGlowing) {
   if (id) {
     block = Blockly.Block.getById(id);
     if (!block) {
-      return;
+      throw 'Tried to glow block that does not exist.';
     }
   }
-  if (block) {
-    block.setGlow(isGlowing);
-  }
+  block.setGlow(isGlowing);
 };
 
 /**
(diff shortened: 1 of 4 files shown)


core/workspace_svg.js
```diff
@@ -455,6 +455,24 @@ Blockly.WorkspaceSvg.prototype.highlightBlock = function(id) {
...
 
 /**
+ * Glow/unglow a block in the workspace.
+ * @param {?string} id ID of block to find.
+ * @param {boolean} isGlowing Whether to glow the block.
+ */
+Blockly.WorkspaceSvg.prototype.glowBlock = function(id, isGlowing) {
+  var block = null;
+  if (id) {
+    block = Blockly.Block.getById(id);
```
rachel-fenichel, line 465: Might be worth throwing if there wasn't a block by that ID.  I think that'll be an illegal state.

tmickel, line 465: Good call, added

Issue #130, opened 2016-03-21
Pants block connections priority/closeness for wrapping behavior


tmickel:
Hard to describe what's going on here :)
Video sent to Rachel.

PR #136, opened 2016-03-28
Don't let c-blocks connect in the middle of a stack

rachel-fenichel:
#130

### DIFF
--- a/core/connection.js
+++ b/core/connection.js
@@ -400,13 +400,6 @@ Blockly.Connection.prototype.isConnectionAllowed = function(candidate,
     return false;
   }
 
-  // Don't offer to connect an already connected left (male) value plug to
-  // an available right (female) value plug.
-  if (candidate.type == Blockly.OUTPUT_VALUE) {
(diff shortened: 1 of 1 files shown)


core/connection.js
```diff
@@ -411,6 +411,28 @@ Blockly.Connection.prototype.isConnectionAllowed = function(candidate,
   var firstStatementConnection =
       this.sourceBlock_.getFirstStatementConnection();
 
+  // Scratch-specific behaviour:
+  // If this is a c-shaped block, statement blocks cannot be connected
+  // anywhere other than inside the first statement input.
+  if (firstStatementConnection) {
+    // Can't connect if there is alread a block inside the first statement
```
tmickel: Hmm...I think the actual behavior we want is...if there is a block inside the first statement, insert it (don't wrap).

Right now it seems you can insert a "full pants" with something inside it only in the last slot of another pants (if that makes sense).

rachel-fenichel: You're totally right.  This breaks that.

Issue #133, opened 2016-03-24
Use translate3d when supported


tmickel:
We made a test branch that uses a translate3d style property on all SVG components that currently use translate.This seems to improve block drag smoothness on touch devices by quite a bit.

We'd probably like to detect when a browser supports translate3d, and when it does, use it instead of translate.

The performance of it degrades (back to normal levels) when multiple blocks are on the workspace. I'm currently investigating why this might be and whether we can get them all to the same level of smoothness.

@picklesrus @NeilFraser any reasons not to do this?

PR #157, opened 2016-04-01
Move block dragging to a separate surface/layer

tmickel:
This gives improved block dragging performance (we're getting a smooth 60 FPS) as discussed in #133 (note: I didn't actually switch to using translate3d in this PR - just moving blocks to a separate surface during dragging. Will hopefully get in translate3d in a separate PR, but even just this separate layer seems to give the same effect).

The general approach is to create a new layer by having a top-level SVG just for dragging. Blocks move themselves in and out of the drag SVG when they are in unrestricted drag mode. I've had to add another wrapper element to encompass blocklySvg and the drag layer in order to position the drag layer above the blocklySvg. The drag layer is the same size and maintains a group with the same translation and scale as the actual blockCanvas to simplify the work of moving between coordinate systems.

@rachel-fenichel @picklesrus would you like to review?


core/block_svg.js
```diff
@@ -237,6 +244,11 @@ Blockly.BlockSvg.terminateDrag_ = function() {
   if (Blockly.dragMode_ == Blockly.DRAG_FREE) {
     // Terminate a drag operation.
     if (selected) {
+      if (selected.isOnDragSurface_) {
```
rachel-fenichel: When would selected not be on the drag surface?  That is, is it an error if it is not on the drag surface at this point?

core/block_svg.js
```diff
@@ -290,8 +302,16 @@ Blockly.BlockSvg.prototype.setParent = function(newParent) {
   if (this.parentBlock_ && svgRoot) {
     // Move this block up the DOM.  Keep track of x/y translations.
     var xy = this.getRelativeToSurfaceXY();
-    this.workspace.getCanvas().appendChild(svgRoot);
-    svgRoot.setAttribute('transform', 'translate(' + xy.x + ',' + xy.y + ')');
+    if (Blockly.selected == this) {
```
rachel-fenichel: This code doesn't seem like it belongs here.  You haven't checked if newParent is null, so this says
if told to set a parent on the block in Blockly.selected, actually move it up to the drag surface.  For any other block just move it to the workspace canvas.

Can you explain?

core/block_svg.js
```diff
@@ -315,6 +335,8 @@ Blockly.BlockSvg.prototype.setParent = function(newParent) {
 Blockly.BlockSvg.prototype.getRelativeToSurfaceXY = function() {
```
rachel-fenichel, line 325: Add a comment that the drawing surface will either be the drag surface or the main canvas.

core/block_svg.js
```diff
@@ -337,14 +360,22 @@ Blockly.BlockSvg.prototype.moveBy = function(dx, dy) {
...
 };
 
 /**
+* Set this block to an absolute translation.
+* @param {number} x Horizontal translation.
+* @param {number} y Vertical translation.
+*/
+Blockly.BlockSvg.prototype.translate = function(x, y) {
+  this.getSvgRoot().setAttribute('transform', 'translate(' + x + ',' + y + ')');
+}
```
rachel-fenichel: Semicolon

core/block_svg.js
```diff
@@ -337,14 +360,22 @@ Blockly.BlockSvg.prototype.moveBy = function(dx, dy) {
...
-  this.getSvgRoot().setAttribute('transform',
-      'translate(' + (xy.x + dx) + ',' + (xy.y + dy) + ')');
+  this.translate(xy.x + dx, xy.y + dy);
   this.moveConnections_(dx, dy);
   event.recordNew();
   Blockly.Events.fire(event);
 };
 
 /**
+* Set this block to an absolute translation.
```
rachel-fenichel: Indent

core/block_svg.js
```diff
@@ -805,6 +836,12 @@ Blockly.BlockSvg.prototype.onMouseMove_ = function(e) {
   }
   if (Blockly.dragMode_ == Blockly.DRAG_FREE) {
     // Unrestricted dragging.
+    if (!this.isOnDragSurface_) {
+      // Move to the drag surface
```
rachel-fenichel: For clarity I think you should have functions for moveToDragSurface and moveFromDragSurface

core/dragsurface_svg.js
```diff
@@ -0,0 +1,102 @@
...
+  this.dragGroup_.setAttribute('transform', transform);
+};
+
+/**
+* Provide a reference to the drag group (primarily for BlockSvg.getRelativeToSurfaceXY).
+* @return {Element} Drag surface group element
+*/
+Blockly.DragSurfaceSvg.prototype.getGroup = function () {
+  return this.dragGroup_;
+}
```
rachel-fenichel: semicolon

core/dragsurface_svg.js
```diff
@@ -0,0 +1,102 @@
...
+ * @param {Number} y Y translation
+ * @param {Number} scale Scale of the group
+ */
+Blockly.DragSurfaceSvg.prototype.translateAndScaleGroup = function (x, y, scale) {
+  var transform = 'translate(' + x + ', ' + y + ') scale(' + scale + ')';
+  this.dragGroup_.setAttribute('transform', transform);
+};
+
+/**
+* Provide a reference to the drag group (primarily for BlockSvg.getRelativeToSurfaceXY).
```
rachel-fenichel: indent

core/dragsurface_svg.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+ /**
+  * Clear the group and hide the surface; move the blocks off onto the provided element.
+  * @param {!Element} newSurface Surface the dragging blocks should be moved to
+  */
+Blockly.DragSurfaceSvg.prototype.clearAndHide = function (newSurface) {
+  newSurface.appendChild(this.dragGroup_.childNodes[0]);
+  this.SVG_.style.display = 'none';
+  goog.asserts.assert(this.dragGroup_.childNodes.length == 0, 'Drag group was not cleared.');
```
rachel-fenichel, line 103: I don't see where you are clearing.  Is it a side effect of appendChild or setting the display attribute?

tmickel, line 103: It's a side effect of appendChild - I'll add a note.

core/workspace_svg.js
```diff
@@ -192,6 +203,9 @@ Blockly.WorkspaceSvg.prototype.createDom = function(opt_backgroundClass) {
     this.addFlyout_();
   }
   this.updateGridPattern_();
+  if (this.dragSurface) {
+    this.dragSurface.createDom();
```
rachel-fenichel: Move the createDom call up into inject.js.  There's no compelling reason for createDom calls to be here (I've checked with Neil).

core/block_svg.js
tmickel: Essentially, the problem was that it was impossible to keep blocks whose parent changed in the drag surface.

The following happened:
1. Drag starts - block is moved to drag surface
2. Blockly changes the parent because it was disconnected. Then this code was previously moving it up to the workspace canvas, out of the drag surface :(
3. Now it's off the drag surface even though it's still dragging (i.e., Blockly.selected == this).

I couldn't think of a better place to put this, any thoughts, or should I just document this better?

core/workspace_svg.js
```diff
@@ -360,6 +374,9 @@ Blockly.WorkspaceSvg.prototype.translate = function(x, y) {
       'scale(' + this.scale + ')';
   this.svgBlockCanvas_.setAttribute('transform', translation);
   this.svgBubbleCanvas_.setAttribute('transform', translation);
+  if (this.dragSurface) {
+    this.dragSurface.translateAndScaleGroup(x, y, this.scale);
```
picklesrus, line 375: Is the translation you build in translateAndScaleGroup the same as the one above?  Did you consider passing that to translateAndScale? Or are you keeping it separate because you plan on changing it to 3d later?

tmickel, line 375: It's the same transform, but yes - kept it separate in case we want to make that 3d.

PR #172, opened 2016-04-06
Correctly ghost the last block in the stack.

rachel-fenichel:
Builds on https://github.com/LLK/scratch-blockly/pull/171


core/block_svg.js
```diff
@@ -886,6 +893,14 @@ Blockly.BlockSvg.prototype.updatePreviews = function(closestConnection,
     if (Blockly.insertionMarker_ && Blockly.insertionMarkerConnection_) {
       Blockly.BlockSvg.disconnectInsertionMarker();
     }
+    // If there's already an insertion marker but it's representing the wrong
+    // block, delete it so we can create the correct one.
```
tmickel, line 897: Hmm, why does this happen?

rachel-fenichel, line 897: Because the insertion marker either represents the first block in the stack of the last one.  Previously it only represented the first one, but now it can switch between.

To avoid creating blocks over and over again we just hide the insertion marker between times that we want to ghost, so it may exist but be the wrong shape.

PR #178, opened 2016-04-08
Initial mockup of LEGO WeDo blocks

thisandagain:
- Adds initial mockup of WeDo blocks as per spec
- Remove unused "looks" and "motion" blocks
- Adds event broadcast / receive blocks
- Adds pause / stop blocks
- Updates horizontal playground

/cc @carljbowman


tests/wrapper/playgrounds/horizontal_playground_compressed.html
```diff
@@ -1,144 +0,0 @@
-<!DOCTYPE html>
```
tmickel, line 1: I think there was a reason for this file - I think it's being used in the example XCode wrapper in tests/. But I don't think it's super important right now.

blocks_horizontal/event.js
```diff
@@ -53,12 +33,63 @@ Blockly.Blocks['event_whenflagclicked'] = {
...
+      "colourTertiary": Blockly.Colours.event.tertiary
     });
+  }
+};
 
-    this.setHelpUrl(function () {
-        return 'halp me plz. k thx bye.';
+Blockly.Blocks['event_whenbroadcastreceived'] = {
+  /**
+   * Block for when flag clicked.
```
tmickel: Also not super important, but noticed this comment and the one for event_broadcast are for the wrong block.

Issue #179, opened 2016-04-08
Scrolling the flyout when touch direction on blocks is not toward workspace.


tmickel:
When we're using a fixed flyout, we need to be able to scroll the pane when touching blocks. Add some test of the direction of the touch, and scroll/create a new block accordingly.

PR #183, opened 2016-04-08
Only create new blocks in flyout at certain angle #179

tmickel:
This is especially important on touch - it's hard to scroll the flyout without this.

I explicitly spelled out the logic for when to initiate a drag per position. If you think of a clever way to encapsulate all of that, it would be great to replace it :)

@thisandagain

### DIFF
--- a/core/flyout.js
+++ b/core/flyout.js
@@ -788,7 +788,8 @@ Blockly.Flyout.prototype.onMouseMoveBlock_ = function(e) {
     }
   }
 
-  // Still dragging within the sticky DRAG_RADIUS.
+  // Don't create a block within the sticky drag radius,
+  // or if we failed the direction check.
   if (Math.sqrt(dx * dx + dy * dy) > Blockly.DRAG_RADIUS && directionCheck) {
     // Create the block.
     Blockly.Flyout.startFlyout_.createBlockFunc_(Blockly.Flyout.startBlock_)(
--- a/core/flyout.js
+++ b/core/flyout.js
@@ -153,7 +153,7 @@ Blockly.Flyout.prototype.verticalOffset_ = 0;
  * @type {number}
  * @private
 */
-Blockly.Flyout.prototype.dragTowardWorkspaceOrthogonalAngleRange_ = 70;
+Blockly.Flyout.prototype.dragAngleRange_ = 70;
 
 /**
  * Creates the flyout's DOM.  Only needs to be called once.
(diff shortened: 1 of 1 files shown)


core/flyout.js
```diff
@@ -751,12 +760,46 @@ Blockly.Flyout.prototype.onMouseMoveBlock_ = function(e) {
...
+    } else {
+      if (direction < -180 + range / 2 || direction > 180 - range / 2) {
+        directionCheck = true;
+      }
+    }
+  }
+
+  // Don't create a block within the sticky drag radius,
+  // or if we failed the direction check.
+  if (Math.sqrt(dx * dx + dy * dy) > Blockly.DRAG_RADIUS && directionCheck) {
```
rachel-fenichel: Swap the order to short-circuit if directionCheck is false.

core/flyout.js
```diff
@@ -751,12 +760,46 @@ Blockly.Flyout.prototype.onMouseMoveBlock_ = function(e) {
...
+    if (Blockly.Flyout.startFlyout_.toolboxPosition_ == Blockly.TOOLBOX_AT_TOP) {
+      if (direction < 90 + range / 2 && direction > 90 - range / 2) {
+        directionCheck = true;
+      }
+    } else {
+      if (direction > -90 - range / 2 && direction < -90 + range / 2) {
+        directionCheck = true;
+      }
+    }
+  } else {
```
rachel-fenichel, line 814: Comment: // Vertical

core/flyout.js
```diff
@@ -751,12 +760,46 @@ Blockly.Flyout.prototype.onMouseMoveBlock_ = function(e) {
   }
   var dx = e.clientX - Blockly.Flyout.startDownEvent_.clientX;
   var dy = e.clientY - Blockly.Flyout.startDownEvent_.clientY;
-  // Still dragging within the sticky DRAG_RADIUS.
-  if (Math.sqrt(dx * dx + dy * dy) > Blockly.DRAG_RADIUS) {
+
+  // Direction goes from -180 to 180, with 0 toward the right and 90 on top.
+  var direction = Math.atan2(dy, dx) / Math.PI * 180;
```
rachel-fenichel: All of this should be in a separate function.

core/flyout.js
```diff
@@ -751,12 +760,46 @@ Blockly.Flyout.prototype.onMouseMoveBlock_ = function(e) {
...
+
+  // Do a direction check based on the flyout position
```
rachel-fenichel: Based on flyout position and orientation.

rachel-fenichel: "Do a direction check" doesn't really mean anything.  How about:

"Only drag a block out of the flyout if the user is dragging toward the workspace.  Otherwise the drag might be the start of a scroll inside the flyout.  Use the flyout position and orientation to determine which action to take."

core/flyout.js
```diff
@@ -141,6 +141,13 @@ Blockly.Flyout.prototype.height_ = 0;
 Blockly.Flyout.prototype.verticalOffset_ = 0;
 
 /**
+ * Angle size at which blocks are created from a fixed flyout on touch.
```
rachel-fenichel: Rephrase/clarify.

core/flyout.js
```diff
@@ -141,6 +141,21 @@ Blockly.Flyout.prototype.height_ = 0;
...
+ * Example:
+ * Flyout                                                  Edge   Workspace
+ * [block] /  <-within this angle, drags "toward workspace" |
+ * [block] ---- orthogonal to flyout boundary ----          |
+ * [block] \                                                |
+ * The angle is given in degrees from the orthogonal.
+ * @type {number}
+ * @private
+*/
+Blockly.Flyout.prototype.dragTowardWorkspaceOrthogonalAngleRange_ = 70;
```
rachel-fenichel: I think it's fine to call this dragAngle now that it has a good comment.

PR #187, opened 2016-04-08
Implement translate3d during drag

tmickel:
@picklesrus

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -819,10 +819,7 @@ Blockly.BlockSvg.prototype.moveToDragSurface_ = function() {
   // is equal to the current relative-to-surface position,
   // to keep the position in sync as it move on/off the surface.
   var xy = this.getRelativeToSurfaceXY();
-  if (Blockly.is3dSupported() && this.getSvgRoot().hasAttribute('transform')) {
-    // Existing transform needs to be overwritten by translate3d
-    this.getSvgRoot().removeAttribute('transform');
-  }
+  this.clearTransformAttributes_();
   this.translate(xy.x, xy.y, Blockly.is3dSupported());
   // Execute the move on the top-level SVG component
   this.workspace.dragSurface.setBlocksAndShow(this.getSvgRoot());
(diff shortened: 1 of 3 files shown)


core/block_svg.js
```diff
@@ -829,6 +835,13 @@ Blockly.BlockSvg.prototype.moveToDragSurface_ = function() {
  */
  Blockly.BlockSvg.prototype.moveOffDragSurface_ = function() {
   this.workspace.dragSurface.clearAndHide(this.workspace.getCanvas());
+  // Translate to current position, turning off 3d.
+  var xy = this.getRelativeToSurfaceXY();
+  if (Blockly.is3dSupported() && this.getSvgRoot().hasAttribute('style')) {
+    // Remove conflicting style attribute before applying translate
+    this.getSvgRoot().removeAttribute('style');
```
rachel-fenichel: Why not put checking for conflicting style and transform attributes inside translate()?

core/utils.js
```diff
@@ -271,6 +271,18 @@ Blockly.getRelativeXY_ = function(element) {
       xy.y += parseFloat(r[3]);
     }
   }
+
+  // Third, check for style="transform: translate3d(...)".
+  var style = element.getAttribute('style');
+  var r3;
+  r3 = style && style.indexOf('translate3d') > -1 && style.match(Blockly.getRelativeXY_.XY_3D_REGEXP_);
+  if (r3) {
```
rachel-fenichel:
```
if (style && style.indexOf('translate3d') > -1 && style.match(Blockly.getRelativeXY_.XY_3D_REGEXP_) {
  xy.x += parseFloat(style[1])
  if (style[3]) {
    xy.y += parseFloat(style[3])
  }
}
```

core/block_svg.js
tmickel: I'm worried the check might slow it down - translate needs to be called at 60fps.

core/utils.js
tmickel: I think this change would actually be incorrect - we need to save the result from style.match. I was matching the style from the other transform check in the function. However I think it will be best to just refactor both to make it clear what's going on.

core/utils.js
```diff
@@ -320,6 +344,53 @@ Blockly.getSvgXY_ = function(element, workspace) {
 };
 
 /**
+ * Cached value for whether 3D is supported
+ * @type {boolean}
+ * @private
+ */
+Blockly.cache3dSupported_ = null;
```
rachel-fenichel: Move this above function definitions.

PR #190, opened 2016-04-11
Create basic framework for animating fields on disposal

tmickel:
Adds functions and timers for WidgetDiv disposal. In essence, allow WidgetDivs users to provide extra time before the WidgetDiv is removed/disposed (and before the DOM is hidden), and an extra callback for when the timer is up (so properties can be cleaned up before the DOM is finally hidden).

### DIFF
--- a/core/widgetdiv.js
+++ b/core/widgetdiv.js
@@ -132,11 +132,8 @@ Blockly.WidgetDiv.hide = function() {
     Blockly.WidgetDiv.disposeAnimationTimer_ = null;
     Blockly.WidgetDiv.owner_ = null;
     Blockly.WidgetDiv.hideAndClearDom_();
-    return;
   } else if (Blockly.WidgetDiv.isVisible()) {
     // No animation timer set, but the widget is visible
-    Blockly.WidgetDiv.owner_ = null;
-    Blockly.WidgetDiv.hideAndClearDom_();
     // Start animation out (or immediately hide)
     Blockly.WidgetDiv.dispose_ && Blockly.WidgetDiv.dispose_();
     Blockly.WidgetDiv.dispose_ = null;
@@ -148,6 +145,7 @@ Blockly.WidgetDiv.hide = function() {
       );
     } else {
       // No timer provided - auto-hide the DOM now.
+      Blockly.WidgetDiv.owner_ = null;
       Blockly.WidgetDiv.hideAndClearDom_();
     }
     Blockly.Events.setGroup(false);
--- a/core/widgetdiv.js
+++ b/core/widgetdiv.js
@@ -61,7 +61,7 @@ Blockly.WidgetDiv.dispose_ = null;
  * @type {Function}
  * @private
  */
-Blockly.WidgetDiv.disposeAnimateFinished_ = null;
+Blockly.WidgetDiv.disposeAnimationFinished_ = null;
 
 /**
  * Timer ID for the dispose animation.
@@ -75,7 +75,7 @@ Blockly.WidgetDiv.disposeAnimationTimer_ = null;
  * @type {number}
  * @private
  */
-Blockly.WidgetDiv.disposeAnimateTimerLength_ = 0;
+Blockly.WidgetDiv.disposeAnimationTimerLength_ = 0;
 
 
 /**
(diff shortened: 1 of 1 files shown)


core/widgetdiv.js
```diff
@@ -48,12 +48,37 @@ Blockly.WidgetDiv.owner_ = null;
...
  */
 Blockly.WidgetDiv.dispose_ = null;
 
 /**
+ * Optional function called at the end of a dispose animation.
+ * Set by whichever object is using the widget.
+ * @type {Function}
+ * @private
+ */
+Blockly.WidgetDiv.disposeAnimateFinished_ = null;
```
rachel-fenichel: Call all of these disposeAnimate\* or disposeAnimation*, but don't mix them.

core/widgetdiv.js
```diff
@@ -89,20 +121,52 @@ Blockly.WidgetDiv.show = function(newOwner, rtl, dispose) {
...
  */
 Blockly.WidgetDiv.hide = function() {
-  if (Blockly.WidgetDiv.owner_) {
+  if (Blockly.WidgetDiv.disposeAnimationTimer_) {
+    // An animation timer is set already.
+    // This happens when a previous widget was animating out,
+    // but Blockly is hiding the widget to create a new one.
+    // So, short-circuit the animation and clear the timer.
+    window.clearTimeout(Blockly.WidgetDiv.disposeAnimationTimer_);
+    Blockly.WidgetDiv.disposeAnimateFinished_ && Blockly.WidgetDiv.disposeAnimateFinished_();
```
rachel-fenichel: Is this the same as 
if (disposeAnimateFinished) {
disposeAnimateFinished();
}

?

(checking if I understand js yet)

core/widgetdiv.js
```diff
@@ -89,20 +121,52 @@ Blockly.WidgetDiv.show = function(newOwner, rtl, dispose) {
...
-    goog.dom.removeChildren(Blockly.WidgetDiv.DIV);
+    // If we want to animate out, set the appropriate timer for final dispose.
+    if (Blockly.WidgetDiv.disposeAnimateFinished_) {
+      Blockly.WidgetDiv.disposeAnimationTimer_ = window.setTimeout(
+        Blockly.WidgetDiv.hide, // Come back to hide and take the first branch.
+        Blockly.WidgetDiv.disposeAnimateTimerLength_ * 1000
+      );
+    } else {
+      // No timer provided - auto-hide the DOM now.
+      Blockly.WidgetDiv.hideAndClearDom_();
```
rachel-fenichel, line 149: You have already called hideAndClearDom_() by now.  Are your if/elses matched properly?

core/widgetdiv.js
tmickel: Yes - I don't like the idiom personally...

core/widgetdiv.js
tmickel, line 149: Good catch here, thanks! I've fixed up the logic - hideAndClearDom_ should only be called if we're not trying to do an animation, or after the animation.

core/widgetdiv.js
```diff
@@ -89,20 +121,50 @@ Blockly.WidgetDiv.show = function(newOwner, rtl, dispose) {
...
+    Blockly.WidgetDiv.hideAndClearDom_();
+  } else if (Blockly.WidgetDiv.isVisible()) {
+    // No animation timer set, but the widget is visible
+    // Start animation out (or immediately hide)
     Blockly.WidgetDiv.dispose_ && Blockly.WidgetDiv.dispose_();
     Blockly.WidgetDiv.dispose_ = null;
-    goog.dom.removeChildren(Blockly.WidgetDiv.DIV);
+    // If we want to animate out, set the appropriate timer for final dispose.
+    if (Blockly.WidgetDiv.disposeAnimationFinished_) {
+      Blockly.WidgetDiv.disposeAnimationTimer_ = window.setTimeout(
```
picklesrus, line 142: You might have better luck with requestAnimationFrame than setTimeout.  It is usually recommended over setTimeout for animation.  There are lots of blog posts about it. Here's a semi-random one http://www.html5rocks.com/en/tutorials/speed/rendering/#toc-raf

Yours doesn't appear to be an expensive animation, but it might be worth a quick check.

tmickel, line 142: Hmmm...I think a timeout is more appropriate here. Basically the user of the WidgetDiv has two callbacks (dispose_ and disposeAnimationFinished_) and the timeout is for providing some time for the WidgetDiv user in between the two. A user of WidgetDiv should likely use setAnimationFrame for the actual rendering of the animation. Does that make sense?

picklesrus, line 142: Oh, whoops, yes. Sorry about that.  I misread and thought this was calling rendering of the animation.  I see that's in the first branch though.  Carry on!

PR #202, opened 2016-04-12
Fix the reported height of blocks with fields

### LINKED ISSUE #153: Toolbox scrollbar overlaps input fields
toolbox scrollbar

Steps to reproduce:
1. Open horizontal_playground.html
2. Select Pants tab.
3. Reduce window horizontal size until a scrollbar appears on the toolbox.

Expected behaviour:
- When selecting Pants tab, the toolbox becomes bigger to create margin under the input field.
- OR the toolbox becomes bigger when the scrollbar is rendered.

drigz:
These weren't taking their fields into account when calculating their
height. Fixes #153.

### DIFF
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -236,7 +236,7 @@ Blockly.BlockSvg.prototype.getHeightWidth = function(opt_ignoreFields) {
   var height = this.height;
   var width = this.width;
   // Add the size of the field shadow block.
-  if (!opt_ignoreFields && this.getValueInput_()) {
+  if (!opt_ignoreFields && this.getFieldShadowBlock_()) {
     height += Blockly.BlockSvg.FIELD_Y_OFFSET;
     height += Blockly.BlockSvg.FIELD_HEIGHT;
   }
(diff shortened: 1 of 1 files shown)


core/block_render_svg_horizontal.js
```diff
@@ -603,3 +609,18 @@ Blockly.BlockSvg.prototype.renderDrawTop_ =
...
+ */
+Blockly.BlockSvg.prototype.getValueInput_ = function() {
+  for (var i = 0, child; child = this.childBlocks_[i]; i++) {
+    if (child.isShadow()) {
+      return child;
+    }
+  }
+
+  return null;
+}
```
rachel-fenichel: semicolon and newline

core/block_render_svg_horizontal.js
```diff
@@ -603,3 +609,18 @@ Blockly.BlockSvg.prototype.renderDrawTop_ =
   }
   steps.push('z');
 };
+
+/**
+ * Get the field shadow block, if this block has one.
+ * @return {Blockly.BlockSvg} The field shadow block, or null if not found.
+ * @private
+ */
+Blockly.BlockSvg.prototype.getValueInput_ = function() {
```
rachel-fenichel: Are we guaranteed to only have one field per block?  This will work for now but we will probably have to come back to this for vertical scratch.

Add a comment that this is scratch-specific logic.

getValueInput is a misnomer (it's a block, not an input).  How about getFieldShadowBlock_()?

core/block_render_svg_horizontal.js
drigz: Added the semicolon, thanks. What newline are you referring to?

core/block_render_svg_horizontal.js
drigz: > Are we guaranteed to only have one field per block? This will work for now but we will probably have to come back to this for vertical scratch.

For horizontal, yes. For vertical, the [grammar](https://cloud.githubusercontent.com/assets/3409578/13230643/2c4b4518-d974-11e5-9765-9f8ef4e82efa.png) doesn't show fields being rendered outside the main block area, so this special case isn't necessary.

> getValueInput is a misnomer (it's a block, not an input). How about getFieldShadowBlock_()?

Thanks, I was unsure about the naming as different terms are used in different places.

PR #210, opened 2016-04-13
Implement translation of entire drag surface SVG to prevent repaint.

tmickel:
-Extend DragSurfaceSvg so that the entire SVG surface can be translated.
-Update getRelativeXY_ to include the drag surface translation when the block is on that surface.
-Translate the surface on the call to moveToDragSurface_ and during the mousemove event.

@picklesrus

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -903,7 +903,7 @@ Blockly.BlockSvg.prototype.onMouseMove_ = function(e) {
   if (Blockly.dragMode_ == Blockly.DRAG_FREE) {
     var dx = oldXY.x - this.dragStartXY_.x;
     var dy = oldXY.y - this.dragStartXY_.y;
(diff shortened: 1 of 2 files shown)


core/block_svg.js
```diff
@@ -896,7 +903,7 @@ Blockly.BlockSvg.prototype.onMouseMove_ = function(e) {
   if (Blockly.dragMode_ == Blockly.DRAG_FREE) {
     var dx = oldXY.x - this.dragStartXY_.x;
     var dy = oldXY.y - this.dragStartXY_.y;
-    this.translate(newXY.x, newXY.y, Blockly.is3dSupported());
+    this.workspace.dragSurface.translateSurface(newXY.x, newXY.y, Blockly.is3dSupported());
```
tmickel: Noticed there is no actual argument here - will fix.

rachel-fenichel: Just removing the third arg, since it's called inside translateSurface?

After that fix, looks good.

PR #218, opened 2016-04-14
Slight refactor of colour handling; add additional properties

tmickel:
This allows us to override colour properties at injection-time. This is nice because it means we can override things like insertion marker colour on a per-workspace basis.

Also adds a bunch of properties to our colour library (toolbox, flyout, scrollbars, insertion markers), provides a per-block opacity property (set on insertion markers). Also made our horizontal workspace in "dark mode" for fun.

@thisandagain

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -87,6 +87,12 @@ Blockly.Css.inject = function(hasCss, pathToMedia) {
   // Strip off any trailing slash (either Unix or Windows).
   Blockly.Css.mediaPath_ = pathToMedia.replace(/[\\\/]$/, '');
   text = text.replace(/<<<PATH>>>/g, Blockly.Css.mediaPath_);
+  // Dynamically replace colours in the CSS text, in case they have
+  // been set at run-time injection.
+  for (var colourProperty in Blockly.Colours) {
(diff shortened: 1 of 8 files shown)


core/css.js
```diff
@@ -87,6 +87,12 @@ Blockly.Css.inject = function(hasCss, pathToMedia) {
   // Strip off any trailing slash (either Unix or Windows).
   Blockly.Css.mediaPath_ = pathToMedia.replace(/[\\\/]$/, '');
   text = text.replace(/<<<PATH>>>/g, Blockly.Css.mediaPath_);
+  // Dynamically replace colours in the CSS text, in case they have
+  // been set at run-time injection.
+  for (var colourProperty in Blockly.Colours) {
+    if (!Blockly.Colours.hasOwnProperty(colourProperty)) continue;
```
rachel-fenichel: prefer 
if (condition) {
stuff;
}

to 
if !(condition) {
continue;
}
stuff

tests/horizontal_playground.html
```diff
@@ -45,6 +50,14 @@
       minScale: 0.25,
       scaleSpeed: 1.1
     },
+    colours: {
```
thisandagain, line 53: :+1:

PR #219, opened 2016-04-14
Separate CORNER_RADIUS from block positioning in flyout; set radius to 0

tmickel:
We want the default flyout to be unrounded. For some reason this property was bound up in the positioning of blocks, so I separated it out.

### DIFF
--- a/core/flyout.js
+++ b/core/flyout.js
@@ -113,7 +113,7 @@ Blockly.Flyout.prototype.autoClose = true;
 Blockly.Flyout.prototype.CORNER_RADIUS = 0;
 
 /**
- * Corner radius of the flyout background.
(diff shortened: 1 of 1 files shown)


core/flyout.js
```diff
@@ -110,7 +110,14 @@ Blockly.Flyout.prototype.autoClose = true;
  * @type {number}
  * @const
  */
-Blockly.Flyout.prototype.CORNER_RADIUS = 8;
+Blockly.Flyout.prototype.CORNER_RADIUS = 0;
+
+/**
+ * Corner radius of the flyout background.
```
rachel-fenichel: Update comment.

tmickel: Thanks - that was silly...

PR #227, opened 2016-04-15
Allow WidgetDiv to hide immediately, no animation

tmickel:
And add calls to this where the div animation may become out of sync with the field itself.

When hide() is normally called by hideChaff, the animate-out was being triggered. This was bad, because often hideChaff was called before the block itself was being moved. Then the WidgetDiv was animating out, but the field was visible in a different location - just generally looking weird.

A bit of a spaghetti fix but unfortunately since hideChaff is used for so many things in so many places, trying to fix it there seemed even more of a spaghetti fix.

### DIFF
--- a/core/widgetdiv.js
+++ b/core/widgetdiv.js
@@ -127,7 +127,7 @@ Blockly.WidgetDiv.hide = function(opt_noAnimate) {
     // This happens when a previous widget was animating out,
     // but Blockly is hiding the widget to create a new one.
     // So, short-circuit the animation and clear the timer.
-    Blockly.WidgetDiv.disposeAnimationTimer_ && window.clearTimeout(Blockly.WidgetDiv.disposeAnimationTimer_);
+    window.clearTimeout(Blockly.WidgetDiv.disposeAnimationTimer_);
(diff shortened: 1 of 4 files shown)


core/widgetdiv.js
```diff
@@ -119,14 +119,15 @@ Blockly.WidgetDiv.show = function(newOwner, rtl, opt_dispose,
...
  */
-Blockly.WidgetDiv.hide = function() {
+Blockly.WidgetDiv.hide = function(opt_noAnimate) {
   if (Blockly.WidgetDiv.disposeAnimationTimer_) {
     // An animation timer is set already.
     // This happens when a previous widget was animating out,
     // but Blockly is hiding the widget to create a new one.
     // So, short-circuit the animation and clear the timer.
-    window.clearTimeout(Blockly.WidgetDiv.disposeAnimationTimer_);
+    Blockly.WidgetDiv.disposeAnimationTimer_ && window.clearTimeout(Blockly.WidgetDiv.disposeAnimationTimer_);
```
rachel-fenichel: You don't need to add this--it's checked on line 125.

tmickel: Yep - missed that in clean-up. Thanks! anything else?

Issue #147, opened 2016-03-29
String Input States - Default & Focusing


carljbowman:
## Horizontal: Sizing / Radius
**Alignment has more nuance based on the block type. e.g Hat Block vs Default Block**
![r4_standard-string-input-block](https://cloud.githubusercontent.com/assets/3409578/14120751/4a2dc00a-f5c2-11e5-8970-246aee189235.png)
![r4_standard-string-input-block](https://cloud.githubusercontent.com/assets/3409578/14121129/1f626338-f5c4-11e5-986a-2f71794e5a58.png)

## Sizing: Default / Focus 
![r1_string-input-focusing](https://cloud.githubusercontent.com/assets/3409578/14120639/a5a5181c-f5c1-11e5-8fbf-128e5d36c6c3.jpg)
**Hacky Prototype:**
[fields-prototype.zip](https://github.com/LLK/scratch-blockly/files/194161/fields-prototype.zip)

## Color: Default / Focus
![r1_string-input-focus-color](https://cloud.githubusercontent.com/assets/3409578/14120689/ebbffa4c-f5c1-11e5-8283-6ba031e311c5.jpg)
![r1_horizontal-colors](https://cloud.githubusercontent.com/assets/3409578/14120778/6b76f358-f5c2-11e5-9358-d8258030ced3.jpg)

## Input: Desktop / Device
![r1_string-input-focus-input](https://cloud.githubusercontent.com/assets/3409578/14120698/f755058c-f5c1-11e5-8101-34a83ecfe28b.jpg)

Issue #146, opened 2016-03-29
Number Inputs States - Default & Focusing


carljbowman:
## Horizontal: Sizing / Radius
**Alignment has more nuance based on the block type. e.g Hat Block vs Default Block**
![r4_standard-int-input-block](https://cloud.githubusercontent.com/assets/3409578/14119913/761c9988-f5be-11e5-84c6-def80444e4fe.png)
![r4_standard-int-input-block radius](https://cloud.githubusercontent.com/assets/3409578/14121155/3c7eb750-f5c4-11e5-9683-ff15689cac1c.png)

## Sizing: Default / Focus 
![r1_number-input-focusing](https://cloud.githubusercontent.com/assets/3409578/14119982/bf21ec96-f5be-11e5-86bf-a34bd03ce69d.jpg)
**Hacky prototype:**
[fields-prototype.zip](https://github.com/LLK/scratch-blockly/files/194161/fields-prototype.zip)

## Color: Default / Focus
![r1_number-input-focus-color](https://cloud.githubusercontent.com/assets/3409578/14120008/df45abd4-f5be-11e5-9652-63732a495f93.jpg)
![r1_horizontal-colors](https://cloud.githubusercontent.com/assets/3409578/14120046/1d6a69b8-f5bf-11e5-8cb7-a147ab42bb94.jpg)

## Input: Desktop / Device
![r1_number-input-focus-input](https://cloud.githubusercontent.com/assets/3409578/14120034/0cb7f4d2-f5bf-11e5-927f-4cf657c51a84.jpg)

PR #230, opened 2016-04-15
Implement number and string field specs

tmickel:
This implements everything in #146 and #147, minus verification that it works on vertical (in theory, I think it should) (also minus the keyboard/numpad dropdown).


core/field.js
```diff
@@ -78,7 +78,7 @@ Blockly.Field.prototype.name = undefined;
  * Maximum characters of text to display before adding an ellipsis.
  * @type {number}
  */
-Blockly.Field.prototype.maxDisplayLength = 5;
+Blockly.Field.prototype.maxDisplayLength = 4;
```
rachel-fenichel, line 82: Is this the same for strings and numbers?  If so, add comment.  If not, rename to maxStringDisplayLength

PR #233, opened 2016-04-18
Implement drop-down menus and icon picker field

tmickel:
What's done:
-The drop-down menu, I think to completion. It animates in and out, stays within bounds, and adjusts appropriately.
-The icon picker. Icons are shown in the drop-down and you can click them to select a new one. The values are appropriately saved in the XML and events are created for the change. Added the basic accessibility properties from the closure dropdown as well. It should respond to touch. You define which icons are available and their machine-readable text in the block definition.

What's not done:
- The final visual design of the icon picker. I made something up but we still need to figure out exactly what this should look like, what the colors should be and how it should behave for selection states, etc.
- The final visual implementation of what happens when you press the picker button. I put in a changing-color version just to get us started. But I think we want this to match the look and feel of the text input eventually (a big mess, because that means we need to make a WidgetDiv for the animation :()
- Layout for the "cross" icon picker for tilt direction (should be easy).
- Putting in the rest of the actual menus. Right now there is one example, the "when broadcast" block, using ScratchJr icons. Need assets from Carl to do the rest.

@thisandagain @drigz 

Rachel - I know this is a big PR. Let me know if you have any questions. Also if it's too big, I'm sure Andrew would be happy to review instead :)


blocks_horizontal/event.js
```diff
@@ -38,6 +38,35 @@ Blockly.Blocks['event_whenflagclicked'] = {
...
+              value: 'blue', width: 50, height: 50, alt: 'Blue'},
+            {src: Blockly.mainWorkspace.options.pathToMedia + 'icons/LetterGet_Green.svg',
+              value: 'green', width: 50, height: 50, alt: 'Green'},
+            {src: Blockly.mainWorkspace.options.pathToMedia + 'icons/LetterGet_Orange.svg',
+              value: 'orange', width: 50, height: 50, alt: 'Orange'},
+            {src: Blockly.mainWorkspace.options.pathToMedia + 'icons/LetterGet_Purple.svg',
+              value: 'purple', width: 50, height: 50, alt: 'Purple'},
+            {src: Blockly.mainWorkspace.options.pathToMedia + 'icons/LetterGet_Red.svg',
+              value: 'red', width: 50, height: 50, alt: 'Red'},
+            {src: Blockly.mainWorkspace.options.pathToMedia + 'icons/LetterGet_Yellow.svg',
```
rachel-fenichel, line 62: will this src ever not be in blockly.mainworkspace.options.pathToMedia?

tmickel, line 62: could be I suppose - for example if you're pulling something from the internet or something or using multiple paths to media.

core/dropdowndiv.js
```diff
@@ -0,0 +1,314 @@
+/**
```
rachel-fenichel, line 1: Missing license.

core/scrollbar.js
```diff
@@ -455,9 +456,10 @@ Blockly.Scrollbar.prototype.onMouseDownKnob_ = function(e) {
       'mouseup', this, this.onMouseUpKnob_);
   Blockly.Scrollbar.onMouseMoveWrapper_ = Blockly.bindEvent_(document,
       'mousemove', this, this.onMouseMoveKnob_);
-  // When the scrollbars are clicked, hide the WidgetDiv without animation
+  // When the scrollbars are clicked, hide the WidgetDiv/DropDownDiv without animation
   // in anticipation of a workspace move.
   Blockly.WidgetDiv.hide(true);
+  Blockly.DropDownDiv.hideWithoutAnimation();
```
rachel-fenichel, line 462: Feels weird that the widgetDiv has hide() and hide(true) but the DropdownDiv has hide() and hideWithoutAnimation()

core/dropdowndiv.js
tmickel, line 1: @thisandagain What license header should we use on new files?

core/scrollbar.js
tmickel, line 462: Mmm yes...after implementing both I think the separate hideWithoutAnimation is better. I'll make a bug to fix the WidgetDiv to be similar.

core/dropdowndiv.js
```diff
@@ -0,0 +1,314 @@
...
+ * the arrow will point at this origin and box will positioned below or above it.
+ * If we can maintain the container bounds at the primary point, the arrow will
+ * point there, and the container will be positioned below it.
+ * If we can't maintain the container bounds at the primary point, fall-back to the
+ * secondary point and position above.
+ * @param {number} primaryX Desired origin point x, in absolute px
+ * @param {number} primaryY Desired origin point y, in absolute px
+ * @param {number} secondaryX Secondary/alternative origin point x, in absolute px
+ * @param {number} secondaryY Secondary/alternative origin point y, in absolute px
+ * @param {Function=} opt_onHide Optional callback for when the drop-down is hidden
```
rachel-fenichel, line 182: Is there ever a case when it would fail to render at both the primary and the secondary locations?

rachel-fenichel, line 182: Never mind, I see that if neither location is valid it'll render at the primary.

blocks_horizontal/event.js
```diff
@@ -55,6 +87,10 @@ Blockly.Blocks['event_whenbroadcastreceived'] = {
           "height": 40,
```
drigz, line 87: Is the "src" still necessary? Is it guaranteed to be overridden by the dropdown? If I remove it it still seems to work.

blocks_horizontal/event.js
```diff
@@ -55,6 +87,10 @@ Blockly.Blocks['event_whenbroadcastreceived'] = {
...
           "alt": "flag",
           "flip_rtl": true
+        },
+        {
+          "type": "input_value",
+          "name": "CHOICE"
```
drigz, line 93: This "CHOICE" appears in four places: event_whenbroadcastreceived definition, dropdown_whenbroadcast definition, and twice in the toolbox XML. I guess it'll also need to be used at least once in each generator eg valueToCode("CHOICE") and getFieldValue("CHOICE").

I suspect this is a necessary side-effect of using shadow blocks for the inputs. Can you think of a way of reducing the repetition?

core/dropdowndiv.js
```diff
@@ -0,0 +1,337 @@
+/**
+ * @fileoverview A div that floats on top of the workspace, for drop-down menus.
+ * The drop-down can be kept inside the workspace, animate in/out, etc.
+ * @author tmickel@mit.edu (Tim Mickel)
+ */
+
+'use strict';
+
+goog.provide('Blockly.DropDownDiv');
```
drigz, line 19: nit: inconsistent use of drop-down/dropdown (in comments) and DropDown/Dropdown in identifiers.

blocks_horizontal/event.js
tmickel, line 87: Not purely guaranteed - suppose for example your dropdown has no items in it. In practice though I think you can leave out an src as it will be overridden.

blocks_horizontal/event.js
tmickel, line 93: I can't think of a way. I'd be happy to name them two different things if there were a clearer name for either or both. Let me know if you have ideas!

core/dropdowndiv.js
tmickel, line 19: Haha...I suppose I should stick to drop-down and DropDown

core/dropdowndiv.js
thisandagain, line 1: BSD-3-Clause: See [scratch-boilerplate](https://github.com/LLK/scratch-boilerplate)

core/scrollbar.js
thisandagain, line 462: +1

PR #234, opened 2016-04-18
Implementation of grow on drag

tmickel:
Here is a partial implementation of grow-on-drag.

Essentially it adds another wrapper to the drag layer - this one that gets a scale factor. It has to be separate from the translating drag layer itself, because that has its own `transform` property so we can't transition it.

I suspect we'll want one more thing - to "ungrow" when we're dropping on the workspace (but not dropping into a stack). That might be harder to pull off...

### DIFF
--- a/core/dragsurface_svg.js
+++ b/core/dragsurface_svg.js
@@ -82,6 +82,22 @@ Blockly.DragSurfaceSvg.prototype.scaleWrapper_ = null;
 Blockly.DragSurfaceSvg.prototype.scale_ = 1;
 
 /**
+ * Cached X value for the translation of the drag surface.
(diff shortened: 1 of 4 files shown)


core/dragsurface_svg.js
```diff
@@ -24,6 +24,20 @@ Blockly.DragSurfaceSvg = function(container) {
   this.container_ = container;
 };
 
+/**
+ * Amount the drag surface should grow the blocks on a drag, relative to 1.
+ * @type {Number}
+ * @const
+ */
+Blockly.DragSurfaceSvg.scaleOnDrag = 1.1;
```
tmickel: noticed the BAD_NAMING, I'll fix it

PR #235, opened 2016-04-19
Split the block-positioning logic out of block_svg

### LINKED ISSUE #232: Pants blocks drive blocks with fields upwards
Forking off from #224.

First, a bit of history. Let's cast our minds back to 0169999, a few weeks ago, where the following issue was observed:

jumping original
_downwards jumping in horizontal rendering, 0169999_

When an insertion marker for a pants block was added, it was moved so that its connector (specifically, the one on the left of the bay) lined up with that of the statement block (that it would wrap). In this case, this puts it at the same height of the statement block. When the statement block is rerendered within the pants block, it was moved downwards to fit inside, which is undesirable and causes the issue shown.

This was fixed by d22dc5c, which used `height of statement block - minimum block height` to calculate the required y difference between the two blocks, and adjust the insertion markers position.

Because this code is in block_svg.js rather than block_render_svg_horizontal.js, it created the following side effect:

jumping sideeffect
_upwards jumping in vertical rendering, d22dc5c_

This became worse when c3094d6 started including fields in block height, meaning the side effect is observed in horizontal rendering too:

jumping regression
_upwards jumping in horizontal rendering, c3094d6_

With the current design, this is a horizontal-specific issue, as the vertical grammar offsets the connections. This means no such special-case is necessary: a connection's offset from the top-left corner of the block is constant.

repeats
_design of vertical and horizontal repeat, as of Feb 22_

drigz:
This allows the horizontal renderer to position its blocks differently, to cope with the fact that the tabs on its repeat blocks move to line up. This commit also makes the preview and new block positioning use the same code.

Fixes #232.

### DIFF
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -654,8 +654,9 @@ Blockly.BlockSvg.prototype.getFieldShadowBlock_ = function() {
  * @param {!Blockly.Block} newBlock The block to position - either the first
  *     block in a dragged stack or an insertion marker.
  * @param {!Blockly.Connection} newConnection The connection on the new block's
- *     stack - either a connection on newBlock, or the NEXT_STATEMENT connection
- *     on the stack if the stack's being dropped before another block.
+ *     stack - either a connection on newBlock, or the last NEXT_STATEMENT
+ *     connection on the stack if the stack's being dropped before another
+ *     block.
  * @param {!Blockly.Connection} existingConnection The connection on the
  *     existing block, which newBlock should line up with.
  */
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -647,3 +647,32 @@ Blockly.BlockSvg.prototype.getFieldShadowBlock_ = function() {
 
   return null;
 };
+
+/**
+ * Position an new block correctly, so that it doesn't move the existing block
+ * when connected to it.
+ * @param {Blockly.Connection} newConnection The connection on the new block.
+ * @param {Blockly.Connection} existingConnection The connection on the existing
+ *     block.
+ */
+Blockly.BlockSvg.prototype.positionNewBlock =
+    function(newConnection, existingConnection) {
+  var newBlock = newConnection.sourceBlock_;
+
+  // We only need to position the new block if it's before the existing one,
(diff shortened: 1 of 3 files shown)


core/block_render_svg_horizontal.js
```diff
@@ -647,3 +647,32 @@ Blockly.BlockSvg.prototype.getFieldShadowBlock_ = function() {
 
   return null;
 };
+
+/**
+ * Position an new block correctly, so that it doesn't move the existing block
+ * when connected to it.
+ * @param {Blockly.Block} newBlock The connection on the new block.
```
rachel-fenichel: Type info says it's a block, as does the name, but the comment says it's a connection.  Please fix.

core/block_render_svg_horizontal.js
```diff
@@ -647,3 +647,32 @@ Blockly.BlockSvg.prototype.getFieldShadowBlock_ = function() {
...
+ * @param {Blockly.Connection} newConnection The connection on the new block's
```
rachel-fenichel: {!Blockly.Connection}, and also for the other params--you'll have a bad day if any of them are null.

https://developers.google.com/closure/compiler/docs/js-for-compiler#types

rachel-fenichel: Add a comment explaining why it's the connection on the new block's stack, and not just on the new block itself.

core/block_render_svg_horizontal.js
```diff
@@ -651,11 +651,13 @@ Blockly.BlockSvg.prototype.getFieldShadowBlock_ = function() {
...
  * when connected to it.
- * @param {Blockly.Block} newBlock The connection on the new block.
- * @param {Blockly.Connection} newConnection The connection on the new block's
- *     stack.
- * @param {Blockly.Connection} existingConnection The connection on the existing
- *     block.
+ * @param {!Blockly.Block} newBlock The block to position - either the first
+ *     block in a dragged stack or an insertion marker.
+ * @param {!Blockly.Connection} newConnection The connection on the new block's
+ *     stack - either a connection on newBlock, or the NEXT_STATEMENT connection
```
rachel-fenichel: *or the last NEXT_STATEMENT connection on the stack

PR #236, opened 2016-04-19
Allow moving an insertion marker leftwards

drigz:
If you drag a pants block over another block, then try to drag it left so that it's before the other block, the insertion marker won't move until you drag away and return. This is nothing severe but slightly clunky.

drag left broken
_dragging left before the fix_

The expected behaviour can be achieved by patching updatePreviews to recognise an insertion marker as invalid if localConnection has changed, not just if highlightedConnection has changed.

drag left fixed
_dragging left after the fix_

(the gifs were captured after rebasing onto bugfix/232 for clarity)

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -998,7 +998,7 @@ Blockly.BlockSvg.prototype.updatePreviews = function(closestConnection,
   // Remove an insertion marker if needed.  For Scratch-Blockly we are using
   // grayed-out blocks instead of highlighting the connection; for compatibility
   // with Web Blockly the name "highlightedConnection" will still be used.
-  if (Blockly.highlightedConnection_ &&
+  if (Blockly.highlightedConnection_ && blockly.localConnection_ &&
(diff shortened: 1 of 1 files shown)


core/block_svg.js
```diff
@@ -999,7 +999,8 @@ Blockly.BlockSvg.prototype.updatePreviews = function(closestConnection,
   // grayed-out blocks instead of highlighting the connection; for compatibility
   // with Web Blockly the name "highlightedConnection" will still be used.
   if (Blockly.highlightedConnection_ &&
-      Blockly.highlightedConnection_ != closestConnection) {
+      (Blockly.highlightedConnection_ != closestConnection ||
```
rachel-fenichel, line 1002: Do you also need to check if Blockly.localConnection_ is non-null?

drigz, line 1002: I suspect it's unnecessary but I added it for safety's sake as localConnection_ is dereferenced inside - thanks for the suggestion.

PR #266, opened 2016-04-21
Implement stack glowing

tmickel:
-Slight refactor of glowBlock vs. glowStack
-Implement stack highlight as spec'd
-Correction of colour implementation for vertical
-Move drag shadow to group of surface so the two filters don't interfere
-Add buttons to playground to glow/unglow blocks and stacks

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -95,7 +95,14 @@ Blockly.BlockSvg.prototype.dragStartXY_ = null;
  * @type {boolean}
  * @private
  */
-Blockly.BlockSvg.prototype.isGlowing_ = false;
+Blockly.BlockSvg.prototype.isGlowingBlock_ = false;
+
+/**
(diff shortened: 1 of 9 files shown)


core/block_svg.js
```diff
@@ -174,15 +181,30 @@ Blockly.BlockSvg.prototype.unselect = function() {
...
+Blockly.BlockSvg.prototype.setGlowBlock = function(isGlowingBlock) {
+  this.isGlowingBlock_ = isGlowingBlock;
   this.updateColour();
 };
 
 /**
+ * Glow the stack starting with this block, to highlight it visually as if it's running.
+ * @param {boolean} isGlowingStack Whether the stack starting with this block should glow.
+ */
+Blockly.BlockSvg.prototype.setGlowStack = function(isGlowingStack) {
```
rachel-fenichel, line 196: What if this isn't the first block in a stack?

PR #289, opened 2016-04-26
Avoid some unnecessary calls to resize()

rachel-fenichel:
@picklesrus PTAL for sanity check.

Goal is to speed up flyout opening.

### DIFF
--- a/core/flyout.js
+++ b/core/flyout.js
@@ -565,6 +565,7 @@ Blockly.Flyout.prototype.hide = function() {
  *     Variables and procedures have a custom set of blocks.
  */
 Blockly.Flyout.prototype.show = function(xmlList) {
+  Blockly.Events.disable();
   this.hide();
   // Delete any blocks from a previous showing.
   var blocks = this.workspace_.getTopBlocks(false);
@@ -628,7 +629,6 @@ Blockly.Flyout.prototype.show = function(xmlList) {
       // block.
       child.isInFlyout = true;
     }
-    block.render();
     var root = block.getSvgRoot();
     var blockHW = block.getHeightWidth();
     block.moveBy((this.horizontalLayout_ && this.RTL) ? -cursorX : cursorX, cursorY);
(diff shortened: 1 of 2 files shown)


core/xml.js
```diff
@@ -327,15 +327,19 @@ Blockly.Xml.domToBlock = function(workspace, xmlBlock) {
...
     }
     // Populating the connection database may be defered until after the blocks
-    // have renderend.
-    setTimeout(function() {
-      if (topBlock.workspace) {  // Check that the block hasn't been deleted.
-        topBlock.setConnectionsHidden(false);
-      }
-    }, 1);
+    // have rendered.
+    if (!workspace.isFlyout) {
```
picklesrus, line 331: Don't quite follow this one.  Why does this workspace not apply to flyout?

core/flyout.js
```diff
@@ -565,6 +565,7 @@ Blockly.Flyout.prototype.hide = function() {
  *     Variables and procedures have a custom set of blocks.
  */
 Blockly.Flyout.prototype.show = function(xmlList) {
+  Blockly.Events.disable();
```
tmickel: Hmmmmmmm...I'm not sure whether or not we should actually disable events here. It means we'll need to look up the flyout blocks on change for the scratch-vm instead of just tracking them with the create/delete events. Was this part of the optimization or a separate one?

rachel-fenichel: Good point.  I forgot that we need the create/delete events.

The events I'm trying to avoid are the Move events that come from BlockSvg.moveBy.  Disable/enable are pretty lightweight, so I can just move them closer to the moves that I'm concerned about.

core/xml.js
rachel-fenichel, line 331: We don't need to worry about tracking connection locations of blocks in the flyout.  They'll never be dragged.  The blocks that _are_ dragged from the flyout are created in the main workspace, so this won't apply.

core/flyout.js
```diff
@@ -1022,14 +1021,20 @@ Blockly.Flyout.placeNewBlock_ = function(originBlock, workspace,
  * @private
  */
 Blockly.Flyout.prototype.filterForCapacity_ = function() {
```
tmickel, line 1023: Actually, I think we'd like to take this feature out eventually (limiting block counts). I'll file a bug for that (but seems good for now)

PR #293, opened 2016-04-26
Implement field_number stub and num-pad on mobile

tmickel:
What's done:
- field_number including min, max, precision
- Number pad that shows up on Android, iOS, and other mobile devices
- Using these in the example blocks
- Improvements to rendering of text field sizing

What's not done:
- Selection behavior on both Android and iOS is sometimes annoying. For example, on the second tap on Android, a context menu shows up with useless options like "web search" and a giant Google menu appears on the bottom. To resolve this, might implement a custom selection and cursor in the future. Might already be resolved in our wrapper applications.


core/utils.js
```diff
@@ -641,12 +641,13 @@ Blockly.genUid.soup_ = '!#%()*+,-./:;=?@[]^_`{|}~' +
  * Measure some text using a canvas in-memory.
  * @param {string} fontSize E.g., '10pt'
  * @param {string} fontFamily E.g., 'Arial'
+ * @param {string} fontWeight E.g., '600'
  * @param {string} text The actual text to measure
  * @return {number} Width of the text in px.
  */
-Blockly.measureText = function(fontSize, fontFamily, text) {
+Blockly.measureText = function(fontSize, fontFamily, fontWeight, text) {
```
rachel-fenichel, line 648: prefer changing the signature by adding params to the end, not the middle, I think.

Issue #307, opened 2016-05-02
Fix that code runs when you click a drop-down menu (currently hacked to work in horizontal)


tmickel:
This is very important...

When you tap the drop-down, stacks run in Scratch. This is because the drop-down button is actually a block (we implemented all field chrome as shadow blocks). But, this means click events are registered for it and we run the stack (!) I thought originally this was a problem caused by the focusing of fields, (which was also a problem and fixed by Neil). But now I'm realizing we need to handle this.

tmickel:
The horizontal hack has the unfortunate side-effect that vertical blocks don't trigger click events when you tap their fields.

PR #320, opened 2016-05-04
Hack to fix problem with drop-down clicks

tmickel:
A temporary and highly specific fix for #307.

It appears that when clicking in a shadow, `this` is actually the parent block (something I hope Neil is working on: https://github.com/google/blockly/issues/336). I don't know what to do about this right now, so I added a highly specific hack. It rests on the fact that `e.target` is actually correct here. So we can compare `e.target` to the properties of `this` (checking if it's actually `this`'s SVG path or if it's the image field element on `this`).

Rachel - a quick check for sanity would be helpful :) but it seems to work for now...

@drigz


core/block_svg.js
```diff
@@ -681,7 +681,22 @@ Blockly.BlockSvg.prototype.onMouseDown_ = function(e) {
  * @private
  */
 Blockly.BlockSvg.prototype.onMouseUp_ = function(e) {
-  if (Blockly.dragMode_ != Blockly.DRAG_FREE && !Blockly.WidgetDiv.isVisible()) {
+  // XXX: Hack to fix drop-down clicking issue for Google I/O.
```
rachel-fenichel: Make this highly specific hack a separate function.

core/block_svg.js
```diff
@@ -681,7 +681,22 @@ Blockly.BlockSvg.prototype.onMouseDown_ = function(e) {
...
+  // We cannot just check isShadow, since `this` is the parent block.
+  // See: https://github.com/google/blockly/issues/336
+  // click target is a non-shadow block path
+  var isActualPath = (e.target === this.svgPath_ &&
+    e.target.parentNode === this.getSvgRoot());
+  var isImageField = false;
```
rachel-fenichel: Only need to do this traversal if !isActualPath.  When this is a function, just return early if isActualPath is true.

PR #333, opened 2016-05-11
Vertical rounded corners and tiny positioning fixes

tmickel:
-Remove squaring of corners when a block is in a stack
-Add rounded corners to the right side of vertical blocks
-Remove positioning logic related to the shadow on Blockly blocks

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -136,6 +136,16 @@ Blockly.BlockSvg.BOTTOM_RIGHT_CORNER =
     Blockly.BlockSvg.CORNER_RADIUS + ' 0 0,1 -' +
     Blockly.BlockSvg.CORNER_RADIUS + ',' +
     Blockly.BlockSvg.CORNER_RADIUS;
+
+/**
+ * SVG path for drawing the rounded bottom-left corner.
+ * @const
+ */
+Blockly.BlockSvg.BOTTOM_LEFT_CORNER =
+    'a ' + Blockly.BlockSvg.CORNER_RADIUS + ',' +
+     Blockly.BlockSvg.CORNER_RADIUS + ' 0 0,1 -' +
+     Blockly.BlockSvg.CORNER_RADIUS + ',-' +
+     Blockly.BlockSvg.CORNER_RADIUS;
 /**
  * SVG path for drawing the top-left corner of a statement input.
  * Includes the top notch, a horizontal space, and the rounded inside corner.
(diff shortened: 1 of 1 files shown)


core/block_render_svg_vertical.js
```diff
@@ -473,26 +484,14 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(iconWidth, inputRows) {
   // Should the top and bottom left corners be rounded or square?
   if (this.outputConnection) {
     this.squareTopLeftCorner_ = true;
```
rachel-fenichel, line 499: If you're getting rid of squared off corners, why would the top left ever be squared?

core/block_render_svg_vertical.js
```diff
@@ -613,7 +612,8 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
       cursorX = Math.max(cursorX, inputRows.rightEdge);
       this.width = Math.max(this.width, cursorX);
       steps.push('H', cursorX);
-      steps.push('v', row.height);
+      steps.push(Blockly.BlockSvg.TOP_RIGHT_CORNER);
+      steps.push('v', row.height - Blockly.BlockSvg.CORNER_RADIUS * 2);
```
rachel-fenichel, line 625: Comment that it's *2 because you are stopping before the rounded corner on the bottom as well.

core/block_render_svg_vertical.js
```diff
@@ -636,11 +636,6 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
     } else if (row.type == Blockly.NEXT_STATEMENT) {
       // Nested statement.
       var input = row[0];
-      if (y == 0) {
```
rachel-fenichel, line 639: Removing because you'll never see this case, right?

core/block_render_svg_vertical.js
```diff
@@ -715,15 +711,11 @@ Blockly.BlockSvg.prototype.renderDrawBottom_ =
   }
 
   // Should the bottom-left corner be rounded or square?
```
rachel-fenichel: Remove comment or replace with "draw bottom left corner"

core/block_render_svg_vertical.js
```diff
@@ -715,15 +711,11 @@ Blockly.BlockSvg.prototype.renderDrawBottom_ =
...
-    steps.push('H 0');
-  } else {
-    steps.push('H', Blockly.BlockSvg.CORNER_RADIUS);
-    steps.push('a', Blockly.BlockSvg.CORNER_RADIUS + ',' +
-               Blockly.BlockSvg.CORNER_RADIUS + ' 0 0,1 -' +
-               Blockly.BlockSvg.CORNER_RADIUS + ',-' +
-               Blockly.BlockSvg.CORNER_RADIUS);
-  }
+  steps.push('H', Blockly.BlockSvg.CORNER_RADIUS);
+  steps.push('a', Blockly.BlockSvg.CORNER_RADIUS + ',' +
```
rachel-fenichel: This should be a constant.

core/block_render_svg_vertical.js
tmickel, line 499: It needs to be squared in some situations. The only scenario left currently is when there's a "hat" above the block. But I think there may be additional scenarios when we start implementing blocks with outputs. I'll add a comment here, eventually maybe we should rename the property.

core/block_render_svg_vertical.js
tmickel, line 639: exactly - we will never ever have this

tmickel, line 639: there are also a few other cases in here we will never have (external inputs come to mind) that I think we might want to do a more complete audit of eventually

Issue #276, opened 2016-04-22
Don't do dispose animation on undo


tmickel:
The block should disappear immediately, not animate out as if we dropped it into the trash/toolbox

NeilFraser:
Changed in Blockly's develop branch.

tmickel:
sounds good - thanks, Neil!

PR #352, opened 2016-05-24
Merge upstream may 24

### LINKED ISSUE #341: Using "Delete All Blocks" plays too many sounds at once
To test, load the horizontal playground and click "Sprinkles!"

Then right-click and select "Delete X blocks".

You'll hear a grating noise. This is the sound of every block being deleted, the noise plays once for every deleted block.

It only plays the sound once for each block stack - using "Spaghetti!" (despite saying to delete nearly 4000 blocks) only plays the sound a few times.

tmickel:
Should fix #341 and #276
Thanks @NeilFraser !


tmickel:
Not #274 - #276...

PR #353, opened 2016-05-24
Restore animated drop-down with no collisions

### LINKED ISSUE #299: Re-add drop-down animate-in
We removed it in I/O because on Android it was sometimes flying in from (0, 0). Discussion below.

tmickel:
Fixes #299.

Use the `left` and `top` properties of the drop-down div to avoid clobbering the initial `translate` offset on Android. Also has the benefit that we no longer have to store the initial X and Y positions (as those are stored in `left` and `top`...

### DIFF
--- a/core/dropdowndiv.js
+++ b/core/dropdowndiv.js
@@ -124,8 +124,9 @@ Blockly.DropDownDiv.createDom = function() {
   Blockly.DropDownDiv.DIV_.appendChild(Blockly.DropDownDiv.arrow_);
 
   // Transition animation for transform: translate() and opacity.
-  Blockly.DropDownDiv.DIV_.style.transition = 'transform ' + Blockly.DropDownDiv.ANIMATION_TIME + 's, ' +
-  'opacity ' + Blockly.DropDownDiv.ANIMATION_TIME + 's';
+  Blockly.DropDownDiv.DIV_.style.transition = 'transform ' +
+    Blockly.DropDownDiv.ANIMATION_TIME + 's, ' +
+    'opacity ' + Blockly.DropDownDiv.ANIMATION_TIME + 's';
 };
 
 /**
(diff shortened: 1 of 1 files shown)


core/dropdowndiv.js
```diff
@@ -138,6 +122,10 @@ Blockly.DropDownDiv.createDom = function() {
   Blockly.DropDownDiv.DIV_.appendChild(Blockly.DropDownDiv.content_);
   Blockly.DropDownDiv.arrow_ = goog.dom.createDom('div', 'blocklyDropDownArrow');
   Blockly.DropDownDiv.DIV_.appendChild(Blockly.DropDownDiv.arrow_);
+
+  // Transition animation for transform: translate() and opacity.
+  Blockly.DropDownDiv.DIV_.style.transition = 'transform ' + Blockly.DropDownDiv.ANIMATION_TIME + 's, ' +
```
rachel-fenichel: Eyeballing says this line is probably too long.

core/dropdowndiv.js
```diff
@@ -202,14 +190,19 @@ Blockly.DropDownDiv.show = function(owner, primaryX, primaryY, secondaryX, secon
...
   Blockly.DropDownDiv.arrow_.setAttribute('class',
     metrics.arrowAtTop ? 'blocklyDropDownArrow arrowTop' : 'blocklyDropDownArrow arrowBottom');
-  // First apply initial translation
-  div.style.transform = 'translate(' + metrics.finalX + 'px,' + metrics.finalY + 'px)';
-  // Save for animate out
-  Blockly.DropDownDiv.hideAnimationX_ = metrics.initialX;
-  Blockly.DropDownDiv.hideAnimationY_ = metrics.initialY;
-  // Show the div
+  // First apply initial translation.
+  // We put this in the `left` and `top` properties to avoid colliding with `translate`.
```
rachel-fenichel: Longer comment on what you mean by collision--we might not remember why we did this down the line.

Here's the start of a comment.  Feel free to change it completely.

// When we change 'translate' multiple times in close succession, chrome may choose to wait and apply them all at once.  Since we want the first translation to be immediate and the second to be animated, we saw problems where both translations would be applied after animation was turned on, making the dropdown appear to fly in from (0, 0).  This is a workaround.

rachel-fenichel:
And the benefit that we don't have to keep changing the transition style.

tmickel:
I like your explanation - mostly stolen with a couple of adjustments

PR #355, opened 2016-05-24
Add some vertical blocks and remove some old ones

tmickel:
Add the "add" block; replace the old "if-else" with mutators with our two; add the "equals" boolean block, and add all these to the vertical playground.


tmickel:
screen shot 2016-05-24 at 5 15 10 pm

blocks_vertical/control.js
```diff
@@ -81,6 +81,76 @@ Blockly.Blocks['control_repeat'] = {
...
+    this.jsonInit({
+      "type": "control_if_else",
+      "message0": "if %1 %2 %3 else %4 %5",
+      "args0": [
+        {
+          "type": "input_dummy"
+        },
+        {
+          "type": "input_value",
+          "name": "NAME",
```
rachel-fenichel: Every input has the name "NAME".  Please fix.

blocks_vertical/control.js
```diff
@@ -81,6 +81,76 @@ Blockly.Blocks['control_repeat'] = {
...
+
+Blockly.Blocks['control_if_else'] = {
+  /**
+   * Block for if-else.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "type": "control_if_else",
+      "message0": "if %1 %2 %3 else %4 %5",
```
rachel-fenichel: Too many args?

"message0": "if %1 %2 else %3",
       "args0": [
         {
           "type": "input_value",
           "name": "NAME",
          "check": "Boolean"
        },
        {
          "type": "input_statement",
          "name": "NAME"
        },
        {
          "type": "input_statement",
          "name": "NAME"
        }
      ],

should work

blocks_vertical/control.js
tmickel: yes thanks :)

blocks_vertical/control.js
tmickel: Hmmmm, for some reason that comes out like this:
screen shot 2016-05-24 at 9 17 51 pm
I think we need at least one of the dummy inputs to force the `else` to be separate from rendering the field on the statement input. I've updated it to only have the one.

tmickel: With the one dummy input:
screen shot 2016-05-24 at 9 19 37 pm

rachel-fenichel: Oh I see.  Yes, that looks correct.

PR #362, opened 2016-05-26
Vertical shapes pass

tmickel:
Before:
before
After:
screen shot 2016-05-26 at 5 41 06 pm

This has many changes - probably some things that can be better documented/improved too. I may push a few more commits to clean up some of the messiest areas.

In addition to getting the shapes a lot closer to spec, there is an attempt inside to vertically center all fields within their row. I think it works but haven't tested it extensively.

I've also removed a bunch of old code (e.g., for external inputs, the "puzzle piece" output tab, turning on/off hats, lots of old hard-coded constants).

Things that are clearly broken:
-Reporters/outputs are still very undone.
-In doing this, I decided to not include the notches in the calculated heights of the blocks (just like we did in horizontal) to help with the cases where the notch overlaps the next block. This has messed with the flyout positioning a bit but hopefully shouldn't be too hard to fix.

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -738,8 +738,12 @@ Blockly.BlockSvg.prototype.renderDrawBottom_ = function(steps, connectionsXY,
     if (this.nextConnection.isConnected()) {
       this.nextConnection.tighten_();
     }
-    this.height += Blockly.BlockSvg.NOTCH_HEIGHT;
   }
+  // Always pretend height is extended by NOTCH_HEIGHT, even if there's no next.
+  // This is so blocks with no next connection stretch statement inputs to the
+  // correct size, as if there was a notch. Otherwise the height of a block
+  // with no next doesn't really matter anyway....
+  this.height += Blockly.BlockSvg.NOTCH_HEIGHT;
   // Bottom horizontal line
   steps.push('H', Blockly.BlockSvg.CORNER_RADIUS);
   // Bottom left corner
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -656,7 +656,7 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
       steps.push('h', '-' + Blockly.BlockSvg.STATEMENT_INPUT_INNER_SPACE);
       steps.push(Blockly.BlockSvg.INNER_TOP_LEFT_CORNER);
 
-      steps.push('v', row.height - 2 * Blockly.BlockSvg.CORNER_RADIUS);
+      steps.push('v', row.height - 2 * Blockly.BlockSvg.CORNER_RADIUS - Blockly.BlockSvg.NOTCH_HEIGHT);
 
       steps.push(Blockly.BlockSvg.INNER_BOTTOM_LEFT_CORNER);
       // Bottom notch
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -733,8 +733,8 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
         // If the final input is a statement stack, add a small row underneath.
         // Consecutive statement stacks are also separated by a small divider.
         steps.push(Blockly.BlockSvg.TOP_RIGHT_CORNER);
-        steps.push('v', Blockly.BlockSvg.SEP_SPACE_Y - Blockly.BlockSvg.CORNER_RADIUS);
-        cursorY += Blockly.BlockSvg.SEP_SPACE_Y + Blockly.BlockSvg.CORNER_RADIUS;
+        steps.push('v', Blockly.BlockSvg.MIN_BLOCK_Y - Blockly.BlockSvg.CORNER_RADIUS);
+        cursorY += Blockly.BlockSvg.MIN_BLOCK_Y + Blockly.BlockSvg.CORNER_RADIUS;
       }
     }
     cursorY += row.height;
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -39,22 +39,22 @@ Blockly.BlockSvg.GRID_UNIT = 4;
  * Horizontal space between elements.
  * @const
  */
-Blockly.BlockSvg.SEP_SPACE_X = 10;
+Blockly.BlockSvg.SEP_SPACE_X = 2 * Blockly.BlockSvg.GRID_UNIT;
 /**
  * Vertical space between elements.
  * @const
  */
-Blockly.BlockSvg.SEP_SPACE_Y = 10;
+Blockly.BlockSvg.SEP_SPACE_Y = 2 * Blockly.BlockSvg.GRID_UNIT;
 /**
  * Vertical padding around inline elements.
  * @const
  */
-Blockly.BlockSvg.INLINE_PADDING_Y = 5;
+Blockly.BlockSvg.INLINE_PADDING_Y = 2 * Blockly.BlockSvg.GRID_UNIT;
 /**
  * Minimum height of a block.
  * @const
  */
-Blockly.BlockSvg.MIN_BLOCK_Y = 25;
+Blockly.BlockSvg.MIN_BLOCK_Y = 8 * Blockly.BlockSvg.GRID_UNIT;
 /**
  * Height of horizontal puzzle tab.
  * @const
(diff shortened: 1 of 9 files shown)


jeandrek:
:smiley:

core/block_render_svg_vertical.js
```diff
@@ -39,37 +39,51 @@ Blockly.BlockSvg.GRID_UNIT = 4;
...
+ */
+Blockly.BlockSvg.MIN_STATEMENT_INPUT_HEIGHT = 6 * Blockly.BlockSvg.GRID_UNIT;
+
 /**
- * Width of vertical tab (inc left margin).
+ * Width of vertical notch.
  * @const
  */
-Blockly.BlockSvg.NOTCH_WIDTH = 30;
+Blockly.BlockSvg.NOTCH_WIDTH = 8 * Blockly.BlockSvg.GRID_UNIT;
```
rachel-fenichel, line 85: Missing newline.

core/block_render_svg_vertical.js
```diff
@@ -39,37 +39,51 @@ Blockly.BlockSvg.GRID_UNIT = 4;
...
+/**
+ * Height of vertical notch.
+ * @const
+ */
+Blockly.BlockSvg.NOTCH_HEIGHT = 2 * Blockly.BlockSvg.GRID_UNIT;
```
rachel-fenichel, line 91: Missing newline after this line.

core/block_render_svg_vertical.js
```diff
@@ -83,39 +97,55 @@ Blockly.BlockSvg.CORNER_RADIUS = 4;
...
+Blockly.BlockSvg.START_HAT_HEIGHT = 16;
+
 /**
  * Path of the top hat's curve.
  * @const
  */
-Blockly.BlockSvg.START_HAT_PATH = 'c 30,-' +
-    Blockly.BlockSvg.START_HAT_HEIGHT + ' 70,-' +
-    Blockly.BlockSvg.START_HAT_HEIGHT + ' 100,0';
+Blockly.BlockSvg.START_HAT_PATH = 'c 25,-22 71,-22 96,0';
```
rachel-fenichel, line 121: Missing newline.

core/block_render_svg_vertical.js
```diff
@@ -83,39 +97,55 @@ Blockly.BlockSvg.CORNER_RADIUS = 4;
...
-Blockly.BlockSvg.NOTCH_PATH_LEFT = 'l 6,4 3,0 6,-4';
+Blockly.BlockSvg.NOTCH_PATH_LEFT = (
+  'c 2,0 3,1 4,2 ' +
+  'l 4,4 ' +
+  'c 1,1 2,2 4,2 ' +
+  'h 12 ' +
+  'c 2,0 3,-1 4,-2 ' +
+  'l 4,-4 ' +
+  'c 1,-1 2,-2 4,-2'
+);
```
rachel-fenichel, line 135: Throughout this change: newlines between constants and the next comments.

core/block_render_svg_vertical.js
```diff
@@ -83,39 +97,55 @@ Blockly.BlockSvg.CORNER_RADIUS = 4;
...
+
 /**
- * SVG path for drawing a horizontal puzzle tab from top to bottom.
+ * Amount of padding before the notch, on the left (in LTR).
  * @const
  */
-Blockly.BlockSvg.TAB_PATH_DOWN = 'v 5 c 0,10 -' + Blockly.BlockSvg.TAB_WIDTH +
-    ',-8 -' + Blockly.BlockSvg.TAB_WIDTH + ',7.5 s ' +
-    Blockly.BlockSvg.TAB_WIDTH + ',-2.5 ' + Blockly.BlockSvg.TAB_WIDTH + ',7.5';
+Blockly.BlockSvg.NOTCH_LEFT_PADDING = 3 * Blockly.BlockSvg.GRID_UNIT;
```
rachel-fenichel: Prefer "start" to "left" in anticipation of RTL.

core/block_render_svg_vertical.js
```diff
@@ -340,16 +363,19 @@ Blockly.BlockSvg.prototype.renderFields_ =
     if (!root) {
       continue;
     }
+    // Align all fields vertically by their own height.
```
rachel-fenichel: How does this align all of the fields?

core/flyout.js
```diff
@@ -666,14 +666,10 @@ Blockly.Flyout.prototype.layoutBlocks_ = function(blocks, gaps) {
     }
     var root = block.getSvgRoot();
     var blockHW = block.getHeightWidth();
-    var tab = block.outputConnection ? Blockly.BlockSvg.TAB_WIDTH : 0;
-    if (this.horizontalLayout_) {
```
rachel-fenichel, line 670: Why are you removing the horizontal code?

tmickel, line 670: The key thing is there's no such thing as tab/TAB_WIDTH, unless I'm mistaken...

core/block_render_svg_vertical.js
tmickel: I'll add a better comment :)

core/block_render_svg_vertical.js
tmickel, line 135: Agreed. Just FYI I believe this spacing issue also exists upstream.

core/flyout.js
rachel-fenichel, line 670: Oh wait, you may be a genius.

core/block_render_svg_vertical.js
```diff
@@ -369,18 +395,17 @@ Blockly.BlockSvg.prototype.renderFields_ =
 Blockly.BlockSvg.prototype.renderCompute_ = function(iconWidth) {
   var inputList = this.inputList;
   var inputRows = [];
-  inputRows.rightEdge = iconWidth + Blockly.BlockSvg.SEP_SPACE_X * 2;
-  if (this.previousConnection || this.nextConnection) {
-    inputRows.rightEdge = Math.max(inputRows.rightEdge,
-        Blockly.BlockSvg.NOTCH_WIDTH + Blockly.BlockSvg.SEP_SPACE_X);
-  }
+  // Block will be drawn from 0 (left edge) to rightEdge, in px.
+  inputRows.rightEdge = 0;
```
rachel-fenichel: When does this become nonzero?

core/block_render_svg_vertical.js
```diff
@@ -616,15 +629,14 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
           // Create inline input connection.
           if (this.RTL) {
             connectionX = connectionsXY.x - cursorX -
-                Blockly.BlockSvg.TAB_WIDTH + Blockly.BlockSvg.SEP_SPACE_X +
+                Blockly.BlockSvg.SEP_SPACE_X +
                 input.renderWidth + 1;
```
rachel-fenichel: Move to end of previous line, and same below.

towerofnix:
I'm sure everybody's probably noticed, but I'd like to throw out that there shouldn't be a block slot/notch on the inside of the bottom of the if-else, since you can't place any block after a `stop all scripts`.

tmickel:
@liam4 
That's a good observation. I think this design is not necessarily final. However, the choice to have the notch at the bottom of C-blocks was an intentional one we made in the design phase.

For the 90%+ of cases, a notch does fit there, as in, the block to the right will fit perfectly into the block to the left:
screen shot 2016-05-26 at 8 23 09 pm

As you point out, it's only blocks with no bottom notch (stop all scripts, forever, etc.) that make this look weird. And, we're not sure if maybe this _should_ look like there's a gap in the script - putting a "stop" or another block that doesn't connect at the bottom does mean that the script isn't going to proceed to the next block as usual.

@carljbowman may have more comments. I think we're definitely open to feedback, but we also want to be experimental and trying new designs :)

tmickel:
Another option we might consider is "auto-flattening" the C-block notch for the case where you put in a block that doesn't have a notch at the bottom. We'll hopefully give that a try and think about it more.

towerofnix:
> Another option we might consider is "auto-flattening" the C-block notch for the case where you put in a block that doesn't have a notch at the bottom. We'll hopefully give that a try and think about it more.

+1 for this. I'm slightly biased, because that's what I'm used to, but I think I prefer that best out of your ideas.

tmickel:
@rachel-fenichel I think I fixed most of your comments so far :)

tmickel:
Added a couple of commits to better deal with the positioning of inline inputs.

PR #364, opened 2016-06-01
Feature/pull upstream 6 01

rachel-fenichel:


core/css.js
```diff
@@ -472,13 +472,13 @@ Blockly.Css.CONTENT = [
     'opacity: 0;',
   '}',
 
-  '.blocklyScrollbarKnob {',
-    'fill: $colour_scrollbar;',
+  '.blocklyScrollbarHandle {',
+    'fill: #ccc;',
```
tmickel: Note - this seems like a merge problem overwriting something we put into Scratch-blocks ($coulour_scrollbar*).

core/workspace_svg.js
```diff
@@ -916,6 +917,19 @@ Blockly.WorkspaceSvg.prototype.showContextMenu_ = function(e) {
     }
     Blockly.Events.setGroup(false);
   }
+
+  var deleteOption = {
+    text: deleteList.length == 1 ? Blockly.Msg.DELETE_BLOCK :
```
tmickel, line 922: Is this also? I remember changing this code before. No problem though, we can fix it back after the merge.

core/css.js
rachel-fenichel: Fixed

Issue #361, opened 2016-05-26
"Reset scale" button should reset to startScale


tmickel:
Currently it resets to scale=1.0.

PR #367, opened 2016-06-06
Reset zoom button resets to startScale

tmickel:
For #361.


carlosperate:
This sounds like something that could be merged upstream?

Issue #250, opened 2016-04-20
Insertion markers don't seem to take on size of rendered block in vertical


tmickel:
The height seems to be correct for a block with no internal fields; the width seems to be incorrect even for these. Could this be because I took out rendering of fields for insertion markers? Could we get around this somehow by stealing the metrics from the originating block?
<img width="386" alt="ghosts-vertical" src="https://cloud.githubusercontent.com/assets/120403/14662732/143c8672-0687-11e6-9274-87e0a928e27e.png">

thisandagain:
Resolved in #569

PR #368, opened 2016-06-06
Restore insertion marker fields; set them to display: none

tmickel:
Here is a strawman fix for #250. It reverts #215 and extends the old methodology to vertical blocks.

Basically, the fields are "actually" rendered and then hidden. After thinking about it, I'm pretty sure there's no way around this unless we grab dimensions from the block generating the insertion marker. If you wanted to try implementing that instead, we could do it...

### DIFF
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -522,6 +522,10 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(metrics) {
       }
     }
     if (imageField) {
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
Is the goal for the blocks to be exactly the same size as the originating block?

Agreed that the only way to get proper dimensions without fully rendering is to grab dimensions from the other block.

tmickel:
^ yep, they should be the same size as the originating block. Well....actually, I guess except for statement inputs....

I'll merge for now :)

PR #369, opened 2016-06-06
Initial pass of vertical reporter rendering

tmickel:
Before:
screen shot 2016-06-06 at 5 55 21 pm

After:
screen shot 2016-06-06 at 5 55 38 pm

This is for rendering the blocks, not the fields. Editing the actually fields still makes crazy things happen.
The general method is:
- Determine the shape of the sides by checking the output type.
- Determine the size of that shape by measuring the height of the block (edgeShapeWidth_)
- Skip edgeShapeWidth_ pixels when rendering the top and bottom; render the specialized shape instead.

Also fixes vertical rendering of fields/connections by using `row.height` instead of something wrong...

Still to do:
- Adding extra (dynamic?) padding for the boolean shape.
- Actual implementation of the field editors to match.

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -666,7 +666,16 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
           } else {
             connectionX = connectionsXY.x + cursorX;
           }
-          connectionY = connectionsXY.y + cursorY;
+          // Attempt to center the connection vertically.
+          var connectionYOffset = input.renderHeight / 2;
+          // Read the block which is connected to subtract half its height.
+          if (input.connection.targetConnection) {
+            var sourceBlock = input.connection.targetConnection.getSourceBlock();
+            if (sourceBlock.rendered) {
+              connectionYOffset -= sourceBlock.getHeightWidth().height / 2;
+            }
+          }
+          connectionY = connectionsXY.y + cursorY + connectionYOffset;
           input.connection.moveTo(connectionX, connectionY);
           if (input.connection.isConnected()) {
             input.connection.tighten_();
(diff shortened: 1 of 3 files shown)


core/block_render_svg_vertical.js
```diff
@@ -551,6 +606,19 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(iconWidth, inputRows) {
     inputRows.rightEdge = Math.max(inputRows.rightEdge, 100);
   }
 
+  // Amount of space to skip drawing the top and bottom,
+  // to make room for the left and right to draw shapes (curves or angles).
+  this.edgeShapeWidth_ = 0;
+  this.edgeShape_ = null;
+  if (this.outputConnection) {
+    // Width of the curve/pointy-curve
+    var shape = this.outputConnection.getOutputShape();
```
rachel-fenichel, line 615: Why recompute every time you render?  This shouldn't change.

core/block_render_svg_vertical.js
```diff
@@ -719,6 +806,17 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
     }
     cursorY += row.height;
   }
+  if (this.edgeShape_) {
+    // Draw the right-side edge shape
+    if (this.edgeShape_ === Blockly.Connection.NUMBER) {
+      // Draw a rounded arc
+      steps.push('a ' + this.edgeShapeWidth_ + ' ' + this.edgeShapeWidth_ + ' 0 0 1 0 ' + this.edgeShapeWidth_*2);
```
rachel-fenichel: Spaces before last \* in line

core/block_render_svg_vertical.js
tmickel, line 615: Do you mean the shape specifically? That shouldn't change, although edgeShapeWidth_ could change. I can move the calculation of edgeShape_ somewhere better.

rachel-fenichel, line 615: Yes, just the shape.

Issue #305, opened 2016-04-29
Touch event on PC should open on-screen keyboard


cwillisf:
The convention on a Windows computer with a touch-screen is:
- A **click event** in a user-input field should **not** display an on-screen keyboard, but
- A **touch event** in a user-input field **should** display an on-screen keyboard.

We should probably follow this convention when choosing whether to display (for example) our number pad.

tmickel:
@rachel-fenichel Btw I was curious if you tested this?

rachel-fenichel:
Currently touch events on edge on windows are fairly broken--edge and IE don't implement the touch events API, only pointer events.  I need to go back and test it again but I'm not optimistic right now.

chexbox:
@rachel-fenichel - How long until every company/organisation quits using IE/Edge and Microsoft gives up hope in making a browser?

rachel-fenichel:
@chexbox Please keep your comments constructive.

picklesrus:
This is working now on a surface tablet.

PR #386, opened 2016-06-09
Use bindEvent_ for dropdown buttons

drigz:
This matches the style of the rest of Blockly, and means that the
dropdowns can be tested with Chrome's Device Mode, which was not
possible before due to this bug:
https://bugs.chromium.org/p/chromium/issues/detail?id=133915

This also adds listeners for mousedown/mouseup - this should be better
on PCs with touchscreens (#305), although we'll need a further change to
open the numpad on touch events.


tmickel:
looks good, I was wondering why the touch events weren't working in Chrome Device Mode...

thanks!

Issue #380, opened 2016-06-08
Consider not showing parsed value of field_number in block


tmickel:
Perhaps field_numbers should actually behave more like text inputs.

One problem that appears:
When you type a very long number, it gets parsed to something like "1.2341234143132412e+23". This causes weird issues in editing - at what point does the exponential notation version get shown in the block? After editing? How should the block be sized at the point where it transitions from showing all the digits to the shorter, rounded version?

Another: when a user types ".5", do we always want it to change to "0.5"?

Another: When you start typing a negative number, "-", this evaluates to an invalid number, getting an error state: 

<img width="41" alt="screen shot 2016-06-08 at 12 37 00 pm" src="https://cloud.githubusercontent.com/assets/120403/15902417/c7e93f5c-2d75-11e6-9fb9-412c24f53abd.png">

This could feel like negative numbers are not allowed - not something we want.

PR #387, opened 2016-06-09
Add validation to the numpad

drigz:
This works the same way as upstream - if you close the editor with an
invalid (red-highlighted) input, the editor reverts to the original
input.


tmickel:
Seems like a good improvement for now. I do think we'd like to consider refactoring this a bit for some UX concerns: https://github.com/LLK/scratch-blocks/issues/380

Issue #388, opened 2016-06-09
Include height of notches in block height; special-case it out of statement inputs


tmickel:
The initial implementation doesn't include the height of notches in the block height, so that blocks overlay the bottom of statement inputs. But, this causes some unintended side effects (flyout positioning, etc.). If we update it to include the height of notches and just special-case the height of statements, we'll probably have a better time :)

Issue #385, opened 2016-06-08
Consider conditionally flattening bottom notch


tmickel:
In this situation:
<img width="111" alt="screen shot 2016-06-08 at 6 03 20 pm" src="https://cloud.githubusercontent.com/assets/120403/15912606/5bb900e6-2da3-11e6-871b-5dedd4a4245f.png">

Also in horizontal.

tmickel:
Need more discussion for horizontal.

PR #392, opened 2016-06-10
Fix block heights; conditionally flatten inner statement notch.

tmickel:
#388 and #385.

Fixes positioning in the flyout, too :)

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -808,8 +808,10 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
 
       steps.push(Blockly.BlockSvg.INNER_BOTTOM_LEFT_CORNER);
       // Bottom notch
-      steps.push('h ', Blockly.BlockSvg.STATEMENT_INPUT_INNER_SPACE);
-      steps.push(Blockly.BlockSvg.NOTCH_PATH_LEFT);
+      if (row.statementNotchAtBottom) {
+        steps.push('h ', Blockly.BlockSvg.STATEMENT_INPUT_INNER_SPACE);
+        steps.push(Blockly.BlockSvg.NOTCH_PATH_LEFT);
+      }
       steps.push('H', inputRows.rightEdge);
 
       // Create statement connection.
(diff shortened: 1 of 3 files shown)


core/block_render_svg_vertical.js
```diff
@@ -507,6 +508,26 @@ Blockly.BlockSvg.prototype.renderCompute_ = function(iconWidth) {
     } else {
       input.renderWidth = 0;
     }
+
+    // If the input is a statement input, determine if a notch
+    // should be drawn at the inner bottom of the C.
+    if (input.connection && input.connection.type === Blockly.NEXT_STATEMENT) {
+      // Walk down linkedBlock's nextConnection to determine if
```
rachel-fenichel: This should be a helper function in block.js (maybe getLastBlockInStack()), not implemented inline here.

Looks like the android one is called getLastBlockInSequence()
https://github.com/google/blockly-android/blob/6c4d4376c1390d73d67a13375f72a68dbb341444/blocklylib-core/src/main/java/com/google/blockly/model/Block.java#L568

core/block_render_svg_vertical.js
```diff
@@ -354,6 +354,7 @@ Blockly.BlockSvg.prototype.getHeightWidth = function() {
   if (nextBlock) {
```
rachel-fenichel, line 354: This isn't new, but getHeightWidth is now defined in block_svg.js, block_render_svg_horizontal.js, and block_render_svg_vertical.js.  One of these should be unnecessary.

tmickel, line 354: The one in block_svg should be unnecessary (and overridden).

core/block_render_svg_vertical.js
tmickel: Sounds good.

core/block_render_svg_vertical.js
tmickel, line 354: Removed it.

core/block_render_svg_vertical.js
tmickel: Turns out there is a similar one `lastConnectionInStack_`

tmickel:
Please take another look :) I wasn't sure whether to rename lastConnectionInStack_ to be public...

rachel-fenichel:
Yes, please rename it to public.  Then lgtm.

PR #394, opened 2016-06-10
Add full set of Scratch operators; remove some of Blockly's

tmickel:
-All operators in Scratch 2.0
-Remove blocks/logic.js
-Remove all but the "primitive" numbers from math.js
-Update vertical playground


tmickel:
@rachel-fenichel Just making sure this method seems correct - rather than keeping any of the Blockly blocks for these purposes around.

PR #402, opened 2016-06-13
Draw empty "input shapes" for non-connected value inputs

### LINKED ISSUE #377: Drawing holes for boolean inputs
Compare:
screen shot 2016-06-08 at 12 24 08 pm

To Scratch 2.0:
screen shot 2016-06-08 at 12 25 02 pm

and the spec:
screen shot 2016-06-08 at 12 25 24 pm

We need to somehow draw these "holes" for booleans. Two possible approaches:
1. Create a "boolean shadow" block that just gets put into these blocks and is the right size and shape.
2. When a block has an input with the boolean check type, draw a hole for it like this. This could also be extended to other types if we would ever want that, but would be much more difficult to implement (would we add another path to the SVG??? In Blockly, these are holes, so they're part of the same path)

tmickel:
Previously, nothing was drawn in this case. Now it will look like this:
screen shot 2016-06-13 at 5 38 28 pm

RTL:
screen shot 2016-06-13 at 5 55 17 pm

Not 100% sure that the sizing and positioning is perfect, but I think that's something we can adjust.

Fixes #377. Decided to go for adding the extra shape SVG paths as it's more general, flexible, and closer to Blockly.

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -951,7 +951,8 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
 Blockly.BlockSvg.prototype.renderInputShape_ = function(input, x, y) {
   var inputShape = this.inputShapes_[input.name];
   var inputShapeWidth = 0;
-  if (!input.connection.targetConnection) {
+  // Input shapes are only visibly rendered on non-connected non-insertion-markers.
+  if (!input.connection.targetConnection && !this.isInsertionMarker()) {
     var inputShapeX = 0, inputShapeY = 0;
     // If the input connection is not connected, draw a "hole" shape.
     var inputShapePath = null;
@@ -980,10 +981,7 @@ Blockly.BlockSvg.prototype.renderInputShape_ = function(input, x, y) {
     inputShape.setAttribute('transform',
       'translate(' + inputShapeX + ',' + inputShapeY + ')'
     );
-    // Input shapes are only visibly rendered on non-insertion-markers.
-    if (!this.isInsertionMarker()) {
-      inputShape.setAttribute('style', 'visibility: visible');
-    }
+    inputShape.setAttribute('style', 'visibility: visible');
   } else {
     inputShape.setAttribute('style', 'visibility: hidden');
   }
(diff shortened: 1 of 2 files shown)


tmickel:
Proper Y positioning with larger height blocks:
screen shot 2016-06-13 at 6 01 13 pm

core/block_render_svg_vertical.js
```diff
@@ -857,6 +940,55 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
 };
 
 /**
+ * Render the shape for an input.
+ * If there's a connected block, hide the input shape.
+ * Otherwise, draw and set the position of the input shape.
+ * @param {!Blockly.Input} input Input to be rendered.
+ * @param {Number} x X offset of input.
+ * @param {Number} y Y offset of input.
+ * @return {Number} Amount cursor should be offset (depending on width of render)
```
rachel-fenichel: Can you clarify this comment?  It should definitely say it's in the x direction, but I'm still not really sure what it's saying.

core/block_render_svg_vertical.js
```diff
@@ -857,6 +940,55 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
...
+ */
+Blockly.BlockSvg.prototype.renderInputShape_ = function(input, x, y) {
+  var inputShape = this.inputShapes_[input.name];
+  var inputShapeWidth = 0;
+  // Input shapes are only visibly rendered on non-connected non-insertion-markers.
+  if (!input.connection.targetConnection && !this.isInsertionMarker()) {
```
rachel-fenichel: Flip order of if/else to
if (this.isInsertionMarker() || input.connection.targetConnection) { hide} else { calculate + show }

Optionally, use connection.isConnected() instead of connection.targetConnection.  The latter is public but I wish it weren't.

core/block_svg.js
```diff
@@ -121,8 +121,13 @@ Blockly.BlockSvg.INLINE = -1;
  */
 Blockly.BlockSvg.prototype.initSvg = function() {
   goog.asserts.assert(this.workspace.rendered, 'Workspace is headless.');
+  // "Input shapes" for each input. Used to draw "holes" for unoccupied value inputs.
```
rachel-fenichel: You're using scare quotes on these terms instead of defining them.

rachel-fenichel:
Couple small comments then lgtm.

core/block_render_svg_vertical.js
tmickel: clarified to say it returns the width of the input shape :)

core/block_svg.js
tmickel: good point - I've improved this throughout :)

PR #403, opened 2016-06-13
Color fixes for text drop-down menus

tmickel:
- Arrow color (previously hidden)
- Font color to white with a class
- Fixed block colors

screen shot 2016-06-13 at 6 50 20 pm

### DIFF
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -297,6 +297,7 @@ Blockly.FieldDropdown.prototype.setText = function(text) {
 
   if (this.textElement_) {
     // Update class for dropdown text.
+    // This class is reset every time updateTextNode_ is called.
     this.textElement_.setAttribute('class',
         this.textElement_.getAttribute('class') + ' blocklyDropdownText'
     );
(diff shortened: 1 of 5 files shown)


core/field_dropdown.js
```diff
@@ -300,6 +296,10 @@ Blockly.FieldDropdown.prototype.setText = function(text) {
   this.updateTextNode_();
 
   if (this.textElement_) {
+    // Update class for dropdown text.
```
rachel-fenichel, line 299: Is this ever cleared?  If not, aren't you just adding ' blocklyDropdownText' once for every time the text is changed?

tmickel, line 299: Somehow this is not happening, but I'm not sure why not.

tmickel, line 299: Ok, it's because updateTextNode_ sets the class every time also, in case it becomes (non)-editable. Seems like we should leave as-is?

rachel-fenichel, line 299: Interesting.  Okay, lgtm.

PR #405, opened 2016-06-14
Properly center text in fields based on calculated width

tmickel:
Before:
screen shot 2016-06-13 at 8 04 42 pm

After:
screen shot 2016-06-13 at 8 04 53 pm

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -267,6 +267,16 @@ Blockly.Field.prototype.render_ = function() {
         Blockly.Field.cacheWidths_[key] = width;
       }
     }
+    // Update text centering, based on newly calculated width.
+    var centerTextX = width / 2;
+    // In a text-editing shadow block's field,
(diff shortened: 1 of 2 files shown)


tmickel:
Hmm, need to fix a case for RTL:
screen shot 2016-06-13 at 8 09 07 pm

rachel-fenichel:
Just looking at the screenshots, you might need an extra space before "and wait"--that is, after the dropdown field, if there's a label following.

tmickel:
yep - not sure what's going on with that...

rachel-fenichel:
lgtm for ltr

tmickel:
I've pushed an RTL fix, but I have no idea how it works either :) Arrived at it by flipping around subtractions and minus signs. `text-anchor` causes some weird stuff to happen in RTL...

May try to investigate why this works in the morning.

PR #414, opened 2016-06-15
Insertion markers and replacement markers for value inputs and termin…

### LINKED ISSUE #370: Design - what happens when a block is dragged over an input?
In Scratch 2.0, a white "glow" appears around the old text area/block.
In Blockly, the yellow notch is highlighted.
In Scratch Blocks currently, there are two cases:
- If the input is occupied by a shadow block, an insertion marker replaces the shadow.
- If the input is occupied by another block, the other block is bumped out of the position and replaced by the insertion marker. This appears as a bug, because the other block is never replaced if the drag moves somewhere else.

Some options that seem viable:
- Scratch 2.0 behavior for all cases (whether or not there's a shadow). If we do this, it would be great to have a spec for size/color/blur/etc. of a stroke.
- Bump any block in an input, but replace it back in when the drag moves somewhere else.
- Some mix of the two: e.g., use the insertion marker for a shadow block, but some Scratch 2.0-style stroke for regular blocks in inputs.

### LINKED ISSUE #390: block_svg.js:1168 Uncaught insertionMarkerConnection still connected at the end of disconnectInsertionMarker
Sprinkles seems to trigger this. Just logging the issue so we don't forget :)

### LINKED ISSUE #397: Displacing terminal statement blocks
I'm looking at Blockly [Issue 414](https://github.com/google/blockly/issues/414).  Although that particular issue can be resolved by special casing shadow blocks, the root cause is a change in behaviour made to Blockly to match Scratch.  I'm questioning whether Scratch's behaviour is desirable.

Here is a movie showing Scratch's restricted options when dealing with terminal statement blocks.
  https://youtu.be/Qt33f_J6YxU
Note that this is inconsistent with all other block types, where one can displace an existing block by dragging in a replacement.

If a user drags a terminal statement block over another terminal statement block, it is clear what they want to do.  Can we remove the special exception we added that prevents the user from accomplishing their intended goal?

rachel-fenichel:
…al blocks

Fixes #397 
Fixes #390 
Mebbe fixes #370?

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -296,7 +296,9 @@ Blockly.BlockSvg.terminateDrag_ = function() {
   if (Blockly.dragMode_ == Blockly.DRAG_FREE) {
     // Terminate a drag operation.
     if (selected) {
-      if (Blockly.insertionMarker_) {
+      if (Blockly.replacementMarker_) {
+        Blockly.BlockSvg.removeReplacementMarker();
+      } else if (Blockly.insertionMarker_) {
         Blockly.Events.disable();
         if (Blockly.insertionMarkerConnection_) {
           Blockly.BlockSvg.disconnectInsertionMarker();
@@ -1070,7 +1072,9 @@ Blockly.BlockSvg.prototype.updatePreviews = function(closestConnection,
   // with Web Blockly the name "highlightedConnection" will still be used.
   if (Blockly.highlightedConnection_ &&
       Blockly.highlightedConnection_ != closestConnection) {
-    if (Blockly.insertionMarker_ && Blockly.insertionMarkerConnection_) {
+    if (Blockly.replacementMarker_) {
+      Blockly.BlockSvg.removeReplacementMarker();
+    } else if (Blockly.insertionMarker_ && Blockly.insertionMarkerConnection_) {
       Blockly.BlockSvg.disconnectInsertionMarker();
     }
     // If there's already an insertion marker but it's representing the wrong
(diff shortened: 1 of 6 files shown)


tmickel:
Wow! I'm really excited at how well this works!

core/block_svg.js
```diff
@@ -1131,6 +1118,81 @@ Blockly.BlockSvg.prototype.updatePreviews = function(closestConnection,
...
+  }
+  Blockly.replacementMarker_ = null;
+};
+
+/**
+ * Place and render an insertion marker to indicate what would happen if you
+ * release the drag right now.
+ * @param {Blockly.Connection} localConnection The connection on the dragging
+ *     block.
+ * @param {Blockly.Connection} closestConnection The connnection to pretend to
```
tmickel: I'm a little confused by "pretend to connect to" - the insertion marker is actually connected, right? (but the replacement marker doesn't actually make a connection - "pretend" makes sense to me there).

core/block_svg.js
```diff
@@ -1085,39 +1089,22 @@ Blockly.BlockSvg.prototype.updatePreviews = function(closestConnection,
...
-          insertionMarkerConnection, closestConnection);
-
-      if (insertionMarkerConnection.type == Blockly.PREVIOUS_STATEMENT &&
-          !insertionMarker.nextConnection) {
-        Blockly.bumpedConnection_ = closestConnection.targetConnection;
-      }
-      // Renders insertion marker.
-      insertionMarkerConnection.connect(closestConnection);
-      Blockly.insertionMarkerConnection_ = insertionMarkerConnection;
+    var shouldReplace = (localConnection.type == Blockly.OUTPUT_VALUE ||
```
tmickel, line 1102: Don't strongly care, but might be better to move this logic into its own helper or at least add a comment about the situations where shouldReplace is true.

core/connection.js
```diff
@@ -439,12 +439,8 @@ Blockly.Connection.prototype.isConnectionAllowed = function(candidate) {
...
     }
     case Blockly.OUTPUT_VALUE: {
-      // Don't offer to connect an already connected left (male) value plug to
-      // an available right (female) value plug.
-      if (candidate.targetConnection || this.targetConnection) {
-        return false;
-      }
-      break;
+      // Can't drag an input to an output--you have to move the inferior block.
+      return false;
```
tmickel, line 443: Glad we fixed this as well. Blockly is going to be left as-is for this case, right?

core/block_svg.js
rachel-fenichel: How about "The connection to connect the insertion marker to"?

Split infinitives be damned.

core/connection.js
```diff
@@ -469,9 +465,13 @@ Blockly.Connection.prototype.isConnectionAllowed = function(candidate) {
...
       // Don't let a block with no next connection bump other blocks out of the
-      // stack.
+      // stack.  But covering up a shadow block or stack of shadow blocks is
+      // fine.  Similarly, replacing a terminal statement with another terminal
+      // statement is allowed.
       if (candidate.isConnectedToNonInsertionMarker() &&
-          !this.sourceBlock_.nextConnection) {
+          !this.sourceBlock_.nextConnection &&
+          !candidate.targetBlock().isShadow() &&
+          candidate.targetBlock().nextConnection) {
```
tmickel, line 474: Looks good. I like how it works, too, even with stacks that ends in terminal blocks (I think it's fine that it bumps and allows this situation).

tmickel:
Thanks again - LGTM

core/connection.js
rachel-fenichel, line 443: Yeah, blockly absolutely allows this.  I was surprised that Scratch doesn't--but this is what was causing a lot of the problems on the vertical playground.

core/block_svg.js
rachel-fenichel, line 1102: I'll add a comment.

core/block_svg.js
tmickel: ^ sounds better :)

PR #415, opened 2016-06-15
Fix up vertical text field editor

### LINKED ISSUE #379: Vertical text field editor to spec
Finish implementing:
https://github.com/LLK/scratch-blocks/issues/146
and
https://github.com/LLK/scratch-blocks/issues/147

A couple of the issues:
- Turn off animated growth of editor and font size for vertical blocks. The growth should only happen in horizontal:
- Editor is in the wrong position for vertical blocks:

screen shot 2016-06-08 at 12 32 08 pm

tmickel:
Should fix #379, #146, #147.

Adds a couple of flags to the rendering modules to pin down the difference between vertical and horizontal field editor behavior. Use the flags and properties to create the proper mess of CSS.

The field editing feels almost good after this. There's a weird "lag" between when a character is typed and when the block grows. In Scratch 2.0 this happens in the same frame, which is really nice. Blockly addresses the lag by adding a little bit of space after the typed characters (the field editor is slightly bigger than the text, and it grows after the space is filled in). I wonder how or if we could address the lag? Will raise a new bug for that.

### DIFF
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -252,6 +252,18 @@ Blockly.BlockSvg.FONTSIZE_INITIAL = 12;
 Blockly.BlockSvg.FONTSIZE_FINAL = 14;
 
 /**
+ * Whether text fields are allowed to expand past their truncated block size.
+ * @const{boolean}
+ */
+Blockly.BlockSvg.FIELD_TEXTINPUT_EXPAND_PAST_TRUNCATION = true;
+
+/**
+ * Whether text fields should animate their positioning.
+ * @const{boolean}
+ */
+Blockly.BlockSvg.FIELD_TEXTINPUT_ANIMATE_POSITIONING = true;
+
+/**
  * @param {!Object} first An object containing computed measurements of a
  *    block.
  * @param {!Object} second Another object containing computed measurements of a
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -240,6 +240,18 @@ Blockly.BlockSvg.renderingMetrics_ = null;
 Blockly.BlockSvg.MAX_DISPLAY_LENGTH = 4;
 
 /**
+ * Point size of text field before animation. Must match size in CSS.
+ * See implementation in field_textinput.
+ */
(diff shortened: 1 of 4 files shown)


core/block_render_svg_horizontal.js
```diff
@@ -240,6 +240,30 @@ Blockly.BlockSvg.renderingMetrics_ = null;
 Blockly.BlockSvg.MAX_DISPLAY_LENGTH = 4;
 
 /**
+ * Point size of text field before animation. Must match size in CSS.
+ * See implementation in field_textinput.
+ */
+Blockly.BlockSvg.FONTSIZE_INITIAL = 12;
```
tmickel: I'll rename these actually

core/field_textinput.js
```diff
@@ -300,7 +302,7 @@ Blockly.FieldTextInput.prototype.resizeEditor_ = function() {
     xy.y += 1 * scale;
   }
   if (goog.userAgent.WEBKIT) {
-    xy.x += 0.5;
+    xy.x += 0;
```
tmickel: Ermmm

core/field_textinput.js
```diff
@@ -249,18 +242,26 @@ Blockly.FieldTextInput.prototype.validate_ = function() {
...
+    textWidth += Blockly.FieldTextInput.TEXT_MEASURE_PADDING_MAGIC;
+    textWidth *= scale;
+    width = textWidth;
+  } else {
+    // Set width to (truncated) block size.
+    width = this.sourceBlock_.getHeightWidth().width * scale;
+  }
   // The width must be at least FIELD_WIDTH and at most FIELD_WIDTH_MAX_EDIT
-  var width = Math.max(textWidth, Blockly.BlockSvg.FIELD_WIDTH_MIN_EDIT * scale);
+  width = Math.max(width, Blockly.BlockSvg.FIELD_WIDTH_MIN_EDIT * scale);
```
rachel-fenichel, line 264: Could also use goog.math.clamp
http://www.hobsonassoc.com/scripts/closure/goog/docs/closure_goog_math_math.js.html

(http://imgur.com/gallery/fD5R7Xj)

rachel-fenichel:
lgtm

PR #420, opened 2016-06-16
Add handler for DOM input event to field_textinput

tmickel:
Before:
https://www.dropbox.com/s/j073u9ra4062zil/input-event-before.mov?dl=0

After:
https://www.dropbox.com/s/32ncd54s2avox0g/input-event-after.mov?dl=0

Here's what's going on: the 'keypress' event happens _before_ the value of the HTML input is updated. That's why you can't just resize the block there - the value isn't populated to be read. You could do something crazy where you try to predict the size change from e.keyCode, but that seems like a mess.

Luckily from StackOverflow I learned of this new "input" event. It gets fired right after keypress, but after the value of the input has been updated. By binding it to `onHtmlInputChange_`, the block gets sized appropriately right after keypress, and things look smooth even before keyup. Unfortunately this is not supported by < IE 9, but personally I'm fine with them having a slightly laggier "before" experience.

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -157,6 +157,13 @@ Blockly.FieldTextInput.prototype.showEditor_ = function(opt_quietInput, opt_read
   // Bind to keyPress -- repeatedly resize when holding down a key.
   htmlInput.onKeyPressWrapper_ =
       Blockly.bindEvent_(htmlInput, 'keypress', this, this.onHtmlInputChange_);
+  // For modern browsers (IE 9+, Chrome, Firefox, etc.) that support the
+  // DOM input event, also trigger onHtmlInputChange_ then. The input event
+  // is triggered on keypress but after the value of the text input
+  // has updated, allowing us to resize the block at that time.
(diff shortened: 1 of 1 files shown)


picklesrus:
LGTM.  The video definitely looks better! I'm curious, do we now try all the work on keypress, keyup and input or does getting skipped because we check whether the text has actually changed before we do it?

tmickel:
good question...this would cause `onHtmlInputChange_` to be triggered twice. `setValue` isn't triggered twice, because there's a check that the value has actually changed. But, `resizeEditor_` is definitely triggered twice now. I'm not sure why that is outside the conditional...any thoughts?

Can't immediately find a way to detect if the input event is supported in a browser (in which case we could disable the keyup binding altogether).

picklesrus:
I don't think it needs to be outside the conditional either... but I my current track record of changing when to resize things in blockly is questionable...I'm already on bug #6 :-/

tmickel:
Ok, I moved them inside. Hope it doesn't break everything :)

Issue #425, opened 2016-06-16
Where should reporter blocks' output connections be positioned


tmickel:
Currently, I think they're not positioned in the "logical" place.

Demo: 
https://www.dropbox.com/s/x2vizlc30yi87dk/reporter-output-connection.mov?dl=0

PR #427, opened 2016-06-16
Move actual position of output connection and remove rendering adjustment

tmickel:
This address #425, so blocks with outputs are snapped at their y-half-way-point.


rachel-fenichel:
lgtm.

Related: This change https://github.com/google/blockly/pull/389/files was merged into scratch-blocks, as I recall, but it didn't really do anything in scratch-blocks because the rendering code has diverged significantly.  One of the goals of the change was to make all of the connected blocks move at the same time so we can see the code path better.  This would be good to have in scratch-blocks as well.

tmickel:
Cool, I'll make an issue for that.

Issue #413, opened 2016-06-15
Visual effect for replacement marker


rachel-fenichel:
Replace existing code in highlightForReplacement and highlightShapeForInput.

tmickel:
@carljbowman Here's an issue for what it actually looks like when you "replace" a block in or drop an output into an input. Currently it does the random colors thing. I think we should possibly pick a glow effect and build an SVG filter.

PR #428, opened 2016-06-16
Always update previews for output connections.

tmickel:
Noticed a weird jumpiness in the updates for replacement markers:

before

After this fix:
after

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -1039,6 +1039,11 @@ Blockly.BlockSvg.prototype.handleDragFree_ = function(oldXY, newXY, e) {
     }
   }
 
+  // Always update previews for output connections.
+  if (localConnection && localConnection.type == Blockly.OUTPUT_VALUE) {
+    updatePreviews = true;
+  }
+
   if (updatePreviews) {
     var candidateIsLast = (localConnection == lastOnStack);
     this.updatePreviews(closestConnection, localConnection, radiusConnection,
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -1025,7 +1025,9 @@ Blockly.BlockSvg.prototype.handleDragFree_ = function(oldXY, newXY, e) {
   }
 
   var updatePreviews = true;
-  if (Blockly.localConnection_ && Blockly.highlightedConnection_) {
+  if (localConnection && localConnection.type == Blockly.OUTPUT_VALUE) {
+    updatePreviews = true; // Always update previews for output connections.
+  } else if (Blockly.localConnection_ && Blockly.highlightedConnection_) {
     var xDiff = Blockly.localConnection_.x_ + dxy.x -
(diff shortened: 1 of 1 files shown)


NoMod-Programming:
Just wondering, why is the feedback purple at some times but green at other times?

tmickel:
The colors are completely random right now, as we have to figure out and implement the right visual style (e.g., in Scratch 2.0, the white glow).

rachel-fenichel:
Because of https://github.com/LLK/scratch-blocks/blob/develop/core/block_render_svg_vertical.js#L468 and #413 

:smiley:

rachel-fenichel:
This looks nice but I don't understand why it changes this behaviour.

tmickel:
my remembering of the intention of `updatePreviews` - to "slightly prefer the existing preview over a new preview" - was to keep the connection from moving to a different connection as long as the current connection was connected and the new one wasn't too far away.

The output connections are all very close to each other, much less than `Blockly.CURRENT_CONNECTION_PREFERENCE`, which makes the replacement marker sticky until you get to one pretty far away. Setting `updatePreviews` to just always be true for output connections means it will never get "stuck". Hmm, does that make any more sense?

core/block_svg.js
```diff
@@ -1039,6 +1039,11 @@ Blockly.BlockSvg.prototype.handleDragFree_ = function(oldXY, newXY, e) {
     }
   }
 
+  // Always update previews for output connections.
```
rachel-fenichel: This should be before the if on line 1028, which should turn into an else if

tmickel: fair, updated

core/block_svg.js
```diff
@@ -1025,7 +1025,9 @@ Blockly.BlockSvg.prototype.handleDragFree_ = function(oldXY, newXY, e) {
   }
 
   var updatePreviews = true;
-  if (Blockly.localConnection_ && Blockly.highlightedConnection_) {
+  if (localConnection && localConnection.type == Blockly.OUTPUT_VALUE) {
+    updatePreviews = true; // Always update previews for output connections.
+  } else if (Blockly.localConnection_ && Blockly.highlightedConnection_) {
```
rachel-fenichel, line 1030: ...and in turn, this should probably be 

var updatePreviews = true;
if (!(localConnection && localConnection.type == Blockly.OUTPUT_VALUE) &&
(Blockly.localConnection_ && Blockly.highlightedConnection_)) {

since the first clause is a no-op.

If you want to leave it this way for clarity, that's fine too.

rachel-fenichel:
lgtm

core/block_svg.js
tmickel, line 1030: Haha...actually I started with this version before the pr and decided it was unpleasant. I think let's settle for the current in-between - even though it's a no-op, I think it's least unpleasant.

Issue #423, opened 2016-06-16
String inputs have max width but text isn't truncated


tmickel:
![687474703a2f2f692e6375626575706c6f61642e636f6d2f3374506e49772e706e67](https://cloud.githubusercontent.com/assets/120403/16131901/bd5ab588-33de-11e6-9f63-ec139baee6cc.png)

Rather than truncate the underlying string, I think we should address this in a similar way to how we'll address #380.

PR #429, opened 2016-06-17
Text input restriction and application to FieldNumber

### LINKED ISSUE #380: Consider not showing parsed value of field_number in block

tmickel:
This should fix #380 and #423.

This adds text input "restriction" in addition to validation (validation is now unused). Restriction restricts what can be typed in the field altogether, preventing the "red" input case that reverts to a valid value when finished. The number is never forced to be parsed before being displayed on the block (as the validator did), meaning we can get things like this:
screen shot 2016-06-17 at 11 43 08 am
This has an additional feature that Scratch 2.0 didn't have - `decimalAllowed` and `negativeAllowed`, which I've applied to some obvious cases like repeat loop iterations (which won't allow you to type negative numbers or decimal numbers now).

### DIFF
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -46,7 +46,6 @@ var getNumRestrictor = function(decimalAllowed, negativeAllowed) {
   if (negativeAllowed) {
     pattern += "|[-]";
   }
-  console.log(pattern);
   return new RegExp(pattern);
 };
 
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -48,7 +48,8 @@ Blockly.FieldNumber = function(text, opt_validator, precision, min, max) {
   this.precision_ = precision;
   this.min_ = min;
   this.max_ = max;
-  Blockly.FieldNumber.superClass_.constructor.call(this, text, opt_validator);
+  var numRestrictor = /[\d]|[e]|[-]/; // From Scratch 2.0: restrict except digits, e, and -.
+  Blockly.FieldNumber.superClass_.constructor.call(this, text, opt_validator, numRestrictor);
 };
 goog.inherits(Blockly.FieldNumber, Blockly.FieldTextInput);
 
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -32,23 +32,40 @@ goog.require('goog.userAgent');
 
 
 /**
+ * Return an appropriate restrictor, depending on whether this FieldNumber
+ * allows decimal or negative numbers.
+ * @param {boolean} decimalAllowed Whether number may have decimal/float component.
+ * @param {boolean} negativeAllowed Whether number may be negative.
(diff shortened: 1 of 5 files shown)


core/field_number.js
```diff
@@ -32,23 +32,40 @@ goog.require('goog.userAgent');
...
- * @param {number} max Maximum value of the number.
  * @extends {Blockly.FieldTextInput}
  * @constructor
  */
-Blockly.FieldNumber = function(text, opt_validator, precision, min, max) {
-  this.precision_ = precision;
-  this.min_ = min;
-  this.max_ = max;
-  Blockly.FieldNumber.superClass_.constructor.call(this, text, opt_validator);
+Blockly.FieldNumber = function(text, decimalAllowed, negativeAllowed, opt_validator) {
```
RoboErikG, line 64: On the Android side we've implemented the [extra JSON parameters](https://developers.google.com/blockly/guides/create-custom-blocks/define-blocks#number_field) min, max, and precision to support these cases. This also allows for numbers that step by more than 1 at a time or have a specific range.

Would you mind using those instead of creating new ones?

tmickel, line 64: How would you implement typing restrictions for min, max, and precision?

tmickel, line 64: I suppose we could map precision=0 -> decimalAllowed=false, and min>=0 -> negativeAllowed=false. But the code is really not dealing with min/max/precision anymore, and I'm not sure we ever have a use case for those.

RoboErikG, line 64: You can still use the input restrictions by checking if min && min>=0 for '-' and precision && Math.floor(precision) == precision for '.'. To fully support range and precision they would need to be checked by the validator, but starting with the same properties will make it easier to integrate when we add the full support to Blockly.

tmickel, line 64: If you look at my previous implementation, I used to have this, actually, including a validator that kept it within min, max, and precision. But, I specifically decided to remove it for #380. I could use the same arguments (min, max, precision) if it'll actually make it easier, but I don't ever want this to be validated.

RoboErikG, line 64: Some straw man examples where range will be useful: Robots like the WeDo that take a speed and Music note selection.

PR #435, opened 2016-06-21
Remove `maxBlocks` (fix #309)

### LINKED ISSUE #309: Remove `maxBlocks` feature
Demo has been removed as per #304 but the feature itself should be removed.

tmickel:
Resolves #309

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -780,9 +780,6 @@ Blockly.BlockSvg.prototype.showContextMenu_ = function(e) {
         Blockly.duplicate_(block);
       }
     };
-    if (this.getDescendants().length > this.workspace.remainingCapacity()) {
-      duplicateOption.enabled = false;
-    }
     menuOptions.push(duplicateOption);
 
     if (this.isEditable() && this.workspace.options.comments) {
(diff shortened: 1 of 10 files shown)


RoboErikG:
This is an example of something you might want to support for other devs. Blockly Games uses this for guided lessons like level 4 of the Maze: https://blockly-games.appspot.com/maze?lang=en&level=4&skin=0

tmickel:
@thisandagain thoughts?

thisandagain:
`maxBlocks` runs counter to some of our [design and pedagogical sensibilities](https://scratch.mit.edu/developers) in that it's primary use case is to limit the number of viable options for solving a problem (e.g. Code.org or your example from Blockly Games). While we understand the reasoning for this type of approach, we disagree with this emphasis on "correctness" and instead prefer to put emphasis on providing the user with many pathways and creative possibilities for how they may go about a given task.

RoboErikG:
So it sounds like you want to discourage other devs from using that approach, correct?

tmickel:
At least we aren't going to maintain code that specifically is for that approach, I think.

PR #436, opened 2016-06-21
Horizontal width calculation and notch flattening

tmickel:
This is the horizontal version of #392, resolving #385.

I'm noting that horizontal rendering is full of some unexplained constants/things that should be refactored at some point.

horizontal-notch


core/block_render_svg_horizontal.js
```diff
@@ -741,6 +751,9 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps, connectionsXY, met
   if (this.nextConnection) {
     steps.push(Blockly.BlockSvg.NOTCH_PATH_UP);
 
+    // Include width of notch in block height.
```
rachel-fenichel: width of notch in block _height_?

rachel-fenichel:
One confusing comment then lgtm

PR #437, opened 2016-06-21
Replacement marker glow similar to Scratch 2.0

tmickel:
@carljbowman 

Here is a video:
https://www.dropbox.com/s/0hi38lxj905vu7i/replacement-marker.mov?dl=0

Let me know what you think...


tmickel:
Updated the radius - up for code review :)

PR #447, opened 2016-06-23
Separate output edge shapes from output types

### LINKED ISSUE #442: Determine output edge shape with a separate property from type
It's becoming clear that this will likely be necessary.

tmickel:
Resolve #442

### DIFF
--- a/blocks_vertical/control.js
+++ b/blocks_vertical/control.js
@@ -61,8 +61,7 @@ Blockly.Blocks['control_repeat'] = {
       "args0": [
         {
           "type": "input_value",
-          "name": "TIMES",
-          "check": "Number"
+          "name": "TIMES"
         },
         {
           "type": "input_statement",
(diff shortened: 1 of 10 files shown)


blocks_vertical/control.js
```diff
@@ -61,8 +61,7 @@ Blockly.Blocks['control_repeat'] = {
       "args0": [
         {
           "type": "input_value",
-          "name": "TIMES",
```
rachel-fenichel, line 64: But this still can't take a boolean, right?  So it has to have a check, but should be expanded to string || number

tmickel, line 64: In 2.0 it does, I was just making it match for now. We're having lots of questions around here about what should snap into where, but I figured it would make a good starting point to have us match the current behavior :)

rachel-fenichel, line 64: Oh right, and it casts to a number.  Gotcha.

Issue #260, opened 2016-04-20
Support check-boxes for reporters in flyout


tmickel:
The flyout should allow blocks to have a check-box property. If this is set, the block is shown with a checkbox to the left of it. When the checkbox is checked or unchecked, a UI event is generated so Scratch can show/hide a value monitor. The check-box should have state on initiation (some projects will have this checked in the saved state).

<img width="148" alt="screen shot 2016-04-20 at 11 54 24 am" src="https://cloud.githubusercontent.com/assets/120403/14681358/bac359f0-06ee-11e6-8e31-bad894644485.png">

rachel-fenichel:
Is this also expected to work in horizontal mode?

tmickel:
Horizontal flyout, I'm assuming (not horizontal blocks, which I think should be irrelevant),

Not clear if we'll ever be using horizontal flyout on the vertical blocks. If it's easy enough to place the checkbox above the block in "top" mode and below the block in "bottom" mode, I think that would be nice :). But definitely not the highest priority.

cpseager:
The checkbox should also update if any drop down within the block on the flyout changes - Scratch 2 is flawed in this regard (with the checkbox checked state getting out of sync with the reporters that are actually currently displayed).

This is easiest seen in Scratch 2 with the 'current minute/month/date' etc. block - check the box and then change the drop down to a different type - the checkbox does not update as it should on the parameter change.

tmickel:
Perhaps let's have the state of the checkbox flow from the VM. When you click it trigger a UI event. And have a some method the VM can use to check/uncheck the checkbox?

PR #452, opened 2016-06-23
Checkboxes in the flyout

rachel-fenichel:
Open questions:
What event should be fired on click?  Currently it just logs the id of the block
What field on the block says that a checkbox is necessary?  Right now they're in front of statement blocks.


tmickel:
> What event should be fired on click? Currently it just logs the id of the block

Maybe can we do a UI event that has the block ID and some name, similar to the category switch event?

> What field on the block says that a checkbox is necessary? Right now they're in front of statement blocks.

Something like `hasReporterCheckbox` perhaps?

core/css.js
```diff
@@ -952,5 +952,9 @@ Blockly.Css.CONTENT = [
     'padding: 0;',
   '}',
 
+  '.blocklyFlyoutCheckbox {',
+    'fill: red',
+  '}',
```
tmickel, line 957: is the end goal to have a separate class for when it's checked vs. not checked?

core/flyout.js
```diff
@@ -684,6 +720,44 @@ Blockly.Flyout.prototype.layoutBlocks_ = function(blocks, gaps) {
 };
 
 /**
+ * Create and place a checkbox corresponding to the given block.
+ * @param {!Blockly.Block} block The block to associate the checkbox to.
+ * @param {number} cursorX The x position of the cursor during this layout pass.
+ * @param {number} cursorY The y position of the cursor during this layout pass.
+ * @param {!{height: number, width: number}} blockHW The height and width of the
+ *     block.
+ * @param {Element} root The svg root of the block.
```
tmickel: why pass in these properties (blockHW, root) when you pass in `block`?

core/flyout.js
```diff
@@ -684,6 +720,44 @@ Blockly.Flyout.prototype.layoutBlocks_ = function(blocks, gaps) {
...
+Blockly.Flyout.prototype.createCheckbox_ =
+    function(block, cursorX, cursorY, blockHW, root) {
+  /* eslint-disable indent */
+  var extraSpace = this.CHECKBOX_SIZE + this.CHECKBOX_MARGIN;
+  var checkboxRect = Blockly.createSvgElement('rect',
+    {
+      'class': 'blocklyFlyoutCheckbox',
+      'height': this.CHECKBOX_SIZE,
+      'width': this.CHECKBOX_SIZE,
+      'x': this.RTL ? cursorX + blockHW.width + extraSpace :
```
tmickel, line 743: Hmmm, something seems odd in RTL:
screen shot 2016-06-24 at 11 35 57 am
Thoughts?

core/flyout.js
```diff
@@ -684,6 +720,44 @@ Blockly.Flyout.prototype.layoutBlocks_ = function(blocks, gaps) {
...
+ * @param {number} cursorX The x position of the cursor during this layout pass.
+ * @param {number} cursorY The y position of the cursor during this layout pass.
+ * @param {!{height: number, width: number}} blockHW The height and width of the
+ *     block.
+ * @param {Element} root The svg root of the block.
+ * @private
+ */
+Blockly.Flyout.prototype.createCheckbox_ =
+    function(block, cursorX, cursorY, blockHW, root) {
+  /* eslint-disable indent */
```
tmickel: I'm wondering why you choose to disable eslint vs. have some of the properties come on the next line.

core/flyout.js
```diff
@@ -730,6 +810,11 @@ Blockly.Flyout.prototype.addBlockListeners_ = function(root, block, rect) {
       block.addSelect));
   this.listeners_.push(Blockly.bindEvent_(rect, 'mouseout', block,
       block.removeSelect));
+
+  if (block.flyoutCheckbox) {
+    this.listeners_.push(Blockly.bindEvent_(block.flyoutCheckbox, 'mousedown', null,
```
tmickel: touch events?

core/flyout.js
```diff
@@ -130,6 +137,29 @@ Blockly.Flyout.prototype.MARGIN = 8;
 Blockly.Flyout.prototype.SCROLLBAR_PADDING = 2;
 
 /**
+ * Size of a checkbox next to a variable reporter.
+ * @type {number}
+ * @const
+ */
+Blockly.Flyout.prototype.CHECKBOX_SIZE = 20;
```
tmickel, line 146: Can this be set and calculated in CSS, or is it necessary for layout that it's defined in the JS? It comes to mind that we could want to have the checkbox a different size in touch environments from desktop environments...I wonder how we should deal with that if it comes up.

tmickel:
Also wondering about the way to switch a checkbox to checked/unchecked from the outside. Do you want to save that for a separate PR?

core/flyout.js
rachel-fenichel, line 743: Aha, that's probably what was stopping me from sending out this PR.  Will take a look.

core/flyout.js
rachel-fenichel: What about them?  This is the same way that we add listeners on the blocks in the flyout.

core/flyout.js
rachel-fenichel: Good point.  Will fix.

core/flyout.js
tmickel: so you're saying it should just work? I haven't tried it, I just saw 'mousedown' and wondered :)

tmickel: yep, looks to me like bindEvent_ deals with that. sounds good

core/flyout.js
rachel-fenichel, line 146: That sounds like a bigger question.  Leaving it this way for now.

core/flyout.js
rachel-fenichel: GetHeightWidth() is not cheap--it walks the tree of blocks.  The results are needed here and in the caller function.

root has a simple getter.  I can get it in here instead.

rachel-fenichel:
Setting a checkbox to checked or unchecked from the outside--I think the way to do it is to call flyout.setCheckboxState(blockId, state), which then 
1) gets the block by id and 
2) checks or unchecks the checkbox by calling flyout.checkboxClicked, setting the class [that I haven't created yet] on the associated checkbox.  The blocks have a reference so that's easy.

If that sounds reasonable, I can add it now.  Unclear when we'll have callers though.

tmickel:
^ sounds great to me. If you want a way to test it, I guess we could add a button to the playground that randomly checks/unchecks the checkbox on all the blocks or something.

rachel-fenichel:
Updated with setCheckboxState, better tracking of checkboxes and their blocks, and proper handling of RTL and zoom.

tmickel:
Looks good to me, except it seems checking the checkbox causes the page to get text selected:
jun-27-2016 23-09-42

And I'd like to add the JSON property/option for setting the checkbox asap :)

rachel-fenichel:
#260

PR #467, opened 2016-07-03
Add "When key pressed" block from Scratch 2.0

ghost:
At the moment, the keys O,p,q,r,s,t,u,v,w,x,y,z and Numbers is not added (The rest is ok)
Note: A Fix is needed for the Menu's Arrow
Edit: All Keys from Scratch 2.0 Added


blocks_vertical/event.js
```diff
@@ -139,3 +139,94 @@ Blockly.Blocks['event_broadcastandwait'] = {
     });
   }
 };
+
+Blockly.Blocks['event_keyoptions'] = {
+  /**
+   * Block for when broadcast received.
```
tmickel: This comment looks out-of-date to me.

blocks_vertical/event.js
```diff
@@ -139,3 +139,94 @@ Blockly.Blocks['event_broadcastandwait'] = {
...
+      "colour": Blockly.Colours.event.primary,
+      "colourSecondary": Blockly.Colours.event.secondary,
+      "colourTertiary": Blockly.Colours.event.tertiary
+    });
+  }
+};
+
+Blockly.Blocks['event_whenkeypressed'] = {
+  /**
+   * Block to send a broadcast.
```
tmickel, line 211: Same here.

tmickel:
Thanks for the contribution, @joaoh1! I had two small comments, but other than those I think we'd be happy to merge this in.

PR #475, opened 2016-07-05
Add Pen Blocks and Sound Blocks

ghost:
The Blocks that i added is:
"Clear" Block
"Stamp" Block
"Pen Down" Block
"Pen Up" Block
"Change Pen Color by (Number)" Block
"Set Pen Color to (Number)" Block
"Change Pen Shade to (Number)" Block
"Set Pen Shade to (Number)" Block
"Change Pen Size to (Number)" Block
"Set Pen Size to (Number)" Block
All Sound Blocks

The only Block that's not added is:
"Set Pen Color to (Color)" Block

And the Features not included is:
Dropdown for Notes,Instruments and Drums

### DIFF
--- a/blocks_vertical/sound.js
+++ b/blocks_vertical/sound.js
@@ -1,0 +1,332 @@
+'use strict';
+
+goog.provide('Blockly.Blocks.sound');
+
+goog.require('Blockly.Blocks');
+
+goog.require('Blockly.Colours');
+
+goog.require('Blockly.constants');
+
+Blockly.Blocks['sound_sounds_option'] = {
+  /**
+   * Sound effects drop-down menu.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit(
+      {
+        "message0": "%1",
+        "args0": [
+          {
+            "type": "field_dropdown",
+            "name": "SOUND_MENU",
+            "options": [
+              ['meow', 'MEOW'],
+              ['meow2', 'MEOW2'],
+              ['pop', 'POP']			  
+            ]
+          }
+        ],
+        "inputsInline": true,
+        "output": "String",
+        "colour": Blockly.Colours.sounds.secondary,
+        "colourSecondary": Blockly.Colours.sounds.secondary,
+        "colourTertiary": Blockly.Colours.sounds.tertiary,
+        "outputShape": Blockly.OUTPUT_SHAPE_ROUND
+      });
+  }
+};
+
+Blockly.Blocks['sound_play'] = {
+  /**
+   * Block to play sound.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": "play sound %1",
+      "args0": [
+        {
+          "type": "input_value",
+          "name": "SOUND_MENU"
+        }
+      ],
+      "inputsInline": true,
+      "previousStatement": null,
+      "nextStatement": null,
+      "colour": Blockly.Colours.sounds.primary,
+      "colourSecondary": Blockly.Colours.sounds.secondary,
+      "colourTertiary": Blockly.Colours.sounds.tertiary
+    });
+  }
+};
+
+Blockly.Blocks['sound_playuntildone'] = {
+  /**
+   * Block to play sound until done.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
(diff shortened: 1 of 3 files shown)


ghost:
I Added the Entire Sound Category (But without their Special Dropdowns)

tmickel:
This looks awesome! Thank you @joaoh1!!

tests/vertical_playground.html
```diff
@@ -435,7 +435,7 @@
         <block type="pen_change_pen_size_by">
           <value name="SIZE">
             <shadow type="math_number">
-              <field name="SHADE">1</field>
+              <field name="SIZE">1</field>
```
tmickel: I'm guessing you might mean `<field name="NUM">`? Although I haven't tried it.

tests/vertical_playground.html
```diff
@@ -396,6 +399,54 @@
...
       </category>
+      <category name="Pen">
+        <block type="pen_clear"></block>
+        <block type="pen_stamp"></block>
+        <block type="pen_penup"></block>
+        <block type="pen_pendown"></block>
+        <block type="pen_change_pen_color_by">
+          <value name="COLOR">
+            <shadow type="math_number">
+              <field name="COLOR">10</field>
```
tmickel: All of these fields inside the `math_number` shadows should be `<field name="NUM">`, I think.

blocks_vertical/sound.js
```diff
@@ -0,0 +1,332 @@
...
+};
+
+Blockly.Blocks['sound_playnoteforbeats'] = {
+  /**
+   * Block to play a certain note for certain beats
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": "play drum %1 to %2 beats",
```
tmickel: I think this text should say "play note %1 for %2 beats"

blocks_vertical/sound.js
```diff
@@ -0,0 +1,332 @@
...
+};
+
+Blockly.Blocks['sound_playdrumforbeats'] = {
+  /**
+   * Block to play a certain drum for certain beats
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": "play drum %1 to %2 beats",
```
tmickel: and this "play drum %1 for %2 beats" (not to)

tests/vertical_playground.html
```diff
@@ -396,6 +399,54 @@
           </value>
         </block>
       </category>
+      <category name="Pen">
+        <block type="pen_clear"></block>
+        <block type="pen_stamp"></block>
+        <block type="pen_penup"></block>
+        <block type="pen_pendown"></block>
```
tmickel, line 405: Not important, but we put "pen down" before "pen up" since you usually have to use that one first :)

tests/vertical_playground.html
```diff
@@ -396,6 +399,54 @@
...
+        <block type="pen_change_pen_color_by">
+          <value name="COLOR">
+            <shadow type="math_number">
+              <field name="NUM">10</field>
+            </shadow>
+          </value>
+		</block>   
+        <block type="pen_set_pen_color_to">
```
tmickel, line 414: and in 2.0, "set pen color" comes before "change pen color"

tmickel:
A couple of other minor comments, but looks good to me. @thisandagain ?

ghost:
@tmickel, actually, "Set Pen Color to [Color]" is not added, and "Set Pen Color to [Number]" is after "Change Pen Color by [Number]"

tmickel:
I see, makes sense!

ghost:
Can i include Variables at this Pull Request?

tmickel:
@joaoh1 do you mean variable blocks? I think you're welcome to make some of those, but probably best to put it in a separate PR :) Also - we have some design work to do for variable blocks, so as a warning, your work might get overwritten.

blocks_vertical/sound.js
thisandagain: play **note** %1 **for** %2 beats

thisandagain:
This looks great. Thank you for sending this along. If you don't mind, I'd prefer to merge this as-is with the few minor adjustments we have noted. If we could isolate each PR to a single block category it makes it much easier for us to review & merge.

thisandagain:
This is great. Thanks! 👍

Issue #451, opened 2016-06-23
Input shape paths not getting deleted


rachel-fenichel:
Found this state after 2x sprinkles + fakeManyDrags.  Cannot replicate on a clean instance; can replicate by dragging blocks in from the flyout in this instance.

![image](https://cloud.githubusercontent.com/assets/13686399/16322993/31522b1a-395c-11e6-89c7-f51f19bd465a.png)

svg:

```
[...]

PR #479, opened 2016-07-06
Fix duplicate input shapes #451

tmickel:


rachel-fenichel:
LGTM, but did you figure out why initInputShapes was being called more than once for an input?

tmickel:
initSvg is called for blocks sometimes more than once. Lots of things seem to call it: https://github.com/LLK/scratch-blocks/search?utf8=%E2%9C%93&q=initSvg

PR #480, opened 2016-07-06
Resolve #454 by causing bump for all cases

### LINKED ISSUE #454: Choose behavior of replacing blocks with compatible inputs
As we talked about with Brian today.

Currently in Scratch 2.0, when you replace a block in that has a compatible input, it bumps out.

jun-27-2016 23-15-09

In scratch-blocks currently, it will usually bump out, except if there is exactly one compatible input, in which case it "accepts the input:"

jun-27-2016 23-13-31

We should likely either switch to Scratch 2.0 behavior, or go the full way with the second behavior, and accept it anytime there is at least one compatible input (e.g., replacing in <_ and _> could put the pre-existing predicate in the first input).

tmickel:
There seem to be no tests for this case, so no tests are broken. Also removed `Blockly.Connection.lastConnectionInRow_` since this was the only use.

### DIFF
--- a/core/connection.js
+++ b/core/connection.js
@@ -164,12 +164,6 @@ Blockly.Connection.prototype.connect_ = function(childConnection) {
       shadowDom = Blockly.Xml.blockToDom(orphanBlock);
       orphanBlock.dispose();
       orphanBlock = null;
-    } else if (parentConnection.type == Blockly.INPUT_VALUE) {
-      // Value connections.
-      // If female block is already connected, disconnect and bump the male.
-      if (!orphanBlock.outputConnection) {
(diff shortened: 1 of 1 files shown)


core/connection.js
```diff
@@ -170,15 +170,6 @@ Blockly.Connection.prototype.connect_ = function(childConnection) {
       if (!orphanBlock.outputConnection) {
```
rachel-fenichel: This check/exception is a guard for the next few lines of code.  It shouldn't be necessary now.  In fact, I don't think you need the if (parentConnection.type == Blockly.INPUT_VALUE) branch at all.

tmickel: ah yes-that was silly

PR #482, opened 2016-07-06
Basic fixes for FieldColour

### LINKED ISSUE #163: Colour field needs refactor to not use borderRect_ (or re-add it).
borderRect_ was removed in a refactor, since we are drawing field boundaries in block_render_horizontal/vertical and not otherwise using this.

The colour field was using borderRect_ to show the selected color; instead we should color the fill of the shadow block (or do something else depending on how our color picker is designed).

tmickel:
Fixes #163 by creating a new rect called colourRect_ specifically for the colour field.

And fix sizing to make more sense for our blocks. I am not a fan of using NBSPs to force the width, so fixed it to use some constants instead.

We need to do a lot with the picker chrome, but at least this gives us a placeholder to start playing with values and the relevant blocks.

screen shot 2016-07-06 at 6 29 54 pm

cc: @carljbowman

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -677,7 +677,6 @@ Blockly.Css.CONTENT = [
   '}',
 
   '.blocklyWidgetDiv .goog-palette-table {',
-    'border: 1px solid #666;',
     'border-collapse: collapse;',
   '}',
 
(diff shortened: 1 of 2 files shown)


tmickel:
Now:
screen shot 2016-07-06 at 6 35 00 pm
:)

PR #493, opened 2016-07-07
Fix angle picker: use drop-down div, Scratch-style direction

tmickel:
Again need to do design on this, but at least it lets us start using the angle field where appropriate.
screen shot 2016-07-07 at 12 52 53 pm

### DIFF
--- a/blocks/math.js
+++ b/blocks/math.js
@@ -32,32 +32,24 @@ goog.require('Blockly.Colours');
 
 goog.require('Blockly.constants');
 
-
-/**
- * Common HSV hue for all blocks in this category.
- */
-Blockly.Blocks.math.HUE = Blockly.Colours.textField;
-
 Blockly.Blocks['math_number'] = {
   /**
    * Block for generic numeric value.
    * @this Blockly.Block
    */
   init: function() {
-    this.setHelpUrl(Blockly.Msg.MATH_NUMBER_HELPURL);
-    this.setColour(Blockly.Blocks.math.HUE);
-    this.appendDummyInput()
-        .appendField(new Blockly.FieldNumber('0'), 'NUM');
-    this.setOutput(true, 'Number');
-    // Assign 'this' to a variable for use in the tooltip closure below.
-    var thisBlock = this;
-    // Number block is trivial.  Use tooltip of parent block if it exists.
-    this.setTooltip(function() {
-      var parent = thisBlock.getParent();
-      return (parent && parent.getInputsInline() && parent.tooltip) ||
-          Blockly.Msg.MATH_NUMBER_TOOLTIP;
+    this.jsonInit({
+      "message0": "%1",
+      "args0": [
+        {
+          "type": "field_number",
+          "name": "NUM"
+        }
+      ],
+      "output": "Number",
+      "outputShape": Blockly.OUTPUT_SHAPE_ROUND,
+      "colour": Blockly.Colours.textField
     });
-    this.setOutputShape(Blockly.OUTPUT_SHAPE_ROUND);
   }
 };
 
@@ -67,12 +59,20 @@ Blockly.Blocks['math_whole_number'] = {
    * @this Blockly.Block
    */
   init: function() {
-    this.setHelpUrl(Blockly.Msg.MATH_NUMBER_HELPURL);
-    this.setColour(Blockly.Blocks.math.HUE);
-    this.appendDummyInput()
-        .appendField(new Blockly.FieldNumber('0', 0, Infinity, 1), 'NUM');
-    this.setOutput(true, 'Number');
-    this.setOutputShape(Blockly.OUTPUT_SHAPE_ROUND);
+    this.jsonInit({
+      "message0": "%1",
+      "args0": [
+        {
+          "type": "field_number",
+          "name": "NUM",
+          "min": 0,
+          "precision": 1
+        }
+      ],
+      "output": "Number",
+      "outputShape": Blockly.OUTPUT_SHAPE_ROUND,
+      "colour": Blockly.Colours.textField
+    });
   }
 };
 
@@ -82,12 +82,19 @@ Blockly.Blocks['math_positive_number'] = {
    * @this Blockly.Block
    */
   init: function() {
-    this.setHelpUrl(Blockly.Msg.MATH_NUMBER_HELPURL);
-    this.setColour(Blockly.Blocks.math.HUE);
-    this.appendDummyInput()
-        .appendField(new Blockly.FieldNumber('0', 0, Infinity), 'NUM');
-    this.setOutput(true, 'Number');
-    this.setOutputShape(Blockly.OUTPUT_SHAPE_ROUND);
+    this.jsonInit({
+      "message0": "%1",
+      "args0": [
+        {
+          "type": "field_number",
+          "name": "NUM",
+          "min": 0
+        }
+      ],
+      "output": "Number",
+      "outputShape": Blockly.OUTPUT_SHAPE_ROUND,
+      "colour": Blockly.Colours.textField
+    });
   }
 };
 
@@ -97,11 +104,18 @@ Blockly.Blocks['math_angle'] = {
    * @this Blockly.Block
    */
   init: function() {
-    this.setHelpUrl(Blockly.Msg.MATH_NUMBER_HELPURL);
-    this.setColour(Blockly.Blocks.math.HUE);
-    this.appendDummyInput()
-        .appendField(new Blockly.FieldAngle('90'), 'NUM');
-    this.setOutput(true, 'Number');
-    this.setOutputShape(Blockly.OUTPUT_SHAPE_ROUND);
+    this.jsonInit({
+      "message0": "%1",
+      "args0": [
+        {
+          "type": "field_angle",
+          "name": "NUM",
+          "value": 90
+        }
+      ],
+      "output": "Number",
+      "outputShape": Blockly.OUTPUT_SHAPE_ROUND,
+      "colour": Blockly.Colours.textField
+    });
   }
 };
--- a/blocks/math.js
+++ b/blocks/math.js
@@ -90,3 +90,18 @@ Blockly.Blocks['math_positive_number'] = {
     this.setOutputShape(Blockly.OUTPUT_SHAPE_ROUND);
   }
 };
+
+Blockly.Blocks['math_angle'] = {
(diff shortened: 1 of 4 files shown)


core/field_angle.js
```diff
@@ -195,9 +195,15 @@ Blockly.FieldAngle.prototype.showEditor_ = function() {
           Blockly.FieldAngle.HALF + ',' + Blockly.FieldAngle.HALF + ')'
     }, svg);
   }
-  svg.style.marginLeft = (15 - Blockly.FieldAngle.RADIUS) + 'px';
+
+  Blockly.DropDownDiv.setColour(this.sourceBlock_.parentBlock_.getColour(), this.sourceBlock_.getColourTertiary());
```
rachel-fenichel: nit: line length

blocks/math.js
```diff
@@ -90,3 +90,18 @@ Blockly.Blocks['math_positive_number'] = {
...
   }
 };
+
+Blockly.Blocks['math_angle'] = {
+  /**
+   * Block for angle picker.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.setHelpUrl(Blockly.Msg.MATH_NUMBER_HELPURL);
```
rachel-fenichel: Prefer jsonInit--everything here should be doable with json.

tmickel:
Ok - I went ahead and converted all the math blocks to JSON as well :)

tmickel:
Also, I went ahead and removed the injection of angleValidator. Because I want to allow values like this to be typed:

screen shot 2016-07-08 at 9 06 13 am

Although it makes sense to keep the validator around, I think (and it's used for the actual drawing of the picker).

blocks/math.js
```diff
@@ -67,12 +59,20 @@ Blockly.Blocks['math_whole_number'] = {
...
-        .appendField(new Blockly.FieldNumber('0', 0, Infinity, 1), 'NUM');
-    this.setOutput(true, 'Number');
-    this.setOutputShape(Blockly.OUTPUT_SHAPE_ROUND);
+    this.jsonInit({
+      "message0": "%1",
+      "args0": [
+        {
+          "type": "field_number",
+          "name": "NUM",
+          "min": 0,
```
rachel-fenichel, line 68: Is leaving off max the same as setting it to infinity?

blocks/math.js
```diff
@@ -82,12 +82,19 @@ Blockly.Blocks['math_positive_number'] = {
...
-        .appendField(new Blockly.FieldNumber('0', 0, Infinity), 'NUM');
-    this.setOutput(true, 'Number');
-    this.setOutputShape(Blockly.OUTPUT_SHAPE_ROUND);
+    this.jsonInit({
+      "message0": "%1",
+      "args0": [
+        {
+          "type": "field_number",
+          "name": "NUM",
+          "min": 0
```
rachel-fenichel, line 91: This is the same min as you have for math_whole_number.  One of these should allow negative numbers.

blocks/math.js
```diff
@@ -32,32 +32,24 @@ goog.require('Blockly.Colours');
...
- * Common HSV hue for all blocks in this category.
- */
-Blockly.Blocks.math.HUE = Blockly.Colours.textField;
-
 Blockly.Blocks['math_number'] = {
   /**
    * Block for generic numeric value.
    * @this Blockly.Block
    */
   init: function() {
```
rachel-fenichel, line 40: (Doesn't have to be done in this pr)

Your math blocks are different from the default math blocks in enough ways--mostly involving tooltips and helpurls--that you should move them into blocks_horizontal/ or blocks_vertical.  Otherwise we may be fighting merge conflicts over and over again.

core/field_angle.js
```diff
@@ -51,35 +52,6 @@ Blockly.FieldAngle = function(text, opt_validator) {
 goog.inherits(Blockly.FieldAngle, Blockly.FieldTextInput);
 
 /**
- * Sets a new change handler for angle field.
- * @param {Function} handler New change handler, or null.
- */
-Blockly.FieldAngle.prototype.setValidator = function(handler) {
```
rachel-fenichel, line 57: You've deleted the function, but deleted no calls to it.  What's up with that?

blocks/math.js
tmickel, line 68: should be!

blocks/math.js
tmickel, line 91: I think this is intentional. In my view a "positive number" is a "float > 0": use case is "wait _ seconds" (doesn't make sense to wait for -1 second, but does make sense to wait for 0.5 seconds). In my view a "whole number" is "unsigned int" or "int > 0": use case is "repeat _ times" where it doesn't make sense to repeat -1 times or .5 times. We could possibly have a "signed int" math block, but I haven't found a block that needs that yet. Do you have any in mind?

blocks/math.js
tmickel, line 40: Fair enough. My intention in keeping them in blocks/ is that they could be shared between the horizontal and vertical (as they currently are). Maybe we should get rid of blocks/ and put our shared blocks in a blocks_common/ to avoid merge conflicts?

core/field_angle.js
tmickel, line 57: The call is in the super-class constructor:
https://github.com/google/blockly/blob/master/core/field.js#L50

The purpose of this function was to override the old `setValidator`:
https://github.com/google/blockly/blob/master/core/field.js#L227

and to wrap it with the angleValidator. I don't want to override field's `setValidator`, and getting rid of the override does the job I think.

blocks/math.js
towerofnix, line 91: Maybe "change costume # by _"? Then again, that block doesn't exist.. but it's a possible use case.

tmickel, line 91: @liam4: I think if we added that block we should add the corresponding number type :) do you know of any in the current set? maybe "go back _ layers" now that I look at it?

towerofnix, line 91: Yeah, that works - but it feels to me like there should either be a "go forward _ layers" block, or a dropdown (i.e. "go [back/forward] _ layers").

For the sake of not breaking older projects, though, "go back _ layers" should definitely have a signed int input!

liam4: Oh, yes! Are graphic effects affected by decimal values?

towerofnix, line 91: The "loudness" block returns a signed integer: -1 when the microphone isn't connected, 0+ for the volume when it is. That's not an input though.

tmickel, line 91: Hmm, yeah. we'll have to think about the layering blocks, but it might make sense to have the signed int for backwards compatibility.

Graphic effects look to me like they can be affected by decimal values - try "set pixelate effect to 0.1" and then "set pixelate effect to 0.5".

Pen colors is an interesting example - I can't even really see the difference between colors until you start getting to increments of more than 1. So, maybe we should disallow the decimal values. But, I might argue that:
screen shot 2016-07-10 at 11 43 04 am
should be equivalent to:
screen shot 2016-07-10 at 11 43 15 am

tmickel, line 91: (eventually). in fact I think the above loop causes the color to change over 10 frames, so it's not quite equivalent...

towerofnix, line 91: I made a couple of test projects.. but I'm a bit color blind, so it's hard for me to be much help here..

Can you see a transition from one color to another in the projects? It (shade or color) only goes up by 1 each transition so it's probably hard to see anyways.

https://scratch.mit.edu/projects/115972279/
https://scratch.mit.edu/projects/115972387/

tmickel, line 91: Hmm, thanks for that! I can't see the difference either, but I think we should do some poking/thinking about the scratch-flash pen code, too, just to be sure. :)

PullJosh, line 91: I'm not entirely sure I'm understanding correctly, but my interpretation of this conversation is that you want to know whether decimal differences in pen color affect the color being displayed.

If so, the answer is yes:
colordiff
That being said, I'm not sure backwards compatibility is really a big deal in this case. If pen colors were rounded to the nearest integer, I find it hard to believe anybody would mind.

tmickel, line 91: Thanks for doing the measurements, @PullJosh! That's helpful to know. Yes - I agree that it's probably not a big deal. However, I think I'm also leaning towards allowing decimals in those blocks. I suspect there is some benefit of disallowing decimals where it doesn't make sense in context (i.e., repeat 1.5 times) to help guide usage, but the connection is not so clear here (where color mappings to numbers are fairly arbitrary anyway).

towerofnix, line 91: Here's a relevant snippet [from](https://github.com/LLK/scratch-flash/blob/master/src/scratch/ScratchSprite.as#L274-L294) scratch-flash:

``` as3
public function setPenHue(n:Number):void {
  penHue = n % 200;
  if (penHue < 0) penHue += 200;
  updateCachedPenColor();
}

public function setPenShade(n:Number):void {
  penShade = n % 200;
  if (penShade < 0) penShade += 200;
  updateCachedPenColor();
}

private function updateCachedPenColor():void {
  var c:int = Color.fromHSV((penHue * 180) / 100, 1, 1);
  var shade:Number = (penShade > 100) ? 200 - penShade : penShade; // range 0..100
  if (shade < 50) {
    penColorCache = Color.mixRGB(0, c, (10 + shade) / 60);
  } else {
    penColorCache = Color.mixRGB(c, 0xFFFFFF, (shade - 50) / 60);
  }
}
```

As far as I can tell, there's nothing _visibly_ changing (assuming it's rendering as 0xRRGGBB), but you can see in my project that the actual value is changing - otherwise the color/shade would stay the same forever: `int(50 + 0.2) == 50`. (edit2: my math is bad, PullJosh is right)

EDIT: yay, more notes happened before I wrote this (and I didn't see them). What I'm saying in this note is a response to [this](https://github.com/LLK/scratch-blocks/pull/493#discussion_r70184385)..

blocks/math.js
rachel-fenichel, line 40: By "our shared blocks" do you mean "blocks shared between vertical and horizontal" or "blocks shared between google/blockly and scratch-blocks"?  If it's the former--yes, I think that would be a good idea.

tmickel, line 40: Yes, the former (vertical and horizontal). Not clear to me what blocks are/would be shared between google/blockly and scratch-blocks.

core/field_angle.js
rachel-fenichel, line 57: sgtm

PR #496, opened 2016-07-07
Added Variable Blocks

ghost:
All Variables blocks Added
Note: "New Variable" Option is not Working

### DIFF
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -21,15 +21,6 @@ Blockly.Blocks['data_variablemenu'] = {
           {
             "type": "field_variable",
             "name": "VARIABLE",
-            "options": [
-              ['v', 'V'],
-              ['v2', 'V2'],
-              ['v3', 'V3'],
-              ['v4', 'V4'],
-              ['v5', 'V5'],
-              ['v6', 'V6'],
-              ['v7', 'V7'],			  
-            ]
           }
         ],
         "inputsInline": true,
@@ -59,7 +50,7 @@ Blockly.Blocks['data_variable'] = {
       "colour": Blockly.Colours.data.primary,
       "colourSecondary": Blockly.Colours.data.secondary,
       "colourTertiary": Blockly.Colours.data.tertiary,
-      "output": "Number",
+      "output": "String",
       "outputShape": Blockly.OUTPUT_SHAPE_ROUND,	  
       "checkboxInFlyout": true
     });
@@ -100,7 +91,7 @@ Blockly.Blocks['data_changevariableby'] = {
    */
   init: function() {
     this.jsonInit({
-      "message0": "change %1 to %2",
+      "message0": "change %1 by %2",
       "args0": [
         {
           "type": "input_value",
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -6,6 +6,7 @@ goog.require('Blockly.Blocks');
 goog.require('Blockly.Colours');
 goog.require('Blockly.constants');
 
+
 Blockly.Blocks['data_variablemenu'] = {
   /**
    * Variable menu.
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -1,0 +1,167 @@
+'use strict';
+
+goog.provide('Blockly.Blocks.data');
+
+goog.require('Blockly.Blocks');
+
+goog.require('Blockly.Colours');
+
+goog.require('Blockly.constants');
+
+Blockly.Blocks['data_variablemenu'] = {
+  /**
+   * Variable menu.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit(
+      {
+        "message0": "%1",
+        "args0": [
+          {
+            "type": "field_variable",
+            "name": "VARIABLE",
+            "options": [
+              ['v', 'V'],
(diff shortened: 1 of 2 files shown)


blocks_vertical/data.js
```diff
@@ -0,0 +1,167 @@
...
+};
+
+Blockly.Blocks['data_changevariableby'] = {
+  /**
+   * Block to change variable by a certain value
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": "change %1 to %2",
```
tmickel: change %1 by %2

tmickel:
"New variable" seems to work for me :)

Although we have to work on having the "new variable" button and other prompts as @rachel-fenichel is working on, this looks to me like a great start. @rachel-fenichel do you foresee any conflicts?

blocks_vertical/data.js
```diff
@@ -0,0 +1,167 @@
...
+   */
+  init: function() {
+    this.jsonInit(
+      {
+        "message0": "%1",
+        "args0": [
+          {
+            "type": "field_variable",
+            "name": "VARIABLE",
+            "options": [
```
tmickel: I think you can remove this options list, since field_variable populates its own options.

blocks_vertical/data.js
```diff
@@ -0,0 +1,167 @@
...
+        "args0": [
+          {
+            "type": "input_value",
+            "name": "VARIABLE"
+          }
+        ],  
+      "colour": Blockly.Colours.data.primary,
+      "colourSecondary": Blockly.Colours.data.secondary,
+      "colourTertiary": Blockly.Colours.data.tertiary,
+      "output": "Number",
```
tmickel: I think the output type here should also be "String", since variables can have string values.

tests/vertical_playground.html
```diff
@@ -666,7 +667,43 @@
           </value>
         </block>
       </category>
-      <category name="Data" colour="#FF8C1A"></category>
+      <category name="Data" colour="#FF8C1A">
```
rachel-fenichel, line 670: For now we can have this category for testing, but this will eventually be dynamically populated, through Blockly.Variables.flyoutCategory

blocks_vertical/data.js
```diff
@@ -0,0 +1,167 @@
...
+        }		
+      ],
+      "previousStatement": null,
+      "nextStatement": null,
+      "colour": Blockly.Colours.data.primary,
+      "colourSecondary": Blockly.Colours.data.secondary,
+      "colourTertiary": Blockly.Colours.data.tertiary
+    });
+  }
+};
```
rachel-fenichel: Add newline at end of file

blocks_vertical/data.js
```diff
@@ -0,0 +1,167 @@
+'use strict';
+
+goog.provide('Blockly.Blocks.data');
+
+goog.require('Blockly.Blocks');
+
```
rachel-fenichel: No empty between goog.require lines.  Keep the empty line between provide and require.

blocks_vertical/data.js
```diff
@@ -0,0 +1,167 @@
+'use strict';
+
+goog.provide('Blockly.Blocks.data');
+
+goog.require('Blockly.Blocks');
+
+goog.require('Blockly.Colours');
+
+goog.require('Blockly.constants');
+
```
rachel-fenichel, line 8: Two empty lines between the last require and the start of the code.

tmickel:
Perfect, thank you so much for all your contributions, @joaoh1!

blocks_vertical/data.js
towerofnix: (A _little_ off-topic, could fit well in a new issue, but..)

Does this mean that a variable containing the value `true` will now be the same (when called) as a variable containing the value `"true"`?

For example, in this demo:

Demo

`0 + "true"` is `0` and `0 + true` is `1`. Assuming to-number convertion stays the same in Scratch 3.0, and variables _always_ output strings – because their `output` is `"String"` – can we expect `0 + (var whose value is: true)` to be `0`?

tmickel: I think the intention is to keep the behavior the same as in 2.0 with this sort of thing. In fact maybe the right output type here is "Any." We'll make sure the behavior is right when we come around to implementing variables in the VM :) Thanks for pointing this out - it will definitely be helpful then.

jeandrek:
I like how there's a single variable getter block with a dropdown. I think it's more consistent with how in Scratch 1.3+ then the set block has a dropdown rather than a setter block for each variable.

skistaddy:
@Jonathan50 Did you get 42 notifications from this repo too?

mrjacobbloom:
why are the secondary and tertiary colors for data red and green? Just because the colors spec doesn't specify them?

jeandrek:
> why are the secondary and tertiary colors for data red and green? Just because the colors spec doesn't specify them?

It's the same for sensing. I'm guessing it's temporary

tmickel:
@Jonathan50: Not sure if the drop-down is going to stay or not. While it is more consistent, it does add an extra step to pull a specific variable out of the palette and makes them more hidden.
@bjskistad Sorry about all the activity...
@Airhogs777 Yes, these are temporary colors (#FF0000 and #00FF00) until we work out a more complete version of the color set.

PR #498, opened 2016-07-07
Only use "grabby" cursors when dragging

tmickel:
I wanted us to play with this version of mouse cursors to start getting a feel for it. This replaces the Blockly "open hand" cursor with the CSS default (i.e., mouse pointer) cursor. When a drag exceeds DRAG_RADIUS, we switch to the closed cursor (instead of on mouse-down previously).

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -644,8 +644,6 @@ Blockly.BlockSvg.prototype.onMouseDown_ = function(e) {
       Blockly.Events.setGroup(true);
     }
     // Left-click (or middle click)
-    Blockly.Css.setCursor(Blockly.Css.Cursor.CLOSED);
-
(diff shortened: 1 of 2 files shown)


tmickel:
Demo:
demo-cursors

rachel-fenichel:
Personally, not a fan of this behaviour.  I like the indication that the block can be moved.  But I assume you'll be testing it.  Will take a look at the code.

tmickel:
Yep. The confusing bit about having the indication that the block can be moved is that it possibly hides the indication that the block can be clicked to execute. Not sure what to do here, but figured it would be good to have something to start playing with. I also tried cursor = "pointer" (like what you get for an <a> link) but that felt really odd.

Issue #512, opened 2016-07-12
Drag bugs related to merge


tmickel:
@rachel-fenichel 
What I've found so far:
- [x] When creating a new block from the flyout, the block disappears for a moment (until the mouse cursor moves).
- [x] Snapping two blocks together causes the stack to jump
- [x] Dragging sometimes causes this error:

> block_svg.js:198 Uncaught TypeError: Cannot read property 'id' of null

tmickel:
Sometimes I am able to trigger this error when creating a block from the flyout:

> flyout.js:970 Uncaught TypeError: Cannot read property 'contentHeight' of null

More and more of these errors accumulate as I move the mouse.

tmickel:
^ During the above error, the block is invisible until dropped.

tmickel:
Note: snapping statement blocks appears to move the stack downward; snapping in a value block appears to move the stack upward...

tmickel:
- Angle picker is broken

rachel-fenichel:
"Cannot read property 'id' of null" was due to a failure to correctly dispose of connections, meaning that the block would try to bump deleted blocks whose connections were still in the database.  This was due to accidential deletion of one line.
https://github.com/LLK/scratch-blocks/pull/510/files#diff-5148dfb78f1d110be97f1d3539fa647cL260

rachel-fenichel:
I think fixing connection disposal also fixes " snapping statement blocks appears to move the stack downward; snapping in a value block appears to move the stack upward" and "Snapping two blocks together causes the stack to jump".

rachel-fenichel:
What's broken on the angle picker?  I can't get it to show up.  Is that the brokenness?

rachel-fenichel:
"When creating a new block from the flyout, the block disappears for a moment (until the mouse cursor moves)." was caused by not distinguishing between autoclosing flyouts and always open flyouts.  Also fixed in #514.

tmickel:
The angle picker only comes up in the "point in direction" block (the only one, I think, that we have to set an absolute angle).

The error is:
`field_angle.js:215 Uncaught TypeError: Blockly.FieldAngle.angleValidator is not a function`
I can probably investigate...

PR #514, opened 2016-07-12
Fix connection disposal and get rid of startFlyout_

rachel-fenichel:
#512


tmickel:
Looks great, and seems to address all of the problems (minus the angle picker). Thanks!

PR #516, opened 2016-07-13
Make Red the color placeholder instead of black

ghost:
At the moment, if a block dont have a defined colour, it would be replaced by black:
black
(Yes, i removed the "When flag clicked! block's color to tests)
But at Scratch 2.0, the red have this function,
In this PR, i made red as the Color Placeholder


tmickel:
Cool! Makes sense to me.

PR #539, opened 2016-07-15
Allow Variable use text as Value

ghost:
In Scratch 2.0, you can set text at "set variable to [value]" block
In Scratch Blocks, it only accept numbers
Well, i made this PR to allow variables to accept text like Scratch 2.0

(Remember: I don't know if this is a feature or a bug '_')


tmickel:
Thanks!

PR #558, opened 2016-08-09
Adjust vertical alignment of horizontal field text

### LINKED ISSUE #555: Text in horizontal fields not vertically aligned
block text alignment
_before and after 3f3862a_

The text is closer to the bottom of the field after the change. Was this intentional?

drigz:
Fixes #555.

This doesn't seem to affect the alignment when the editor is open. @tmickel, do you think it would affect anything else?

block text alignment 2
_before regression, after regression, after fix_


tmickel:
looks solid - thanks!

Issue #556, opened 2016-08-09
Large spacing in horizontal toolboxes


drigz:
![block spacing](https://cloud.githubusercontent.com/assets/632207/17514492/3019123a-5e33-11e6-9df2-d43fe4c0444a.png)
_changes in horizontal spacing, old (top), new (bottom)_

Two commits have changed the horizontal spacing in the toolbox, 991dba0 and 6539a63. Were these changes intentional?

If so, is it possible to adjust this? Changing Flyout.prototype.MARGIN affects the spacing from the top/side as well.

rachel-fenichel:
Yes, it's possible to add a new variable (Flyout.prototype.BLOCK_SPACING_X/Y?) to separate the margin and the inter-block spacing.  It looks like 6539a63 changed the margin from 8 to 12, leading to the noticeable change in size.

To make this change you would modify the isNan case on this line: https://github.com/LLK/scratch-blocks/blob/develop/core/flyout.js#L658

No, I don't recall any intent to move the blocks that far apart.

tmickel:
@drigz looks to me like this is fixed?

drigz:
It's now customizable by applications, so as long as you're happy with the default spacing being larger than before, this is fixed.

PR #565, opened 2016-08-12
Add GAP_X/Y to the flyout

drigz:
This makes it easy for users to customise the spacing in the flyout,
without having to repeat the same value of "gap" for every block.

This doesn't affect #556, although these constants could be changed to
adjust fix it.


rachel-fenichel:
This belongs in google/blockly, not just scratch-blocks.  Please make the pull request over there.

tmickel:
noticed you made the PR over there, so I'll close this one :)

PR #569, opened 2016-08-15
Size insertion markers with creating block metrics

tmickel:
Here's a possible "final" fix for #250.

When an insertion marker is created, the block that's creating it saves its rendered width to the insertion marker's `insertionMarkerMinWidth_`. Then, the insertion marker uses that property to properly set its width.

I also switched back to not drawing inputs/input shapes/icons in insertion markers, for the performance benefit, and because the width is correct anyway now.

before:
before

after:
after

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -84,6 +84,12 @@ Blockly.BlockSvg.prototype.height = 0;
 Blockly.BlockSvg.prototype.width = 0;
 
 /**
+ * Minimum width of block if insertion marker; comes from inserting block.
+ * @type {number}
+ */
+Blockly.BlockSvg.prototype.insertionMarkerMinWidth_ = 0;
+
+/**
  * Opacity of this block between 0 and 1.
  * @type {number}
  * @private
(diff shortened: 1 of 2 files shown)


core/block_svg.js
```diff
@@ -124,16 +130,18 @@ Blockly.BlockSvg.INLINE = -1;
  */
 Blockly.BlockSvg.prototype.initSvg = function() {
   goog.asserts.assert(this.workspace.rendered, 'Workspace is headless.');
-  // Input shapes are empty holes drawn when a value input is not connected.
-  for (var i = 0, input; input = this.inputList[i]; i++) {
-    input.init();
-    if (input.type === Blockly.INPUT_VALUE) {
-      this.initInputShape(input);
+  if (!this.isInsertionMarker()) { // Insertion markers not allowed to have inputs or icons
```
rachel-fenichel, line 133: What about statement inputs?

tmickel, line 133: hmm, I think maybe the comment is misleading. the inputs will still exist, so statement inputs will still be rendered. but they aren't initialized, so fields on them aren't generated/initialized, and their SVG elements are not created, as desired: https://github.com/LLK/scratch-blocks/blob/develop/core/input.js#L208

PR #595, opened 2016-09-15
Import procedures from Blockly

tmickel:
There is obviously a _lot_ to work on here. But I figured it could let us get started implementing custom blocks in the VM, more or less. I'm imagining the data model we'll need for procedures is pretty close to this (except we need the parameters to be typed, shadow blocks to be generated in various places). Let me know if you have objections...

screen shot 2016-09-15 at 3 04 29 pm

### DIFF
--- a/blocks_vertical/more.js
+++ b/blocks_vertical/more.js
@@ -49,28 +49,7 @@ Blockly.Blocks['procedures_defnoreturn'] = {
     this.setTooltip(Blockly.Msg.PROCEDURES_DEFNORETURN_TOOLTIP);
     this.setHelpUrl(Blockly.Msg.PROCEDURES_DEFNORETURN_HELPURL);
     this.arguments_ = [];
-    this.setStatements_(true);
-    this.statementConnection_ = null;
-  },
-  /**
-   * Add or remove the statement block from this function definition.
-   * @param {boolean} hasStatements True if a statement block is needed.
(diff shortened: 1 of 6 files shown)


PullJosh:
Interesting to have two different define blocks (one with and one without a built-in report block). Has any thought been given to what happens if a "report" block is placed somewhere other than inside a custom reporter define block?

tmickel:
@PullJosh This is not a designed feature, it's pulled directly from Blockly.

tmickel:
So no, no thought from us - currently it does something like this: :)
screen shot 2016-09-15 at 3 22 35 pm

PullJosh:
If it helps any, we (as a community) have had a bit of discussion on this topic: https://scratch.mit.edu/discuss/topic/199574/?page=3#post-1990777

jeandrek:
Yay!!

Why does Scratch 2.0 have "define" instead of "to" like Logo?

MegaApuTurkUltra:
@Jonathan50 because that's how it was designed. I personally think "define" is more clear than "to"

rachel-fenichel:
LGTM, knowing that we have plenty of future work.

One nit: more.js isn't a great filename.  Not much information in it.  I suggest procedures.js or functions.js, even though thta means it won't match the category name.

tmickel:
^ agreed - will make an issue for that!

PR #602, opened 2016-09-16
Provide an importable module

rschamp:
Allows this from other packages:

``` javascript
const ScratchBlocksVertical = require('scratch-blocks');
const ScratchBlocksHorizontal = require('scratch-blocks/dist/horizontal');
```

How it works:
When another package requires `scratch-blocks`, webpack looks at the package.json and finds the `webpack` entry pointing to `shim/index.js`. 
- `shim/index.js` exports horizontal and vertical Blockly from `msg/messages.js`
  - This requires `goog` from `shim/blockly_compressed_[horizontal,vertical].goog`
    - `shim/blockly_compressed_[horizontal,vertical].goog` exports the `goog` var from the actual `blockly_compressed_[horizontal,vertical]`
  - `Blockly` is provided from `shim/blocks_compressed_[horizontal,vertical].Blockly`
- `shim/blocks_compressed_[horizontal,vertical]` exports `Blockly` from `blocks_compressed`, but the `Blockly` provided to `blocks_compressed` is the one exported by `shim/blockly_compressed_[horizontal,vertical].Blockly`
- `shim/blockly_compressed_[horizontal,vertical].Blockly` exports `Blockly` from `shim/blockly_compressed_[horizontal,vertical]`
- `shim/blockly_compressed_[horizontal,vertical]` exports `{goog: goog, Blockly: Blockly}` from `blockly_compressed_[horizontal,vertical]`

The intermidate shims are required due to the way the imports and exports loaders work — you can't do a multi-level loading, or this would be simpler to do.  You can also set this up more simply in webpack (see first commit), but then you would need to build something for other packages to import. This makes a persistent shim that just exports the product of the compressed files (and messages.js).

Other packages don't need to know about the details of this shimming process: imports-loader and exports-loader will be provided by this package when webpack requires the modules out of this package.


tmickel (approved):
Wow cool! I'm excited, looks great to me!

rschamp:
I decided it would be better if we provided a module that could be imported without webpack (e.g., for tests).  If you want to you can still import the unbuilt module, but for efficiency, the vertical library is the default imported when you `require('scratch-blocks')`.

thisandagain:
Looks great. Thanks @rschamp

PR #609, opened 2016-09-18
Procedure UI Fix (Without Changing "to" to "Define")

TheBrokenRail:


tmickel:
In general we are working on a redesign of the custom blocks UI in scratch-blocks, so we probably won't be able to take PRs improving it until we are done with that. But since this is a simple fix, I think we can take it.

PR #620, opened 2016-09-22
Merge sept 22

rachel-fenichel:
Includes better multitouch support.


rachel-fenichel:
@tmickel updated to fix merge conflicts, which mostly derive from refactors done in only one branch.  I need to be careful going forward--it's already painful to do merges like these.

Demo is up as usual, and tested on iOS with chrome and safari; on android with chrome; and on linux with chrome and firefox.

tmickel:
It's working really well for me!! Very exciting to see this.

Issue #365, opened 2016-06-04
Class attributes to distinguish blocks


mrjacobbloom:
This is a feature request. When people start to create themes for Scratch 3 they'll want a way to distinguish the inputs from the blocks themselves with CSS styles. It'd be nice if there was a class-name on the paths to separate the two.

tmickel:
Sounds like a nice idea to me so Scratchers could customize how things look with user-styles :)

We should definitely wait until we play out #238, since that would restructure how things are arranged.

mrjacobbloom:
Whenever y'all revisit this, consider adding classes to distinguish block shape and category as well. That'll make creating themes/skins significantly easier.

rachel-fenichel:
On a deeply nested stack of blocks, adding the class to each block may be
an expensive operation.

mrjacobbloom:
I've been thinking about how to do this. Shape classes could be added on the fly during rendering/initialization based on connections.

Category isn't as easy to determine. I see a few options:
- The easiest way is to use the substring of block.type before the first underscore
  - The problem with this option is that it depends on a naming convention and is therefore susceptible to human error
- Compare the block's color[s] to Blockly.Colours
  - There are some edge cases where this would fail. For example, if two categories shared the same color.
  - It'll also involve a for-loop since we're searching for values instead of keys, and that could be kind of expensive.
    - The for-loop could be avoided by generating a reversed colors database. For example `{"#FFFFFF": "categoryName"}`. But that seems like overkill...
- Add category names to the block definitions
  - This would involve an overhaul of the block definitions but I think it's the cleanest solution. Down the line, this could also simplify how we set block colors.

Sometime down the line, would you guys be willing to accept a PR that adds this functionality?

tmickel:
Regarding category: I would personally lean toward option 3 (adding category names to block definitions) for exactly the reason you say. The color-boilerplate reduction would possibly be worth the addition regardless of making it easier to user-style the blocks. In fact, if we went this route, we could set the colors of blocks entirely through CSS on the category name and throw out the Blockly.Colours dictionary for blocks. That's not how Blockly approaches the color problem right now, but I think we could entertain it.

mrjacobbloom:
Have you guys discussed #238 at all? I'd be willing to get started on a pull request to add the block shape class code.

tmickel:
Hey @Airhogs777! We have discussed it. It's looking like it might happen, but might take some time. So it might be best to wait before adding the class code.

[...]

PR #621, opened 2016-09-24
Classify blocks by shape

mrjacobbloom:
This is my first attempt at implementing the shapes part of #365. My initial idea had been to add classes to the elements. After some consideration, I decided to put the list in a data-attribute instead to avoid interfering with Blockly's classes (like `blocklyDraggable`, etc). This data attribute is called `data-shapes` and users should still be able to access it with CSS and JS.

Thoughts:
- Right now the function runs on render. It might be better to have it run on block init, but then we'd have to make a special exception for blocks that change shape (like `stop [ v]`)
- The shape names I used were whatever seemed most obvious to me. They're definitely up for discussion

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -983,7 +983,11 @@ Blockly.BlockSvg.prototype.renderClassify_ = function() {
   var shapes = [];
   
   if(this.outputConnection) {
-    shapes.push("reporter");
+    if(this.isShadow_) {
+      shapes.push("input");
+    } else {
+      shapes.push("reporter");
+    }
     if(this.edgeShape_ === Blockly.OUTPUT_SHAPE_HEXAGONAL) {
       shapes.push("boolean");
     } else if(this.edgeShape_ === Blockly.OUTPUT_SHAPE_ROUND) {
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -573,6 +573,8 @@ Blockly.BlockSvg.prototype.render = function(opt_bubble) {
 
   var inputRows = this.renderCompute_(cursorX);
   this.renderDraw_(cursorX, inputRows);
+  
+  this.renderClassify_();
 
   if (opt_bubble !== false) {
     // Render all blocks above this one (propagate a reflow).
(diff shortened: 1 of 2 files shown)


core/block_render_svg_vertical.js
```diff
@@ -974,6 +976,55 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(iconWidth, inputRows) {
...
 /**
+ * Give the block an attribute 'data-shapes' that lists its shape[s]
+ * @private
+ */
+Blockly.BlockSvg.prototype.renderClassify_ = function() {
+  var shapes = [];
+  
+  if(this.outputConnection) {
+    if(this.isShadow_) {
+      shapes.push("input");
```
tmickel: "Input" is a pretty overloaded word in scratch-blocks land. We could do "argument", maybe?

core/block_render_svg_vertical.js
```diff
@@ -974,6 +976,55 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(iconWidth, inputRows) {
...
+    } else {
+      shapes.push("reporter");
+    }
+    if(this.edgeShape_ === Blockly.OUTPUT_SHAPE_HEXAGONAL) {
+      shapes.push("boolean");
+    } else if(this.edgeShape_ === Blockly.OUTPUT_SHAPE_ROUND) {
+      shapes.push("string");
```
tmickel: Not just for strings - also for numbers. How about just "round"?

core/block_render_svg_vertical.js
```diff
@@ -974,6 +976,55 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(iconWidth, inputRows) {
...
+    } else {
+      // this shouldn't happen anymore because string-shaped reporters were phased out
```
tmickel: Drop the empty else

core/block_render_svg_vertical.js
```diff
@@ -974,6 +976,55 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(iconWidth, inputRows) {
...
+      if (input.connection && input.connection.type === Blockly.NEXT_STATEMENT) {
+        statementCount++;
+      }
+    }
+    
+    if(statementCount) {
+      shapes.push("c-block");
+      if(statementCount > 1) {
+        shapes.push("else");
+      }
```
tmickel: maybe `"c-" + statementCount`, i.e., c-1, c-2, c-3, etc.?

core/block_render_svg_vertical.js
```diff
@@ -974,6 +976,55 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(iconWidth, inputRows) {
...
+      shapes.push("hat"); // c-block+hats are possible (e.x. reprter procedures)
+    } else if(!statementCount) {
+      shapes.push("stack"); //only call it "stack" if it's not a c-block
+    }
+    if(!this.nextConnection) {
+      shapes.push("end");
+    }
+  }
+  
+  this.svgGroup_.setAttribute('data-shapes', shapes.join(" "));
```
tmickel: The whole PR mixes double-quote strings and single-quote strings. Let's keep it to single-quote strings?

tmickel:
Hey @Airhogs777,

Very cool! I like using the data attribute quite a lot.

Running on render seems mostly to make sense to me. There is a slight performance hit, I suppose, but should be negligible compared to SVG repaint, for example.

Made some naming/style suggestions.

What do you think, @rachel-fenichel ?

core/block_render_svg_vertical.js
mrjacobbloom: I like that idea a lot. I'll leave the "c-block" shape in as well since I worry that it'll make it harder to detect any c-block with CSS if we just use the numbers

core/block_render_svg_vertical.js
mrjacobbloom: "round" is much better

mrjacobbloom: It looks like this also applies to color pickers, dropdowns that use a shadow block, etc. "round" is clearly the best option here

mrjacobbloom:
Should I add (nearly) duplicate code to block_render_svg_horizontal.js or should that be a separate PR?

core/block_render_svg_vertical.js
```diff
@@ -974,6 +976,51 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(iconWidth, inputRows) {
 };
 
 /**
+ * Give the block an attribute 'data-shapes' that lists its shape[s]
```
rachel-fenichel: Nit: period at the end of a comment.

core/block_render_svg_vertical.js
```diff
@@ -974,6 +976,51 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(iconWidth, inputRows) {
...
+ * @private
+ */
+Blockly.BlockSvg.prototype.renderClassify_ = function() {
+  var shapes = [];
+  
+  if(this.outputConnection) {
```
rachel-fenichel: Style, here and everywhere: space between "if" and the leading parenthesis.

core/block_render_svg_vertical.js
```diff
@@ -974,6 +976,51 @@ Blockly.BlockSvg.prototype.renderDraw_ = function(iconWidth, inputRows) {
...
+    var inputList = this.inputList;
+    var statementCount = 0;
+    for (var i = 0, input; input = inputList[i]; i++) {
+      if (input.connection && input.connection.type === Blockly.NEXT_STATEMENT) {
+        statementCount++;
+      }
+    }
+    
+    if(statementCount) {
+      shapes.push('c-block');
```
rachel-fenichel, line 1007: Is the order of the elements in shapes going to matter at any point?

rachel-fenichel:
Looks solid!  My comments are mostly just style nits.

core/block_render_svg_vertical.js
mrjacobbloom: You mean the order of items in the list? I imagine this would be used with css's [a~=b] selector, which expects a space-separated list but shouldn't be affected by the order of the list. In JS, I imagine most use-cases would look for one list-item at a time so that shouldn't be an issue

mrjacobbloom:
I made those changes. Is there anything else you think needs adjustment?

tmickel (approved):

tmickel:
Thanks again @Airhogs777 !

PR #622, opened 2016-09-24
Publish to NPM

rschamp:
Move build dependencies back to devDependencies.

The build is now a prepublish step, not a postinstall step since the build will be supplied with the package when installed.

Use travis-after-all to deploy to npm only after all tests pass in all Node environments.

Only release non-PR builds for master and develop branches. For convenience, these branches are defined in the Settings panel for the repo in Travis. This way builds can be rerun without a new push with different values for these vars (e.g. if the token needs to change or a release should happen outside of the normal release branches)

Since this happens in `after_script`, we can't use Travis's built-in npm deployment. So create an .npmrc for authentication for `npm publish`.

Publish versions with the package.json version + "-[git sha]", e.g. 0.1.0-93d2f. This allows us to deploy every push to GitHub without considering version bumps. This is intended to be a temporary solution until we have more systematic releases.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,15 @@
+# Localization / I18N
+common.pyc
+.settings
+.project
+*.pyc
+*.komodoproject
+/nbproject/private/
+
+# Unused by scratch-blocks
+javascript_compressed.js
+lua_compressed.js
+php_compressed.js
+python_compressed.js
+
+/accessible/*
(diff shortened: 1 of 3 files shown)


rschamp:
I didn't attempt to publish the playgrounds to `gh-pages` because the uncompressed versions require the closure library. If there's a way to avoid that setup, then I'd be happy to give it a go.

tmickel (approved):

thisandagain:
👍

PR #623, opened 2016-09-26
Drop goog.Timer, replacing with `setTimeout`/`clearTimeout`

tmickel:
goog.Timer has some strange scoping situation where it was expecting something to be set to `window`...but it wasn't after we changed our build process. Ultimately these functions just call `setTimeout` and `clearTimeout` under the hood, so I actually don't see a reason for them.

@rschamp


tmickel:
Going to merge this so we can proceed-happy to revert if it's a problem!

PR #624, opened 2016-09-27
Build compressed js before publishing

rschamp:
Untracks the compiled build assets, and builds them with Travis before each npm publish so that they're available on npm install. This ensures that the compiled assets always reflect the source files.


tmickel (approved):

rschamp:
The wiki is updated with the info that you must `npm install` to get updated built assets when running from a git checkout.

PR #630, opened 2016-09-27
Position drop-down based on actual block dimensions

### LINKED ISSUE #587: Text drop-down appears below stack of blocks
See:
screen shot 2016-09-06 at 4 52 28 pm

tmickel:
`getHeightWidth` gets the height/width of all blocks in the stack below the block, but `block.height` and `block.width` should be the actual relevant dimensions of the block. Fixes #587.

Before:

screen shot 2016-09-27 at 2 39 17 pm

After:

screen shot 2016-09-27 at 2 39 03 pm

### DIFF
--- a/core/dropdowndiv.js
+++ b/core/dropdowndiv.js
@@ -177,7 +177,7 @@ Blockly.DropDownDiv.setColour = function(backgroundColour, borderColour) {
 Blockly.DropDownDiv.showPositionedByBlock = function(owner, block,
       opt_onHide, opt_secondaryYOffset) {
   var scale = block.workspace.scale;
-  var bBox = block.getHeightWidth();
+  var bBox = {width: block.width, height: block.height};
   bBox.width *= scale;
   bBox.height *= scale;
   var position = goog.style.getPageOffset(block.getSvgRoot());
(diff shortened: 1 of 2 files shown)


core/dropdowndiv.js
```diff
@@ -177,7 +177,7 @@ Blockly.DropDownDiv.setColour = function(backgroundColour, borderColour) {
 Blockly.DropDownDiv.showPositionedByBlock = function(owner, block,
       opt_onHide, opt_secondaryYOffset) {
   var scale = block.workspace.scale;
-  var bBox = block.getHeightWidth();
```
rachel-fenichel: Why was getHeightWidth giving the wrong values?

Does this solve the problem of the dropdown showing up below the entire block instead of below the field?

rachel-fenichel:
I think this will still fail if you have a dropdown in a c-block that isn't in the last row, though I think that's not a case you'll hit with the current set of blocks.

tmickel:
Hmmmm...so, it should be positioned like this: https://github.com/google/blockly/blob/develop/core/field_dropdown.js#L156 ?

PR #632, opened 2016-09-27
Set checkboxes in drop-downs to scroll with menu

tmickel:
Previously the checkbox in the Closure menu would not scroll when a menu was scrolled. Since it now seems likely that we'll stick with the Closure menu, I thought I would fix this.

Before: see current https://llk.github.io/scratch-vm/

After:
scroll

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -917,17 +917,6 @@ Blockly.Css.CONTENT = [
     'width: 16px;',
   '}',
 
-  /* BiDi override for the selected/checked state. */
-  /* #noflip */
-  '.blocklyWidgetDiv .goog-menuitem-rtl .goog-menuitem-checkbox,',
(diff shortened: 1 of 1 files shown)


core/css.js
```diff
@@ -934,6 +934,9 @@ Blockly.Css.CONTENT = [
   '.blocklyDropDownDiv .goog-option-selected .goog-menuitem-icon {',
      /* Client apps may override the URL at which they serve the sprite. */
     'background: url(<<<PATH>>>/sprites.png) no-repeat -48px -16px !important;',
+    'position: static;', /* Scroll with the menu. */
+    'float: left;',
```
rachel-fenichel, line 927: rtl?

tmickel: yeah..hmmm..maybe closure itself should fix the issue where the checkbox doesn't scroll, I guess. but I can try to think of some rtl hack or something

PR #638, opened 2016-09-29
Classify blocks by category

mrjacobbloom:
The categories piece of #365. If a category is included in the block definition, then the block's svgGroup gets an attribute `data-category` with the category. This is more of a proof-of-concept than a serious PR.

The changes to block definitions would look something like this:

```
this.jsonInit({
      ...
      "category": Blockly.Categories.event,
      ...
    });
```

Once implemented, a system like this might even allow for a transition from inline styles to CSS (way down the line).


tmickel:
Makes sense to me! Do you plan to add them to the blocks in a future PR - should we merge this now - or do you want to continue working on this?

mrjacobbloom:
I was waiting to hear what you thought about that. I was thinking one commit per category in this PR. Would that work for you?

tmickel:
@Airhogs777 sounds great!

mrjacobbloom:
@tmickel that should be all of them. I've tested it in my copy of the repo and it looks like each block has the correct category attribute.

tmickel:
Thanks again!

PR #644, opened 2016-10-01
Add attribute for argument type

mrjacobbloom:
This adds an attribute `data-argument-type` to inputs so they can be styled based on that (for example, styling dropdowns with a secondary color but leaving inputs as white).

The way I achieved this was a little hacky, so I'd understand if you're more reluctant to accept this PR

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1280,6 +1280,7 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign) {
         }
       } while (altRepeat);
       if (field) {
+        field.fieldType = element['type'].replace(/field_/, '');
         fieldStack.push([field, element['name']]);
       } else if (input) {
         if (element['check']) {
--- a/core/field.js
+++ b/core/field.js
@@ -140,9 +140,9 @@ Blockly.Field.prototype.init = function() {
   // Build the DOM.
   this.fieldGroup_ = Blockly.createSvgElement('g', {}, null);
   if (this.sourceBlock_.isShadow_) {
-    this.sourceBlock_.svgGroup_.setAttribute('data-field-type', this.fieldType);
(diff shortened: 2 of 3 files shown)


core/block.js
```diff
@@ -1280,6 +1280,7 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign) {
         }
       } while (altRepeat);
       if (field) {
+        field.fieldType = element['type'].replace(/field_/, '');
```
tmickel, line 1283: Any particular reason for this replacement?

core/field.js
```diff
@@ -139,6 +139,11 @@ Blockly.Field.prototype.init = function() {
   }
   // Build the DOM.
   this.fieldGroup_ = Blockly.createSvgElement('g', {}, null);
+  if (this.sourceBlock_.isShadow_) {
```
tmickel, line 142: `isShadow()`?

core/field.js
```diff
@@ -139,6 +139,11 @@ Blockly.Field.prototype.init = function() {
...
+    this.sourceBlock_.svgGroup_.setAttribute('data-argument-type', this.fieldType);
+  } else {
+    this.fieldGroup_.setAttribute('data-argument-type', this.fieldType); // square dropdowns
```
tmickel, line 145: possibly not just dropdowns - potentially any field not in a shadow block

core/block.js
tmickel: Also, how does this work for blocks that don't use the JSON initialization?

mrjacobbloom: the replacement is just to shave some extra text from the attribute. There's no explicit reason to remove it

mrjacobbloom: you're right, I get "undefined" for the dropdown in the stop block

mrjacobbloom:
This way feels too hacky, I think I'll start over and see if I can find a better way to do this

PR #645, opened 2016-10-02
Add attribute for argument type (take 2)

mrjacobbloom:
This adds an attribute `data-argument-type` to inputs so they can be styled based on that (for example, styling dropdowns with a secondary color but leaving inputs as white).

This system should be more robust than #644 and I made some of the changes suggested on that PR


tmickel:
Neat - this version looks good to me!

PR #646, opened 2016-10-02
Add classes for insertion-markers and replacement-rings

mrjacobbloom:
If a block is an insertion-marker, it'll have the class `blocklyInsertionMarker`. If a reporter, input, or inputShape has the replacementGlowFilter, it'll also have the class `blocklyReplaceable`.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -625,6 +625,7 @@ Blockly.Block.prototype.setInsertionMarker = function(insertionMarker) {
   if (this.isInsertionMarker_) {
     this.setColour(Blockly.Colours.insertionMarker);
     this.setOpacity(Blockly.Colours.insertionMarkerOpacity);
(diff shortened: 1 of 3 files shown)


tmickel:
Just wondering, why are you using the class here (all your others are using the data attribute)?

mrjacobbloom:
I guess because category and shape are general information that you'd expect to find on any block, but replacement/insertion are more like special cases that only happen one-at-a-time. It could just as easily be an attribute though

mrjacobbloom:
The main reason I started using data attributes is because the list of shapes could be several long and I didn't want to worry about how to remove unwanted shapes if a block's shape changed without affecting other classes already on the block. That's not an issue in these cases

tmickel:
Ok, makes sense to me!

PR #647, opened 2016-10-02
Dropdown menu categories

mrjacobbloom:
When an input opens a dropdown menu, that dropdown will have the same `data-category` attribute as its parent block


tmickel:
Nice!

PR #650, opened 2016-10-03
Rename more.js; recompile 10/3

tmickel:


rschamp:
There shouldn't be any need for recompiles anymore. Did a merge add back the build assets?

tmickel:
Hmmm...maybe. They're definitely there, but I can do another commit to remove them.

rschamp:
I swear I deleted them, but the PR that did so doesn't show that happening, so I guess not? https://github.com/LLK/scratch-blocks/pull/624

PR #654, opened 2016-10-03
Add compressed vertical playground

tmickel:


rschamp (approved):
Looks good. I think we can use this for the basis of a playground published to GitHub pages.

PR #657, opened 2016-10-04
Move square dropdown border under text

mrjacobbloom:
For styling purposes, it's easier to have the background rect for a square dropdown behind the text element. That way, users can add a fill color to create a background for the field without covering the text.

This doesn't seem to break the click event to open the menu, although I haven't tested it on mobile.


mrjacobbloom:
Sorry about those extra commits, GitHub briefly went down while I was pushing to my fork and the commit failed to attach to this PR.

I added the class `blocklyBlockBackground` to the rect, so now both square and round dropdowns' backgrounds can be selected with the CSS selector `g[data-argument-type="dropdown"] > .blocklyBlockBackground`.

tmickel:
Makes sense to me.

PR #661, opened 2016-10-04
Switch from Replacing argType to Appending To It

mrjacobbloom:
When a field updates its argType (its `data-argument-type` attribute), now it'll append a string instead of writing over what was there before. So you'll end up with things like `dropdown variable` and `text angle`

I'm actively experimenting with the themes to figure out what works an what doesn't, and I think this change would make it easier to style things like variable dropdowns without excessive selectors

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -238,11 +238,16 @@ Blockly.Field.prototype.setVisible = function(visible) {
 };
 
 /**
- * Sets the field's argType (used for styling).
+ * Sets or appends a string to the field's argType (used for styling).
  * @param {string} argType New argType.
  */
 Blockly.Field.prototype.setArgType = function(argType) {
-  this.argType_ = argType;
+  if (this.argType_ == null) {
+    this.argType_ = '';
+  } else {
+    this.argType_ += ' ';
+  }
+  this.argType_ += argType;
 };
 
 /**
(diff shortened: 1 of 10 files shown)


mrjacobbloom:
would you prefer if the function name setArgType changed to something more descriptive of its new functionality? (maybe createOrAppendToArgType or just changeArgType?)

tmickel:
Hmmm, I'm not sure I completely understand. Can you illustrate how `setArgType` is called multiple times - does the superclass (text) call it once, and then the subclass (angle)? Could we avoid the multiple calls, and just have angle call it with "text angle"?

tmickel:
I'd prefer it if we didn't have to do the appending, but if you're 100% sure it's necessary, the name should reflect it. (appendDataArgType?)

P.S. thanks for all your work on theming - should be really great to have. Would even be great to transition some of our internal styling to just use CSS!

Airhogs777:
We could store them in an array and join them after

mrjacobbloom:
Yes, right now it's called by the superclass and then the subclass.
If you only want the function to run once, I suppose angle could set it to 'text angle' before the superclass constructor runs, and then text would check if it's already set and only set it if it's unset.
If we keep the function running multiple times, I like how the the class hierarchy kind of keeps track of itself. But if we go with that then the function needs a new name -- right now I'm leaning towards `addArgType`. And now that you point it out, appending is kind of clumsy. How about if we store them in an array and `getArgType` (or `getArgTypes`?) joins them?

I'm fine with either of those two options, but I'm leaning towards the latter.

tmickel:
The array idea sounds good!

PR #662, opened 2016-10-05
Text+drop-downs and number+drop-downs

### LINKED ISSUE #531: Implementation of text-editing/dropdown combo
Scratch 2.0:
94d32b2e-2d6f-11e6-8536-974e7fb9de83

I think best to implement this as an extension to field_textinput. Maybe we should separate out the functionality in field_dropdown so it can be shared between the two?

tmickel:
Here are two new types of field, the FieldTextDropDown and FieldNumberDropDown.

Demo:
num-drop

These are for when we want to allow users to type values, but also provide a fixed set of convenience options. Most of the code of the actual fields is inherited from either FieldTextInput or FieldDropDown; there is some logic for deciding when to call to the drop-down methods (chose to inherit the text input).

Beyond that, there are a couple of rendering tweaks to make sure things are positioned correctly. And, additional code to render a drop-down arrow on top of the text-editor when needed.

Resolves #531.

cc: @ericrosenbaum


core/field_number.js
```diff
@@ -127,6 +110,24 @@ Blockly.FieldNumber.NUMPAD_DELETE_ICON = 'data:image/svg+xml;utf8,' +
...
+ * Return an appropriate restrictor, depending on whether this FieldNumber
+ * allows decimal or negative numbers.
+ * @param {boolean} decimalAllowed Whether number may have decimal/float component.
+ * @param {boolean} negativeAllowed Whether number may be negative.
+ * @return {!RegExp} Regular expression for this FieldNumber's restrictor.
+ */
+Blockly.FieldNumber.getNumRestrictor = function(decimalAllowed, negativeAllowed) {
+  var pattern = "[\\d]"; // Always allow digits.
+  if (decimalAllowed) {
+    pattern += "|[\\.]";
```
cpseager, line 120: Hard coding with .  is not great for future localisation for languages such as French and German that use comma as the decimal separator. Scratch 2 never supported European numbers correctly, but it would be great if Scratch 3 did!

tmickel: That's a pretty good point and deserves some attention. Probably not going to change this behavior for this PR, but I'll make an issue of it here and in scratch-vm.

core/field_numberdropdown.js
```diff
@@ -0,0 +1,61 @@
...
+    (opt_precision == 0) ||
+    (Math.floor(opt_precision) != opt_precision);
+  this.negativeAllowed_ = (typeof opt_min == 'undefined') || isNaN(opt_min) || opt_min < 0;
+  var numRestrictor = Blockly.FieldNumber.getNumRestrictor(
+    this.decimalAllowed_, this.negativeAllowed_
+  );
+  Blockly.FieldNumberDropdown.superClass_.constructor.call(
+    this, value, menuGenerator, opt_validator, numRestrictor
+  );
+  this.setArgType('numberdropdown');
```
mrjacobbloom: #661 might change the name of this function

tmickel:
Thanks @Airhogs777 - merged and fixed!

blocks_vertical/data.js
```diff
@@ -206,6 +206,66 @@ Blockly.Blocks['data_listcontents'] = {
...
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": "%1",
+      "args0": [
+        {
+          "type": "field_numberdropdown",
+          "name": "INDEX",
+          "value": "1",
+          "min": 0,
```
rachel-fenichel: Are these zero or one indexed?

blocks_vertical/data.js
```diff
@@ -206,6 +206,66 @@ Blockly.Blocks['data_listcontents'] = {
...
+      ],
+      "output": "String",
+      "category": Blockly.Categories.data,
+      "outputShape": Blockly.OUTPUT_SHAPE_ROUND,
+      "colour": Blockly.Colours.textField
+    });
+  }
+};
+
+Blockly.Blocks['data_listindexrandom'] = {
```
rachel-fenichel, line 239: Why are these different blocks?

core/field_numberdropdown.js
```diff
@@ -0,0 +1,61 @@
...
+ *     to validate any constraints on what the user entered.  Takes the new
+ *     text as an argument and returns the accepted text or null to abort
+ *     the change.
+ * @extends {Blockly.FieldTextInput}
+ * @constructor
+ */
+Blockly.FieldNumberDropdown = function(value, menuGenerator, opt_min, opt_max, opt_precision, opt_validator) {
+  this.decimalAllowed_ = (typeof opt_precision == 'undefined') || isNaN(opt_precision) ||
+    (opt_precision == 0) ||
+    (Math.floor(opt_precision) != opt_precision);
```
rachel-fenichel: Move to previous line

core/field_numberdropdown.js
```diff
@@ -0,0 +1,61 @@
...
+  this.negativeAllowed_ = (typeof opt_min == 'undefined') || isNaN(opt_min) || opt_min < 0;
```
rachel-fenichel: This parsing of options is the same as in FieldNumber.  Make it into a function that you can call in both constructors.

core/field_textdropdown.js
```diff
@@ -0,0 +1,149 @@
...
+  // Prevent the drop-down handler from changing the field colour on open.
+  this.disableColourChange_ = true;
+};
+
+/**
+ * Close the input widget if this input is being deleted.
+ */
+Blockly.FieldTextDropdown.prototype.dispose = function() {
+  if (this.mouseUpWrapper_) {
+    Blockly.unbindEvent_(this.mouseUpWrapper_);
```
rachel-fenichel, line 93: Also need to call Blockly.Touch.clearTouchIdentifier() when ending a touch, or you'll get stuck (most obvious on ios).

core/field_textdropdown.js
```diff
@@ -0,0 +1,149 @@
...
+Blockly.FieldTextDropdown.prototype.positionArrow = function(x) {
+  return Blockly.FieldDropdown.prototype.positionArrow.call(this, x);
+};
+
+/**
+ * If the drop-down isn't open, show the text editor.
+ */
+Blockly.FieldTextDropdown.prototype.showEditor_ = function() {
+  if (!this.dropDownOpen_) {
+    Blockly.FieldTextDropdown.superClass_.showEditor_.call(this, null, null,
```
rachel-fenichel: What does the superClass refer to here?

core/field_textinput.js
```diff
@@ -175,6 +178,36 @@ Blockly.FieldTextInput.prototype.showEditor_ = function(opt_quietInput, opt_read
...
+    dropDownArrow.style.cursor = 'pointer';
+    // Magic number for positioning the drop-down arrow on top of the text editor.
+    var dropdownArrowMagic = '11px';
+    if (this.sourceBlock_.RTL) {
+      dropDownArrow.style.left = dropdownArrowMagic;
+    } else {
+      dropDownArrow.style.right = dropdownArrowMagic;
+    }
+    if (opt_arrowCallback) {
+      htmlInput.dropDownArrowMouseWrapper_ = Blockly.bindEvent_(dropDownArrow,
```
rachel-fenichel, line 205: arrowCallback needs to call Blockly.Touch.clearTouchIdentifier, unless you want to bind to mouseup here to make sure you end the gesture.

blocks_vertical/data.js
tmickel: they are actually one-indexed lists: updated the min!

blocks_vertical/data.js
tmickel: Hmmm, they have different drop-down menus (one contains "all" as an option (for delete), one contains "random"). Not sure how to do this otherwise?

rachel-fenichel: Oh I see--one is for delete and one is for insert.

core/field_textdropdown.js
tmickel: the superclass of the text drop-down is FieldTextInput - it's showing the text editor.

rachel-fenichel: Resolved offline

PR #677, opened 2016-10-13
Procedures as mutations with Scratch 2.0 data

tmickel:
Removes most of the existing Blockly procedure blocks, and replaces with a simple mechanism to read/parse Scratch 2.0 procedures.

### DIFF
--- a/blocks_common/math.js
+++ b/blocks_common/math.js
@@ -43,7 +43,8 @@ Blockly.Blocks['math_number'] = {
       "args0": [
         {
           "type": "field_number",
-          "name": "NUM"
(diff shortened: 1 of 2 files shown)


dekrain:
It's only one hat with lots of undefined in my workspace

tmickel:
@dekrain The flyout procedure category needs to be updated - as well as our procedure creation mechanism. But this provides the basic model data we need to deal with 2.0 procedures.

dekrain:
I deleted my comment, b/c i copy/paste this before. After updating with npm, it works fine

PR #679, opened 2016-10-13
Provide default toolbox XML

tmickel:
Between the blocks playgrounds (vertical/horizontal/compressed/multi), the VM playground, and the GUI, we are now reproducing toolbox XML approximately five times. It's become madness and it's leading to lots of bugs.

This is a stopgap proposal that we provide default toolbox XML with the vertical and horizontal blocks. I don't think it's a great solution in the long-term (would like to generate the toolbox dynamically for example), but will keep the standard toolbox in one place while we haven't designed a long-term solution.


rschamp (approved):
This seems like a good plan for the interim, thanks!

PR #686, opened 2016-10-15
Better IE and Edge Support

TheBrokenRail:
Fixes Toolbox Not Loading

### DIFF
--- a/core/options.js
+++ b/core/options.js
@@ -18,6 +18,7 @@
  * limitations under the License.
  */
 
+
 /**
  * @fileoverview Object that controls settings for the workspace.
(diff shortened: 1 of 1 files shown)


tmickel:
Hey @TheBrokenRail:

Could you explain what was broken before, and how this fixes it? Can't use IE/Edge myself so I believe there is something wrong, but just wondering what the problem was.

thisandagain:
Hi @TheBrokenRail unless I'm mistaken the `DOMParser` API is available in all of the platforms we are planning to support:
http://caniuse.com/#feat=xml-serializer

PR #687, opened 2016-10-16
Add Wedo Blocks To Vertical Grammer

TheBrokenRail:


thisandagain:
Thanks! We are still rethinking the extensions interface (see https://github.com/LLK/scratch-vm/issues/194) and so we won't be open to adding this to the VM for some time, but this is interesting to look at in the meantime.

Right now it looks like this is not passing tests due to some issues with your line breaks. See the travis logs for more info:
https://travis-ci.org/LLK/scratch-blocks/jobs/167980299

In the future I recommend running the tests locally before submitting a PR.

Issue #549, opened 2016-07-27
We should only call DropDownDiv.hideWithoutAnimation when the dropDown div is visible


picklesrus:
We should figure out why DropDownDiv.hideWithoutAnimation and clearContent are called on all mouse/touch move events during a workspace drag.  If the div isn't present, we should be able to skip this step and, while it doesn't cost a ton on desktop, it is taking taking a few milliseconds of each frame on a MotoX, and between .5 and 1ms on a nexus5 and a nexus9.

Here's the trace of the call from chrome's timeline view:
Blockly.onMouseMove_:blockly.js:245
Blockly.WorkspaceSvg.scroll: workspace_svg.js:1397
Blockly.DropDownDiv.hideWithoutAnimation: dropdowndiv.js:372
Blockly.DropDownDiv.clearContent:dropdowndiv.js:152

rachel-fenichel:
Closed by #698

PR #698, opened 2016-10-19
Cache height of the toolbox category box since it does not change.  A…

picklesrus:
…lso only call hideWithoutAnimation on the dropdown div if it is visible (Bug #549).

Caching the height avoids a forced re-layout (because of the call to element.offsetHeight) on every mouse move event during a workspace drag.

### DIFF
--- a/core/dropdowndiv.js
+++ b/core/dropdowndiv.js
@@ -386,6 +386,9 @@ Blockly.DropDownDiv.hide = function() {
  * Hide the menu, without animation.
  */
 Blockly.DropDownDiv.hideWithoutAnimation = function() {
+  if (!Blockly.DropDownDiv.isVisible()) {
+    return;
+  }
   var div = Blockly.DropDownDiv.DIV_;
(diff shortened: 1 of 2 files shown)


core/dropdowndiv.js
```diff
@@ -386,16 +386,19 @@ Blockly.DropDownDiv.hide = function() {
...
-  div.style.transform = '';
-  div.style.top = '';
-  div.style.left = '';
-  div.style.display = 'none';
-  Blockly.DropDownDiv.clearContent();
-  Blockly.DropDownDiv.owner_ = null;
-  if (Blockly.DropDownDiv.onHide_) {
-    Blockly.DropDownDiv.onHide_();
-    Blockly.DropDownDiv.onHide_ = null;
+  if (Blockly.DropDownDiv.isVisible()) {
```
tmickel: just a style question - why not `if (!Blockly.DropDownDiv.isVislbe()) return;` or similar?

picklesrus:
Errr, because it is what I happened to type first to see if it worked? :) Changed it to return early instead in case more stuff gets added to that method later

PR #699, opened 2016-10-20
Extentions

TheBrokenRail:
Uses JSON to create blocks.
JSON Importing Is Untested.
JSON Is Imported Currently Using JS Console:

``` js
Blockly.Extensions.EXTENSIONS.push('Your JSON Here');
```

JSON Example:

``` js
[
    {
        "spec":"Block Specs",
        "type":"Block Type",
        "id":"(Extension Name)_(Block Name)"
    }
]
```

Block Types:
- hat
- block
- reporter (Not Working)
- boolean (Not Working)


tmickel:
This looks possibly like it could be a reasonable approach we might take to generate blocks from specs...however I'm not sure if we're ready for it. @thisandagain?

core/extensions.js
```diff
@@ -0,0 +1,65 @@
...
+Blockly.Extensions.EXTENSIONS = [];
+
+/**
+ * Construct the blocks required by the flyout for the procedure category.
+ * @param {!Blockly.Workspace} workspace The workspace contianing procedures.
+ * @return {!Array.<!Element>} Array of XML block elements.
+ */
+Blockly.Extensions.flyoutCategory = function(workspace) {
+  var xmlList = [];
+  for (var v = 0; v < Blockly.Extension.EXTENSIONS.length; v++)
```
dekrain: You forgot to add `{`

dekrain: `{`

TheBrokenRail: Fixed

towerofnix:
> JSON Importing Is Untested.

Just out of curiosity's sake, the way you put this kind of implies you don't actually test your code, or see if it works, before making a pull request..

For something like this, I'd think some example usage would be nice. For example, how do I give a function to a block?

Also, I wonder, should `"code"` be `"spec"`?

TheBrokenRail:
For implementation we could store the code as a string in the JSON and use eval() to run it. To get code we could use the "id" tag.

TheBrokenRail:
Added a function to easily lookup blocks using there id, you could store additional values here like code that could be used with eval()

``` js
Blockly.Extensions.lookupBlock("Your Block ID Here");
```

TheBrokenRail:
I'll try changing code to spec

TheBrokenRail:
Changed!

towerofnix:
> For implementation we could store the code as a string in the JSON and use eval() to run it. To get code we could use the "id" tag.

That doesn't really sound good to me.

Why not just store a function in the object?

``` js
Blockly.Extensions.EXTENSIONS.push([
  {
    spec: 'my awesome block',
    id: 'my_awesome_block',
    type: 'block', // should this be stack?
    cb: function() {
      alert('Hello, world!')
    }
  }
])
```

Remember, old Scratch extensions are also JavaScript based, not JSON based. That's a really handy thing IMO. And since Scratch 3.0 is also JavaScript based, the more direct access to an extension object we can have, the better.. with this we could, for example, even update the spec of a block from _inside_ the extension! (by doing `var myBlock = {..cb: /* modifies myBlock */}; var myExtension = [myBlock];`)

Also on the topic of extensions - how much access should we directly give an extension block to Scratch? In my opinion it should be the same as other blocks -- access through `util`, `this.vm`, and whatever. That way extensions can be as powerful as they should be able to - as powerful as any other block. Isn't that the point of extensions?

core/extensions.js
```diff
@@ -0,0 +1,78 @@
...
+  for (var v = 0; v < Blockly.Extension.EXTENSIONS.length; v++) {
+    for (var i = 0; i < Blockly.Extension.EXTENSIONS[v].length; i++) {
+      var data = JSON.parse(Blockly.Extension.EXTENSIONS[v][i]);
+      var id = data.id;
+      if (id == blockId) {
+        return data;
+      }
+    }
+  }
+  return false;
```
dekrain, line 77: Missing semicolon

core/extensions.js
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Category to separate procedure names from variables and generated functions.
+ */
+Blockly.Extensions.NAME_TYPE = 'EXTENSION';
+
+Blockly.Extensions.EXTENSIONS = [];
+
+/**
+ * Construct the blocks required by the flyout for the procedure category.
+ * @param {!Blockly.Workspace} workspace The workspace contianing procedures.
```
dekrain, line 44: Remove this

dekrain (changes requested):

thisandagain:
Extensions as they exist in Scratch 2.0 are likely to be changed significantly in 3.0. I'd prefer to hold off on this until we have time to discuss and design what this will be. If you have thoughts on the extensions interface in-general, it would be great to hear your thoughts here:
https://github.com/LLK/scratch-vm/issues/194

dekrain:
Ok, I'll try to do it.

Issue #713, opened 2016-10-26
Code clean-up for glow/highlight blocks


tmickel:
After https://github.com/google/blockly/pull/718 lands, merge with our own `glowBlock`. Possibly rename `glowStack` to `highlightStack` to match. Finally, make sure the filters have the unique ID as needed, explained in the aforementioned PR.

cwillisf:
scratch-blocks v2 is a complete rewrite built on top of upstream [Blockly](https://github.com/RaspberryPiFoundation/blockly/). The internal code referenced in this issue no longer exists. Thank you for reporting it!

If you believe this issue is still relevant to scratch-blocks v2, please let us know and we can reopen it.

PR #720, opened 2016-11-02
Feature/merge nov 1

rachel-fenichel:
@tmickel please check that highlighting still works--I don't understand all of the highlighting changes in each repo.  See https://github.com/google/blockly/commit/144be4d49f36fdba260a26edbd170ae75bbc37a6


tmickel:
Looks good to me, checked it out and all works as expected. Note that the highlight changes still require work: https://github.com/LLK/scratch-blocks/issues/713

Not sure if we should just delete the Blockly `highlightBlock` and replace it with our glowBlock, or try to maintain them together.

Also it appears the build is failing because of about 6 linter errors. Assigning you for those, then I think it's good to merge.

Issue #722, opened 2016-11-05
Clicking on background of text fields doesn't focus field.


tmickel:
<img width="42" alt="screen shot 2016-11-05 at 11 55 36 am" src="https://cloud.githubusercontent.com/assets/120403/20031496/c65b6d14-a34e-11e6-895a-6ffaf9229dda.png">

e.g., in the above, if you click on the outer white part not near the text, the field isn't focused. This is a usability problem - makes it hard to focus and leads to lots of executing the block instead. Not sure when this was introduced, it used to work fine.

tmickel:
Looks like it was from the merge:
https://github.com/LLK/scratch-blocks/pull/720/files#diff-393b262d7ed6406aa328f2e57e089e5fL177

That needs to be changed back to `getClickTarget_()`

PR #723, opened 2016-11-07
Fix click target from merge #722

tmickel:
This should fix the issue in #722 from the Blockly merge, where click targets for fields were not correctly set.


thisandagain:
Thanks @tmickel

PR #725, opened 2016-11-08
Deal with unmovable workspaces and long presses in touch mode

rachel-fenichel:
@thisandagain this one's for you.  I think this covers everything, but let me know before I merge it.


thisandagain:
Great. That resolve the issue for me. Thanks @rachel-fenichel

PR #726, opened 2016-11-10
Fix setting constraints for number dropdown fields

### LINKED ISSUE #721: Clicking on Lists Category in Vertical Playground causes error
In the current vertical playground on develop (as well as in the hosted version here: https://utils.asun.co/scratch-blocks/tests/vertical_playground.html) as soon as the "List" category is clicked there is a console error and the app doesn't show the category correctly.

rachel-fenichel:
Fixes #721


tmickel:
Happy this was not too bad - thanks!

PR #730, opened 2016-11-16
Fix bug with dropdowndiv placement.  It was using getPage Offset to d…

picklesrus:
…o positioning calculations.  GetPageOffset takes into account the scroll position which we do not want.  It is not noticable if the body has overflow:hidden like the playgrounds do.


picklesrus:
Here's a screenshot of the problem.  screen shot 2016-11-16 at 10 17 34 am

rachel-fenichel:
lgtm

PR #734, opened 2016-11-17
Use bindEvent for longStop as well

rachel-fenichel:


thisandagain:
Looks good. Tested in Chrome, Firefox, and Edge.

PR #747, opened 2016-12-12
Use existing polyfills to regain IE compatibility

### LINKED ISSUE #736: Replacement marker filter breaks scratch-blocks on IE
https://github.com/LLK/scratch-blocks/blame/develop/core/inject.js#L137

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat#Browser_compatibility

This is simple enough to fix that we should explicitly work around it.

rachel-fenichel:
Fixes #736


thisandagain (approved):
👍

Issue #744, opened 2016-12-06
Github Pages - Deploy playgrounds to GH pages


thisandagain:
- [x] Setup 
    - [x] https://llk.github.io/scratch-blocks
    - [ ] https://llk.github.io/scratch-blocks/playgrounds/horizontal.html
    - [ ] https://llk.github.io/scratch-blocks/playgrounds/vertical.html
- [x] Update playgrounds to use compressed blocks or solve issue with closure-library dep.
- [x] Update travis config to deploy on merge to `develop`

ghost:
Well, i managed to do a gh-pages branch that works.
It uses Compressed Playgrounds (even the Horizontal one)
Note: the travis config was not updated in the branch
https://joaoh1.github.io/scratch-blocks (The Links in the page will not work because it's prepared for PR)
https://joaoh1.github.io/scratch-blocks/playgrounds/vertical.html
https://joaoh1.github.io/scratch-blocks/playgrounds/horizontal.html
I was going to make a PR for this, but there was no "gh-pages" branch.
EDIT: I fixed the Playground links

thisandagain:
Resolved in https://github.com/LLK/scratch-blocks/pull/755

rschamp:
Resolved in #755 (and fixed again in 279fb6331a3f93fb91d97885c25236b92afea823). Addendum to the previous plan, the links are:
*  https://llk.github.io/scratch-blocks/playgrounds/tests/vertical_playground.html
*  https://llk.github.io/scratch-blocks/playgrounds/tests/horizontal_playground.html

Issue #742, opened 2016-12-04
Add Compressed Horizontal Playground


TheBrokenRail:

thisandagain:
Closing in favor of the solution to deploy the playgrounds in #755

PR #748, opened 2016-12-16
Add a Compressed Horizontal Playground

ghost:
It adds a Compressed Horizontal Playground.
This PR "resolves" #742 and a part of #744 
Note: Ignore the first 2 commits of this PR, it was to update my develop branch, and that doesn't affect the rest of the Repository


thisandagain:
The vertical and horizontal playgrounds are primarily designed for development and thus we prefer to use the uncompressed blocks if possible. Closing in favor of #755

PR #750, opened 2016-12-21
Remove explicit file paths from npm scripts

### LINKED ISSUE #718: Make NPM build script work on Windows
It uses Mac/Linux-Slashes which could be avoided it calls to the module instead of its bat/sh file.

rschamp:
Fixes #718. I don't see any reason for these to be explicit, even if this repo is used with `npm link`.


cwillisf (approved):
👍

PR #755, opened 2016-12-24
Build playgrounds for gh-pages

rschamp:
Sets up the necessary structure in the gh-pages directory to run the uncompressed playgrounds. Deploy the gh-pages directory to the gh-pages branch after publishing to NPM.

Adds a landing page at the root to go to the horizontal and vertical playgrounds. Everything in `tests` is available under `/playgrounds/tests/` though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,5 @@ after_script:
     npm --no-git-tag-version version $($(npm bin)/json -f package.json version)-prerelease.$(date +%s)
     npm publish
     # Publish to gh-pages as most recent committer
-    git config --global user.email $(git log --pretty=format:"%ae" -n1)
-    git config --global user.name $(git log --pretty=format:"%an" -n1)
-    npm run --silent deploy -- -x -r https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git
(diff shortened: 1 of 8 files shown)


TheBrokenRail:
How about instead of copying the current playground create a compressed playground that uses the dist importable modules to make the playground more compact.

.travis.yml
```diff
@@ -28,4 +28,8 @@ after_script:
     # Set version to timestamp
     npm --no-git-tag-version version $($(npm bin)/json -f package.json version)-prerelease.$(date +%s)
     npm publish
+    # Publish to gh-pages as most recent committer
+    git config --global user.email $(git log --pretty=format:"%ae" -n1)
+    git config --global user.name $(git log --pretty=format:"%an" -n1)
+    npm run --silent deploy -- -x -r https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git
```
cwillisf: If there's any concern about these config values affecting other `git` commands in undesirable ways, then environment variables are an option too: https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#Committing

cwillisf (approved):

cwillisf:
@TheBrokenRail: the playgrounds attached to Scratch 3.0 repositories are intended for a few purposes: they demo the project or module, but they also offer a place to debug and learn about that module through live tinkering. Your suggestion would potentially benefit people who are interested in using the playground as a demo, especially if they're in an environment where downloads are slow or expensive, but would likely make debugging and learning about the module more difficult.

.travis.yml
rschamp: Oh, cool. I didn't realize you could do that.

thisandagain (approved):
Looks good. Thanks @rschamp

PR #761, opened 2017-01-05
Make Broadcast shadow blocks more consistent.

TheBrokenRail:
This goes with LLK/scratch-vm#368 to update the spec map.


tmickel:
@TheBrokenRail:

Does this actually work as intended? As far as I remember, the "when broadcast" hat as implemented does not evaluate reporters which are dropped into it.

TheBrokenRail:
Fixed!

TheBrokenRail:
I used the "check" value with a value of "Broadcast" as well as the output of the shadow "Broadcast" this makes sure nothing else is allowed (no reporters)

TheBrokenRail:
Any comments (I fixed the reporter issue), if not is this ready to merge along with the VM PR?

thisandagain:
We will cover this in a separate pass where we address drop-down menus.

PR #762, opened 2017-01-05
Fix IE and Edge Default Toolbox

TheBrokenRail:
For some reason IE and edge don't have the children tag when using oParser.


TheBrokenRail:
This Works!

## Edge
edge
## IE
ie

TheBrokenRail:
By the way I had to put the simple horizontal toolbox into a separate variable Blockly.Blocks.defaultToolboxSimple for later. I did this because having two root elements in the xml caused a parsing error.

TheBrokenRail:
This seems ready to merge to me, no failed test and it works. Is it ready to merge?

thisandagain (approved):

PR #775, opened 2017-01-28
Make Zoom center on Workspace and not Blockly

TheBrokenRail:
### Resolves

_Part of #43_

### Proposed Changes

_Make Zoom center on Workspace and not Blockly_

### Reason for Changes

_So zoom does not look offset_

### Test Coverage

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1348,6 +1348,35 @@ Blockly.WorkspaceSvg.prototype.zoomCenter = function(type) {
   var metrics = this.getMetrics();
   var x = metrics.viewWidth / 2;
   var y = metrics.viewHeight / 2;
+  if (this.options.RTL == false) {
+    if (this.options.horizontalLayout == false) {
+      if (this.options.toolboxPosition == 0) {
+        x = x - 250;
+      } else {
+        x = x + 250;
+      }
+    } else {
+      if (this.options.toolboxPosition == 0) {
+        y = y - 158;
+      } else {
+        y = y + 158;
+      }
+    }
+  } else {
+    if (this.options.horizontalLayout == false) {
+      if (this.options.toolboxPosition == 0) {
+        x = x + 250;
+      } else {
+        x = x - 250;
+      }
+    } else {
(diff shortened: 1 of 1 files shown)


core/workspace_svg.js
```diff
@@ -1348,6 +1348,15 @@ Blockly.WorkspaceSvg.prototype.zoomCenter = function(type) {
   var metrics = this.getMetrics();
   var x = metrics.viewWidth / 2;
   var y = metrics.viewHeight / 2;
+  if (this.options.toolboxPosition == 2) {
```
rachel-fenichel: No magic numbers, please.  They make debugging in the future very difficult.

TheBrokenRail: Do you mean the metric.viewHeight or the toolbox Position, if the toolbox position, is there any other way to detect it?

rschamp: I think @rachel-fenichel means to use a constant rather than a magic number to test this. The constants for the toolbox position are defined here: https://github.com/LLK/scratch-blocks/blob/d7745b12fdb5ee9a4761b7e99fd201673e0e7a8e/core/constants.js#L192-L214

TheBrokenRail: This has no "magic numbers" now

core/workspace_svg.js
```diff
@@ -1348,6 +1348,13 @@ Blockly.WorkspaceSvg.prototype.zoomCenter = function(type) {
   var metrics = this.getMetrics();
   var x = metrics.viewWidth / 2;
   var y = metrics.viewHeight / 2;
+  if (this.options.toolboxPosition == Blockly.TOOLBOX_AT_LEFT) {
+    x = x + 250;
+  } else if (this.options.toolboxPosition == Blockly.TOOLBOX_AT_RIHT) {
```
rschamp: Looks like there's a typo for `TOOLBOX_AT_RIGHT` here.

TheBrokenRail: Fixed!

thisandagain:
This is still using magic numbers to perform the offset. Closing until we have time to come up with an alternative approach.

PR #790, opened 2017-02-13
Fix translateSurface for Android 4.4

msintov:
### Resolves

Fixes translateSurface for Android 4.4.

### Proposed Changes

By using the "-webkit-transform" css property instead of "transform", the transformation works on Android 4.4.

### Reason for Changes

Prior to this change, on Android 4.4, the block would not appear to drag from the toolbox to the work area. Instead, the block would pop to the upper left corner of the screen and remain there during the drag.

### Test Coverage

I have tested this change on Android 4.4, Android 6.0, and iOS 9.3.

### DIFF
--- a/core/block_drag_surface.js
+++ b/core/block_drag_surface.js
@@ -151,7 +151,7 @@ Blockly.BlockDragSurfaceSvg.prototype.translateSurface = function(x, y) {
   x = x.toFixed(0);
   y = y.toFixed(0);
   transform =
-    'transform: translate3d(' + x + 'px, ' + y + 'px, 0px); display: block;';
+    '-webkit-transform: translate3d(' + x + 'px, ' + y + 'px, 0px); display: block;';
   this.SVG_.setAttribute('style', transform);
 };
 


picklesrus:
Just an FYI that there were fixes submitted upstream to blockly to fix this as well:
https://github.com/google/blockly/commit/40a063763c74b3f712c3057565966c25d5cfdb10
https://github.com/google/blockly/commit/89ee9809d9ce98b694f937fc74eb6e30104540a8

msintov:
@picklesrus Awesome! Thanks for pointing that out. Looking forward to getting those fixes.

rschamp (approved):
This looks good to me, though I'm curious why replacing `transform` with `-webkit-transform` (rather than adding `-webkit-transform`) doesn't have any effect in non-webkit browsers?  Everything looks fine in Firefox.

The commits mentioned by @picklesrus use a helper method to add both properties.

rschamp:
Per offline discussion, we will be closing this PR in favor of the upstream changes to be merged in soon.

PR #791, opened 2017-02-15
Feature/merge feb 2017

rachel-fenichel:
Merge google/blockly develop into scratch-blocks develop.


rachel-fenichel:
6 jsunit tests fail--I'm looking into them.  

@thisandagain @rschamp please sanity check basic functionality.

rachel-fenichel:
Data category doesn't work (probably missing a call to register a toolbox category)

rachel-fenichel:
Data category population fixed.

RTL text rendering is a bit weird: 
image

rachel-fenichel:
The problem with renaming and deleting variables is still there, but I want to fix that separately.

rachel-fenichel:
Unit tests fixed.

@rschamp there are some changes to package.json that may be wrong and/or blocking Travis, but I don't know enough to debug them.

Other than the RTL text input rendering, I think this is all good.

rachel-fenichel:
RTL text inputs fixed.

thisandagain:
It looks like travis is unhappy due to eslint issues with the tests:
https://travis-ci.org/LLK/scratch-blocks/jobs/202054431

I think we could just make exceptions here to the `no-undef` and `no-unused-vars` rules to get things passing again.

thisandagain:
Extremely minor, but I noticed that we now have a stroke around the insertion shadow.

thisandagain:
In the horizontal blocks:
If you edit a numeric field, the field will return to its original value on blur.

rachel-fenichel:
"In the horizontal blocks:
If you edit a numeric field, the field will return to its original value on blur."
@picklesrus and I have both noticed this as well, but she has it on a version of scratch-blocks pre-merge.  I don't think it's specific to this merge.

"Extremely minor, but I noticed that we now have a stroke around the insertion shadow."
Meaning there should be one and isn't, or shouldn't be one and is?

RE: no-undef: were those rules ever being applied to the unit tests, or is it just that the new unit tests aren't set up to be exceptions and the old ones were?

thisandagain:
#### Horizontal Block Inputs
Ok! Happy to file an issue and follow-up after.

#### Insertion Shadow
There is no stroke on insertion shadows in `develop` (and design spec) and this PR introduces a stroke.

#### Lint Ignore
Looks like you found it. :-)

rachel-fenichel:
Insertion shadow: Fixed, I think.  Looks like that wasn't quite right before, but two bugs combined to become a feature, and I fixed one of those bugs in this PR.

I think I'm all set.  If you're happy with this, please squash and merge it in.  Then I can write some extension demos for you.

thisandagain (approved):

thisandagain (changes requested):
When testing in the VM playground, we get a fatal exception on the line:
```
var flyoutWorkspace = workspace.getFlyout().getWorkspace();
```

```
Uncaught TypeError: workspace.getFlyout is not a function
    at window.onload (playground.js:64)
```
Did the API for this change? @rschamp can you please review this as well?

rschamp:
I see the same thing. It looks like `getFlyout` was made private in workspace, and should be implemented by subclasses. `getFlyout` was removed from workspace_svg.  I think we could just un-remove it from workspace_svg?

rachel-fenichel:
Not quite--we failed at the merge last time and ended up with two versions of getFlyout_ in workspace_svg:
https://github.com/LLK/scratch-blocks/blob/develop/core/workspace_svg.js#L648
https://github.com/LLK/scratch-blocks/blob/develop/core/workspace_svg.js#L528

I removed one to dedupe it.

I can fix this for y'all by making workspace:getFlyout and workspace_svg:getFlyout public, though that's out of sync with mainline blockly.  Is that the fix you want?

thisandagain (approved):

Issue #797, opened 2017-02-16
Right click on shadow block should display full context menu from parent


ed-cooper:
### Expected Behavior / Behavior from Scratch 2.0

If a shadow block is not in editing state, it should display the same context menu items as the main block (eg. duplicate, delete block, help, etc.). If a shadow block is in editing state, it still displays the same context menu with duplicate, delete, etc.

### Actual Behavior
Right-clicking main block: 
Works as expected
![s1](https://cloud.githubusercontent.com/assets/14095134/23034760/c29035c6-f474-11e6-9104-511d396c6f7b.png)

Right-clicking shadow block, whilst not in editing state: 
Should display duplicate and delete options for the main block
![s2](https://cloud.githubusercontent.com/assets/14095134/23034797/e0f49f70-f474-11e6-9ab5-2414f5406d0f.png)

### Operating System and Browser

Windows 10, Chrome 56.0.2924.87 (64-bit)

mrjacobbloom:
I like having clipboard options in the context menu when editing an input, but they could be added to the scratch-blocks context menu instead of using a native one

ed-cooper:
Yes, I agree that functionality like copy and paste is useful

thisandagain:
I don't think that this is a bug. The browser and OS determine how text input fields are rendered to a large extent. Having access to the context menu while editing text seems advantageous for a variety of reasons (copy/paste, accessibility, etc.).

ed-cooper:
Yes, but what about the first case where you right-clicked a shadow block whilst not editing it?

It doesn't feel right that the duplicate / delete options aren't shown

thisandagain:
Ah. Now I understand what you mean. I'm going to edit your original description to clarify.

ed-cooper:
Ok, thanks

rachel-fenichel:
See parent_tooltip_when_inline extension in google/blockly: develop

ed-cooper:
I presume this was meant to be closed after the patch was merged

PR #801, opened 2017-02-20
If a Block is a Shadow Block use context menu from parent (#797)

TheBrokenRail:
### Resolves

_#797_

### Proposed Changes

_If a block does not meet the criteria check if it's parent block does_

### Reason for Changes

_So inputs have proper right click menus._


thisandagain (approved):
Looks good! Thanks. 😄

PR #803, opened 2017-02-20
Fix Sprinkles

TheBrokenRail:
### Resolves

_#739_

### Proposed Changes

_Return toolbox instead of null._

### Reason for Changes

_Stress Testing_

### DIFF
--- a/tests/horizontal_playground.html
+++ b/tests/horizontal_playground.html
@@ -107,8 +107,7 @@
       }
 
       function getToolboxElement() {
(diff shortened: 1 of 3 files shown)


thisandagain (changes requested):
Thanks for the help @TheBrokenRail ! Unfortunately `workspace` is `null` here and thus this patch causes a fatal TypeError on page load.

thisandagain:
@TheBrokenRail please feel free to open a new PR if you would like to take another try at this.

PR #805, opened 2017-02-21
Fit Context Menu Height to Menu

### LINKED ISSUE #111: Context Menu creates scroll bar.
### Expected Behavior

_It is hidden or goes the other way._

### Actual Behavior

bug

### Steps to Reproduce

_Right click at bottom of workspace._

_Windows 10 Google Chrome_

SillyInventor:
### Resolves

Resolves LLK/scratch-gui#111

### Proposed Changes

Calculate div height from as the sum of the heights of all the menu items instead of passing the window height.

### Reason for Changes

Making the menu fit the content prevents scroll bar creation. Possible exception at the very bottom of the screen, but that should be fixed by limiting x & y in a different pull request.

### Test Coverage

None added.

### DIFF
--- a/core/contextmenu.js
+++ b/core/contextmenu.js
@@ -46,6 +46,7 @@ Blockly.ContextMenu.currentBlock = null;
  * @param {boolean} rtl True if RTL, false if LTR.
  */
 Blockly.ContextMenu.show = function(e, options, rtl) {
+  var menuHeight = 0;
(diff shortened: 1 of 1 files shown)


rschamp:
@SillyInventor I think I might be misunderstanding the intent, but this seems to behave the same way as before. Which part of #111 did this intend to solve? Is it a browser-specific fix?

thisandagain:
Closing pending feedback from @SillyInventor

PR #806, opened 2017-02-21
Example use of extensions to define looks blocks

rachel-fenichel:
This is a sample of how to use the new Blockly.Extensions API and the old Blockly.defineBlocksWithJsonArray call to simplify block definitions.

This creates and uses a few simple extensions:
set_color_looks
set_category_looks
set_shape_statement
set_output_string
set_output_number

Looks category before and after this change:

Before:
looks_category_now
After:
looks_category_new

### DIFF
--- a/blocks_vertical/looks.js
+++ b/blocks_vertical/looks.js
@@ -27,592 +27,362 @@ goog.require('Blockly.Colours');
 goog.require('Blockly.constants');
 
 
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
@ericrosenbaum @rschamp @thisandagain FYI

rschamp:
Much cleaner, thank you!

Issue #43, opened 2016-02-02
Design - Zoom controls


thisandagain:
In Scratch on web / desktop we have zoom controls in the bottom right of the editor:
![image](https://cloud.githubusercontent.com/assets/747641/12758819/22f57664-c9ac-11e5-82d5-cb33f461a449.png)

For tablet and desktop / web, we should make a decision as to what we want:
- No controls (no zooming)
- Button controls
- Pinch to zoom controls

One consideration is that "pinch to zoom" will impact the potential for multitouch interaction with blocks.

/cc @tmickel

tmickel:
@carljbowman This is on the list for the prototype, so we'll need a design.

carljbowman:
Assigning both @rachel-fenichel and @picklesrus – Was not sure who to assign as I know there are refactors going on in Blockly that may impact these changes. Feel free to reassign.

_We held off on exploring pinch zoom in touch environments, namely the I/O prototype. As we focus on the desktop experience, pinch zoom will remain out of scope._

### Zoom In/Out Behavior
Overall we should keep Blockly's zoom behavior: zooming towards the center of the workspace viewport. Since the viewport includes the Block Palette and in Scratch Blocks we always have the palette open the zoom feels a bit off. We need to account for the blocks palette width (vertical grammar) or height (horizontal grammar) when zooming.

<img width="557" alt="screen shot 2017-01-17 at 9 14 04 am" src="https://cloud.githubusercontent.com/assets/3409578/22023736/63e4f184-dc95-11e6-8dde-2bafe6586fdd.png">

### Reset Zoom Control Behavior
Currently Blockly does two things when resetting: returns to 100% zoom and repositions the viewport to the center of the blocks. This centering behavior can be a bit disorienting and should be removed. 

Reset should instead just return to 100% zoom and position to the center of the viewport, similarly to how zooming in/out positions.

### Zoom Controls Postion
If this is not already implemented within the code, we should allow the controls to be positionable in the corners of the workspace: top-left, top-right, bottom-left, bottom-right. Default position in Scratch-blocks should be bottom-right (or in RTL bottom-left). 

### Design & Icon Assets
<img width="547" alt="screen shot 2017-01-17 at 9 35 10 am" src="https://cloud.githubusercontent.com/assets/3409578/22024539/499d6b96-dc98-11e6-85e3-50059bf93ce7.png">

[zoom-icons.zip](https://github.com/LLK/scratch-blocks/files/711319/zoom-icons.zip)

[...]

PR #808, opened 2017-02-21
Zoom controls

thisandagain:
### Resolves

GH-43

### Proposed Changes

- Makes trashcan hidden by default
- Updates zoom controls as per #43 and discussion with @carljbowman 
- Replaces use of `sprites.png` with individual SVG assets for each zoom control (easier to maintain)


thisandagain:
Note to self: squash during merge

core/zoom_controls.js
```diff
@@ -40,32 +40,60 @@ Blockly.ZoomControls = function(workspace) {
...
+ * @private
+ */
+Blockly.ZoomControls.prototype.ZOOM_OUT_PATH = 'zoom-out.svg';
+
+/**
+ * Zoom reset icon path.
+ * @type {string}
+ * @private
+ */
+Blockly.ZoomControls.prototype.ZOOM_RESET_PATH = 'zoom-reset.svg';
```
rschamp: Nit on these, private vars conventionally end in `_`

rschamp (changes requested):
Just one syntax thing — tested it out in the GUI and the VM, looks/works great!

thisandagain:
@rschamp Good catch. Fixed.

rschamp (approved):

PR #809, opened 2017-02-21
Dynamically Create Simple Toolbox

TheBrokenRail:
### Proposed Changes

_No loner uses a separate toolbox for simple._

### Reason for Changes

_To make it more dynamic._


rschamp:
@TheBrokenRail No thank you on this one. The default vertical toolbox is a temporary solution while we're working on these separate repos, we do not want automatic generation of toolboxes built-in in the long term.

I would also add that the guideline to have an issue associated with each PR is to prevent a situation like this where you submit a PR we don't have a need for.

Issue #671, opened 2016-10-08
You can select lists with variable blocks and variables with list blocks


TheBrokenRail:
The Title Explains It

dekrain:
So I creating request for use mutations instead of fields

jeandrek:
...Why is there a "feature" tag?

TheBrokenRail:
I don't know It is clearly a bug

TheBrokenRail:
@tmickel Thanks for fixing the tag!

paulkaplan:
Closed by typed variables work.

Issue #572, opened 2016-08-16
"No variable selected" option in variable dropdown


rachel-fenichel:
- [ ] option added to dropdown
- [ ] option automatically set when deleting a variable

tmickel:
Btw, I'd guess we wouldn't actually want this option to be in the dropdown itself/selectable, if that makes sense.

towerofnix:
If, for example, the "no variable selected" option literally has the value of `"NOT_SELECTED"`, if I try to set a variable called `NOT_SELECTED` what will happen? Nothing?

CatsAreFluffy:
The current behavior is ok, except for the fact that it will prioritize local variables. Also deleting a variable and running a script using the variable is the only way to never generate a variable watcher.

towerofnix:
> The current behavior is ok, except for the fact that it will prioritize local variables.

How is that good, or bad, or worse than anything else?

CatsAreFluffy:
Well, you usually create a variable to immediately use it. As long as there are no local variables, or the newest is a local variable, then you an use it directly. However, if it's a global variable and there's a preexisting local variable, then the "set" and "change" and other blocks will show the local variable instead of the global variable, forcing you to use the dropdown.

rachel-fenichel:
@liam4 raises a good point about disallowed names.  We can internationalize the "no variable selected" text, but then we also have to have a variable name that's not allowed that varies by language.

Another option could be to use the empty string for this, since that's not a legal variable name and since the variable value becomes 0 or the empty string when used.

CatsAreFluffy:
> > The current behavior is ok, except for the fact that it will prioritize local variables.

> How is that good, or bad, or worse than anything else?

Convenience. When you create a variable, you usually use it immediately.

Tymewalk:
> Another option could be to use the empty string for this, since that's not a legal variable name and since the variable value becomes 0 or the empty string when used.

@rachel-fenichel That sounds like a good idea. In addition, are there any special characters not allowed in variable names we could use? Is all whitespace allowed?

PR #810, opened 2017-02-22
Feature/variable getter

### LINKED ISSUE #573: Variables - Getter block to match Scratch 2.0
This block should have a single text field, set to the name of the variable, but the right-click menu should behave like the dropdown on a fieldVariable.

rachel-fenichel:
### Resolves
Fixes #573

### Proposed Changes
Adds field_variable_getter, which renders as a label but lists every variable in the right-click menu.  This field updates in response to the right-click menu.

### Reason for Changes
Parity with Scratch 2.0

### Test Coverage
jsunit/field_variable_getter_tests.js

Note that a similar extension will need to be created for list blocks, but it will need to inspect a different list, of all list variables.  Lists and variables are not currently separated in scratch-blocks.  (#671)

### DIFF
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -21,6 +21,7 @@
 'use strict';
 
 goog.provide('Blockly.Blocks.data');
+goog.provide('Blockly.Constants.Data');
 
 goog.require('Blockly.Blocks');
 goog.require('Blockly.Colours');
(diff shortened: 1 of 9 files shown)


rachel-fenichel:
image

rschamp (approved):
Thanks @rachel-fenichel, this looks great, and seems to work well, tested in the GUI.

As you mentioned, the "no variable selected" string should be the empty string, and user-provided strings should be distinct from "system" drop-down items using a prefix or suffix. But I think that can happen in a different PR.

rachel-fenichel:
"No variable selected" is #572.

Adding prefixes is #813.

Issue #816, opened 2017-03-03
Sound effects are too loud relative to Scratch audio


ericrosenbaum:
### Expected Behavior

Ideally the sounds for snapping together and for deleting blocks would be relatively quiet compared to Scratch sounds.

### Actual Behavior

They are a bit too loud.

### Steps to Reproduce

Snap together or delete a block. Compare to sounds from e.g. the "play drum," "play note" blocks.

### Operating System and Browser

Mac OS 10.11.6 Chrome 56

PR #824, opened 2017-03-06
Quieter sound effects for connecting and deleting blocks

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/816

### Proposed Changes

Reduce volume of click and delete sounds by 15dB. 

Also, trim silence from start and end of the files (saving about 75 Kb total).

### Reason for Changes

The sounds were a bit too loud compared to typical Scratch sounds. Now, for example, the sound you hear when connecting two blocks is quieter than the default "pop" sound in the play sound block.

### Test Coverage

No tests.


thisandagain (approved):
🔊 👂

Issue #827, opened 2017-03-07
Visual effects shouldn't take random as input to type


morantsur:
### Expected Behavior

Visual effects blocks should not take a random block as input on the effect type. 

### Actual Behavior

![image](https://cloud.githubusercontent.com/assets/15109618/23673795/94559224-0341-11e7-8d7e-b45964f8dcb1.png)

### Steps to Reproduce

Add a random/join/operation block inside a visual effects block.

rachel-fenichel:
Are there any blocks that should be droppable there?

thisandagain:
No. This was just an oversight on our part I think. This menu should have been implemented as a "rectangle" (non droppable).

thisandagain:
The following blocks have drop-downs that should not accept an input:

### Looks
- `change effect [x] by [n]`
- `set effect [x] to [n]`

### Sound
- `change effect [x] by [n]`
- `set effect [x] to [n]`

### Operators
- `[x] of [n]`

PullJosh:
Would it be possible that a block input for `[x] of [n]` could be useful in the context of a calculator project? (Making it easier to provide built-in calculator functions.)

Probably not a common enough use case to be worth changing the input, but it's at least worth considering.

TheBrokenRail:
Wouldn't keeping them circular make some things easier to code for scratcher, (using variables to set effect, using random to choose a random effect (add access via index support to the blocks), etc).

PR #829, opened 2017-03-09
Do not accept an input for some drop-down menus

thisandagain:
### Resolves

GH-827

### Proposed Changes

- Removes the ability to accept an input from some drop-down menus
    - `motion_setrotationstyle`
    - `looks_changeeffectby `
    - `looks_seteffectto`
    - `sound_changeeffectby`
    - `sound_seteffectto `
    - `operator_mathop`
- Reorder sound effects blocks to match graphic effects sort order
- Minor adjustment to sound effects "clear" block language

### DIFF
--- a/blocks_vertical/default_toolbox.js
+++ b/blocks_vertical/default_toolbox.js
@@ -196,9 +196,6 @@ Blockly.Blocks.defaultToolbox = '<xml id="toolbox-categories" style="display: no
       '</value>'+
     '</block>'+
     '<block type="looks_changeeffectby">'+
-      '<value name="EFFECT">'+
-        '<shadow type="looks_effectmenu"></shadow>'+
-      '</value>'+
       '<value name="CHANGE">'+
         '<shadow type="math_number">'+
           '<field name="NUM">10</field>'+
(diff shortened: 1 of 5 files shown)


ericrosenbaum:
The changes look good!

One requested addition:

The "set rotation style" block also has a drop-down with a fixed set of options that should probably not accept an input, so I suggest making the same change there.

Also, a note:

The < key [space] pressed > boolean reporter does not accept an input in 2.0, but currently in 3.0 it does. I got excited for a minute about this because it could be a feature: you could for example loop through a list of letters, to check if each is pressed, rather than using separate if statements. This already works for letters, but numbers seem to be treated as ascii codes, so they do not work as expected (but as a result checking < key 10 pressed > is true when you press tab!). I'll open a separate issue about these questions.

rachel-fenichel:
Looks good to me.

thisandagain:
@ericrosenbaum Good catch! All set. Give it one more check and then I'd be happy to land this if it looks ok to you. Thanks for filing the new issue about the `key [x] pressed` block. 😄

ericrosenbaum:
Looks good!

PR #830, opened 2017-03-09
Remove partially implemented/unimplemented blocks

rschamp:
Data/list categories are removed to look nice for the playtest.

Resolves LLK/scratch-gui#158


paulkaplan (approved):
LG

PR #832, opened 2017-03-09
Fix borders around the panels

paulkaplan:
Fix the wonky borders to correspond more closely with the mocks. 

Old
screen shot 2017-03-09 at 3 30 05 pm

New
screen shot 2017-03-09 at 3 35 14 pm

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -187,6 +187,7 @@ Blockly.Css.CONTENT = [
     'height: 100%;',
     'position: relative;',
     'overflow: hidden;', /* So blocks in drag surface disappear at edges */
+    'border: 1px solid #ddd;',
   '}',
 
   '.blocklyNonSelectable {',
@@ -439,6 +440,7 @@ Blockly.Css.CONTENT = [
   '.blocklyFlyout {',
     'position: absolute;',
     'z-index: 20;',
+    'border-right: 1px solid #ddd',
   '}',
   '.blocklyFlyoutButton {',
     'fill: #888;',
@@ -528,11 +530,6 @@ Blockly.Css.CONTENT = [
     'text-align: center;',
     'color: $colour_text;',
     'font-weight: 500;',
-  '}',
-
-  '.blocklyMainBackground {',
-    'stroke-width: 1;',
-    'stroke: #c6c6c6;',  /* Equates to #ddd due to border being off-pixel. */
   '}',
 
   '.blocklyMutatorBackground {',
@@ -641,6 +638,9 @@ Blockly.Css.CONTENT = [
     'position: absolute;',
     'font-family: "Helvetica Neue", Helvetica, sans-serif;',
     'z-index: 40;', /* so blocks go over toolbox when dragging */
+    'box-sizing: content-box;',
+    'border-right: 1px solid #ddd;',
+    'border-bottom: 1px solid #ddd;',
   '}',
 
   '.blocklyTreeRoot {',


rschamp:
On Chrome 56 I am seeing scrollbars in the category menu
image

lifeinchords:
Thanks @paulkaplan .. this is also related to #107

rachel-fenichel:
Do you mean https://github.com/LLK/scratch-gui/issues/107 ?

Also, are those blockly scrollbars in the category menu?  And if so, how?

rschamp:
@rachel-fenichel I don't know whose scrollbars those are nor how they got there, but there they are!

rschamp:
They are native scrollbars, apparently because the added 1px of width and height makes the categories overflow.

paulkaplan:
Ah yeah I don't see them because they are hidden when you aren't using a mouse. Will adjust.

lifeinchords:
Ah thanks @rachel-fenichel .. ya, I meant to link to the gui issue, thanks.

paulkaplan:
@rachel-fenichel @rschamp Hmm thoughts on the best way to fix this?

1. Make overflows hidden. Not sure when an overflow would occur, or what controls the sizing to ensure it does/does not. 
2. Change js that controls box sizing to allow one extra px for the border
3. Some css trickery I do not know of? I can't think of anything that would work?

rachel-fenichel:
Calling @picklesrus to say what overflow: hidden would do here.  I think 2 is the correct approach.

rschamp:
+1 to option 2. I believe one version of 3 would be to change `box-sizing` to `content-box`? @lifeinchords?

lifeinchords:
I haven't had a chance to dig into blockly's codebase much yet, so curious to learn more about how it calc's layout, and what constraints are involved. (Hello @rachel-fenichel @picklesrus !)

One thing to share that may be useful, I played with combining `box-sizing: border-box` + css `calc` for widths recently in the `SpriteSelector` layout in Gui. It helped me simplify things and bypass using `overflow: hidden`, which made debugging much easier. Happy to share more if that's useful/relevant.

paulkaplan:
Option 2 makes me worried because it creates a dependency between JS layout code and styling. @rschamp the `content-box` solution works well, so I updated this PR to use that, scrollbar issue is  gone.

picklesrus:
Sorry for the slow response. I like the content box solution - I didn't know about this property. It seems better than the overflow:hidden trick.

@lifeinchords: I'd be happy to learn more. Have a commit I should take a look at?  We'd also be happy to answer questions about layout. Some of the answers will likely unsatisfying... e.g. "because IE version X is a special snowflake" or "I've never paid close attention to that since it has just worked"  but we will do our best :) We'll probably find some stuff we can add to our cleanup list too.

rschamp:
@paulkaplan and I experimented with `box-sizing`, and it would probably work in the short term.  However the current CSS doesn't adapt to other toolbox positions (end, top, bottom), so the right way would be to do this within the toolbox rendering code, where we would then be able to account for the border widths when calculating widths/heights.

lifeinchords:
thanks for sharing thoughts on this @picklesrus. I found `box-sizing` useful in [this commit](https://github.com/LLK/scratch-gui/commit/7f0141e7aff85b775335c00e7a7e9a6fa5c98642), but not sure how readable it is, as I mashed together several tasks/issues in it :(  

Glad to share more on another thread, as it sounds like @rschamp and @paulkaplan suggestions are a better fit for this fix

thisandagain:
@rschamp Ping. What are the next steps on this PR?

rschamp:
We're going to try this again and account for the toolbox position when we do. So, closing for now.

PR #836, opened 2017-03-10
Fix "More Blocks" Flyout and definition block XML export.

TheBrokenRail:
### Resolves

_Part of #607_

### Proposed Changes

_Makes the flyout no longer search for nonexistent blocks, no longer display a useless definition block, makes it display call blocks, and fixes XML export._

### Reason for Changes

_So the "More Blocks" category contains the currently defined custom blocks._


thisandagain:
Hello @TheBrokenRail ! Please only submit PRs for issues that have the "help wanted" label or you have discussed with a member of the Scratch Team. Many thanks.

Issue #843, opened 2017-03-13
"Point in direction" angle picker makes degree symbol appear (e.g. 15º)


ericrosenbaum:
### Expected Behavior

No degrees symbol should appear inside the input for the "point in direction" block. 

### Actual Behavior

Sometimes a degrees symbol appears as part of the number in the input, e.g. -60º

### Steps to Reproduce

Drag out a "point in direction block"

Click the degree input to open the angle picker widget

Rotate the angle picker counter-clockwise past 0, then dismiss the angle picker. Sometimes (!) the degree symbol appears: 

<img width="180" alt="screen shot 2017-03-13 at 1 11 46 pm" src="https://cloud.githubusercontent.com/assets/567844/23867201/8c299122-07f1-11e7-892b-565f4d967e1e.png">

### Operating System and Browser

Mac OS 10.11.6, Chrome 56

rachel-fenichel:
Possibly related to https://github.com/google/blockly/issues/973

rachel-fenichel:
Getting rid of the degree symbol entirely is easy, if that's the desired behaviour.  Just take out the appropriate code from FieldAngle:setText.

The "sometimes" on the degree symbol appearing is probably related to this bug: https://github.com/LLK/scratch-blocks/issues/581

![e3b3ab15-4e99-4900-b1c3-27cafc0c3ad7](https://cloud.githubusercontent.com/assets/13686399/23873743/4fa479a2-07f0-11e7-997f-aa61b3bc19c1.gif)

I could consistently make it happen by moving my mouse while the input was closing, and could consistently make it not happen by leaving my mouse still.  The difference between those two is probably a setText vs. setValue problem (again).

rachel-fenichel:
Possibly fixed here: https://github.com/google/blockly/pull/1044

CatsAreFluffy:
Is this bad though?

CleoQc:
I like the little degree symbol.

CatsAreFluffy:
Yeah, people who know radians but not programming might be confused, and I think the "low-floor" policy applies here :P

towerofnix:
Agreed, though, of course, the degree system has to *always* be visible in that case – regardless of whether the input is `0`, a typed in number, or a degree-picker number.

[...]

PR #844, opened 2017-03-14
Use one mousemove handler and remove it when an angle is selected by …

sclements:
…clicking.

### Resolves

https://github.com/LLK/scratch-blocks/issues/843

### Proposed Changes

Remove the mouse move listener when an angle is picked by clicking.

### Reason for Changes

The selection element shouldn't be interactive after clicking.


rschamp (approved):
Hi @sclements, thanks for your contribution!

This works well locally and the fix looks sensible to me.

PR #848, opened 2017-03-17
Fix #826 by forcing a re-render of the field when the text input is d…

### LINKED ISSUE #826: Horizontal block parameter numerical selection does not display correctly
### Expected Behavior

In a horizontal project, when I enter a number to set the parameter for blocks such as the loop or timer blocks, the number I enter should display in the selected number area at the bottom of the script. This should work for both desktop browsers as well as on mobile devices that show a number dropdown.

### Actual Behavior

After I enter the number, the previously shown number still shows.

### Steps to Reproduce

1. Load horizontal_playground.html on a desktop browser (such as Chrome 56.0 on a Mac)
2. Drag the loop block to the stage
3. At the bottom of the loop block, select the default '4' and type '9' (to replace the '4')
4. Hit return or click away from the block
5. Observe that the number changes back to 4

Mac OS 10.11.6 Chrome 56.0

picklesrus:
…isposed.  Rendering the whole block does not necessarily cause the field to render.

### Resolves

https://github.com/LLK/scratch-blocks/issues/826

### Proposed Changes

Render the field text input when the user is done editing a text input since the new text has been set.

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -489,7 +489,8 @@ Blockly.FieldTextInput.prototype.widgetDispose_ = function() {
       }
     }
     thisField.setText(text);
-    thisField.sourceBlock_.rendered && thisField.sourceBlock_.render();
+    // Rerender the field now that the text has changed.
+    thisField.sourceBlock_.rendered && thisField.render_();
     Blockly.unbindEvent_(htmlInput.onKeyDownWrapper_);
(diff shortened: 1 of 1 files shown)


picklesrus:
@rachel-fenichel, we should decide if we want to make this change upstream as well. We don't necessarily need to because it doesn't cause a bug for us...  but consistency is nice :)

thisandagain (approved):

Issue #823, opened 2017-03-06
Match duplication behavior from Scratch 2.0


thisandagain:
### Expected Behavior

- I expect to be able to duplicate a stack of blocks (e.g. Scratch 2.0)
- When selecting "duplicate" while right clicking a block, that block plus all blocks connected below it (regardless or nesting) should be duplicated (e.g. Scratch 2.0)

### Actual Behavior

- Upon selected "Duplicate" from the context (right click) menu, a single block is duplicated. This duplicated block is not connected to my mouse cursor but is rather added to the workspace directly

### Repro

I am able to repro this by creating a stack of blocks with a loop and other blocks, e.g. screenshot below. I can duplicate each block and the contents of the for-loop, but not the whole stack.

<img width="283" alt="screen shot 2017-03-03 at 2 55 23 pm" src="https://cloud.githubusercontent.com/assets/1320456/23566802/749ec2de-0021-11e7-963d-f5b4636cda7c.png">

### Operating System and Browser

_Mac OS X El Capitan 10.11.6, Chrome Version 56.0.2924.87 (64-bit)_

---
Moved from LLK/scratch-gui#165
/cc @jwzimmer

RoboErikG:
This is also something we want in Blockly and relates to [this issue](https://github.com/google/blockly/issues/384) around Hat Blocks wrapping the blocks below them for code generation.

thisandagain:
@rschamp @carljbowman @RoboErikG @picklesrus @rachel-fenichel I made some refinements to the description above to clarify what I think is the desired behavior. Any thoughts?

rachel-fenichel:
Can we separate out the issue of attaching duplicated blocks to the mouse cursor?

thisandagain:
## Example

<img width="1280" alt="screen shot 2017-03-07 at 3 02 20 pm" src="https://cloud.githubusercontent.com/assets/747641/23675504/95ff109a-0347-11e7-8dfc-ecef69d71b7a.png">
<img width="1392" alt="screen shot 2017-03-07 at 3 02 26 pm" src="https://cloud.githubusercontent.com/assets/747641/23675503/95fe0d8a-0347-11e7-8384-7937a0b8a366.png">
<img width="1392" alt="screen shot 2017-03-07 at 3 02 31 pm" src="https://cloud.githubusercontent.com/assets/747641/23675505/95ff4f4c-0347-11e7-9f6a-bae693954082.png">
[...]

Issue #828, opened 2017-03-07
Duplicated blocks should be attached to the mouse cursor


thisandagain:
### Expected Behavior

- After duplication the new blocks should be attached to the mouse cursor (e.g. Scratch 2.0)

### Actual Behavior

- Blocks are directly added to the workspace

rachel-fenichel:
How does this work on touch?

thisandagain:
I don't think we'll have duplication on touch.

rachel-fenichel:
As in, you won't have it in the context menu as an option?

thisandagain:
I think we are leaning towards not having any context menus for touch environments.

towerofnix:
> I don't think we'll have duplication on touch.

Well, sure, but there's still the "stamp"/duplicate button:

![screen shot 2017-03-07 at 4 50 57 pm](https://cloud.githubusercontent.com/assets/9948030/23677381/44005f24-0356-11e7-8358-155e4869f7cc.png)

Duplicating blocks is an incredibly useful feature. Removing all ways of using it would quite reduce productivity – for instance, implementing the distance formula as this would be much more inconvenient:

![](https://cloud.githubusercontent.com/assets/9948030/23677532/e47f2192-0356-11e7-9fbf-a578a01315d8.png)

Making this just once would already be more tedious without block duplicating functionality, and many users simply don't know about custom blocks – so they'd have to rewrite this entire piece of code, all by hand (..literally.. 😛 ), whenever they want to get the distance between two points!

And it's not like that's the only use case of duplication. Suppose I have a script such as this one:

![demo](https://cloud.githubusercontent.com/assets/9948030/23677855/306ef5cc-0358-11e7-9170-93a87bf2b5cd.png)

Well, you know, it's a relatively complicated script. But it's definitely a realistic one, for a game!

I used duplicate multiple times while making that:

* When using the "initialize enemy class" block.
  * While making the first block, "initialize (blue slime)", I would have used copy/paste to duplicate the "(number) * difficulty" block.
  * While adding more enemies, I used copy/paste to duplicate the entire "initialize enemy" block, which contained the "(number) * difficulty" blocks.
* When making the "initialize enemy class" block's script.
  * I duplicated the "add (block) to enemy class data" block multiple times.
* In the future, when adding more locations.
  * I would duplicate the "set (location ID) to" and "set (location difficulty) to" blocks, to keep track of the variables I'd need to set up for each location.

[...]

PR #850, opened 2017-03-17
Copy entire stack on duplicate

paulkaplan:
Instead of copying a single block, copy the entire stack of connected blocks. Fixes https://github.com/LLK/scratch-blocks/issues/823 by @rachel-fenichel's final suggestion of just removing the `if` block. I left it in there commented out because I thought it would make upstream merges easier, but @thisandagain I'm not sure if that is the right way. 

Pinning the new blocks to the cursor is still open in https://github.com/LLK/scratch-blocks/issues/828

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -292,9 +292,9 @@ Blockly.terminateDrag_ = function() {
  */
 Blockly.copy_ = function(block) {
   var xmlBlock = Blockly.Xml.blockToDom(block);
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
I'd prefer you to remove the code rather than commenting it out.

paulkaplan:
@rachel-fenichel fixed 83c793f

rachel-fenichel:
looks good to me.

thisandagain (approved):

Issue #796, opened 2017-02-16
Loading a new toolbox without categories through updateToolbox causes console error


MichaelZawadzki:
### Expected Behavior

When the workspace is initialized with a non-category toolbox it is expected to be able to load other non-category toolboxes through workspace.updateToolbox without causing console errors. 

### Actual Behavior

With the latest merge of **bugfix/updateToolbox**, this now causes a console error. Before (at SHA 1cf79eed97015b5bddcc628cc24a1be079a4d3f1) this did not occur. 

### Steps to Reproduce

To reproduce, add a method like the following to the playground to trigger a toolbox refresh:

`function setToolbox() {
          var emptyToolbox = '<xml><toolbox></toolbox></xml>';
          workspace.updateToolbox(emptyToolbox);
      };
`

This method can then be trigged by any button, for example the "To JavaScript" button.

Note that you need to first make sure to set the toolbox option to "Simple" in order to have the toolbox with no categories loaded initially (to avoid switching between category toolboxes and non-category toolboxes).

Also note that the toolbox does not need to be empty, its just like this in the example to keep it simple. 

### Operating System and Browser

Reproduced on Chrome on Mac but pretty sure its not isolated to any specific platform and is a general issue.

MichaelZawadzki:
Seems this might simply need a 'if (this.toolbox_)' check around that position call.

rschamp:
Fixed in #858

PR #858, opened 2017-03-29
[Issue-796] adding undefined check before trying to position category…

MichaelZawadzki:
Adding a check for undefined for the case when no categories are defined in the toolbox and updateToolbox is called. 

### Resolves

This resolves issue 796: https://github.com/LLK/scratch-blocks/issues/796
This issue was introduced in this PR: https://github.com/LLK/scratch-blocks/pull/793

### Proposed Changes

Adding an undefined check before attempting to call position() on this.toolbox_, which only exists in the case that the toolbox has categories.

### Reason for Changes

To avoid a console error.

### Test Coverage

No tests added as its fixing existing functionality. 

### Questions

Any naming conventions around branch naming? Should I use bugfix/* next time? Seems this doesn't matter though. 
I am also NOT including a post-compiled version of the lib as I see other PR's don't do this.
Also wondering if I need to link this PR to the original issue somehow, or if github somehow magically does this.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1475,15 +1475,13 @@ Blockly.WorkspaceSvg.prototype.updateToolbox = function(tree) {
     }
     this.options.languageTree = tree;
     this.toolbox_.populate_(tree);
+    this.toolbox_.position();
   } else {
     if (!this.flyout_) {
       throw 'Existing toolbox has categories.  Can\'t change mode.';
     }
     this.options.languageTree = tree;
     this.flyout_.show(tree.childNodes);
-  }
-  if (this.toolbox_) {
-    this.toolbox_.position();
   }
 };
 
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1482,7 +1482,9 @@ Blockly.WorkspaceSvg.prototype.updateToolbox = function(tree) {
     this.options.languageTree = tree;
     this.flyout_.show(tree.childNodes);
   }
-  this.toolbox_.position();
+  if (this.toolbox_) {
+    this.toolbox_.position();
+  }
 };
 
 /**


core/workspace_svg.js
```diff
@@ -1482,7 +1482,9 @@ Blockly.WorkspaceSvg.prototype.updateToolbox = function(tree) {
     this.options.languageTree = tree;
     this.flyout_.show(tree.childNodes);
   }
-  this.toolbox_.position();
```
rachel-fenichel, line 1485: move `this.toolbox_.position();` to just after line 1477.

Context: the if statement on line 1472 checks whether this is a toolbox with categories, and if so it populates the toolbox.  Positioning the toolbox is a reasonable followup to that, and means you don't need an extra if.

You may also need to call `this.flyout_.position()` after line 1483, which is the equivalent for the non-category toolbox.

rachel-fenichel:
Branch naming conventions: the Scratch team uses git flow: http://nvie.com/posts/a-successful-git-branching-model/

Mentioning the original issue by number creates a link in github and the issue will reflect that you mentioned it here.

No need to compile the library for PRs like this--we'd just have a ton of churn on those files.

core/workspace_svg.js
MichaelZawadzki: I see. Happy to move the call to right after line 1477 to avoid the extra if.

As for the call to positioning the flyout... I'm hesitant to add something when there is no clear purpose. Is there any behavior you are concerned about might be bugged without that call? Logically, as I understand it, loading a new set of blocks into the toolbox should not actually reposition it. There is no option to switch between toolbox layouts here, so its not like it will shift anchors (aka top, left, etc). So it seems the toolbox position would not change. One edge-case might be the scroll-bar, as loading a new set of blocks might adjust how it looks given the different number of blocks... but testing this it seems to work just fine: I can see the scrollbar adjust according to the blocks in the toolbox after loading. So, again, I'm hesitant to blindly add a call when I don't see a reason nor can I test if anything has changed, but if you want me to I'm happy to do it (but would prefer not to). Please let me know, I'll make the other change for now.

rachel-fenichel: Hm yeah, #793 made it so that the entire toolbox DOM is recreated when populating it, which isn't true for the flyout.  So we shouldn't need to call position.

Alrighty, I'm happy with this PR.  Thanks!

rachel-fenichel:
@rschamp LGTM, let me know if you have opinions.

rschamp (approved):
Sorry for the delay. This looks good to me too. Thanks @MichaelZawadzki!

PR #866, opened 2017-04-07
Initialize local scrollX variable.

picklesrus:
We were using a local scrollX variable in the vertical flyout without initializing it.
https://github.com/LLK/scratch-blocks/commit/2ea8bc21d284d35800c2be0f01a62f460d035357 commented out the initialization a while back and probably didn't mean to.

I found this because the closure compiler complained about it.

@rachel-fenichel, I'm happy to remove the line above where your commit commented out the original initialization but wanted to check why you left it there first.


rachel-fenichel:
lgtm

Issue #853, opened 2017-03-21
horizontal's dropdown generate javascript code


zhlella:
### Expected Behavior

horizontal's dropdown generate javascript code

### Actual Behavior
![11](https://cloud.githubusercontent.com/assets/16182857/24136561/59d25de4-0e4a-11e7-94c6-0610464e7e16.png)

when generate javascript code, console show:

![12](https://cloud.githubusercontent.com/assets/16182857/24136570/656010d4-0e4a-11e7-8a05-92e0ffc67b0a.png)

### Steps to Reproduce
XML:
![13](https://cloud.githubusercontent.com/assets/16182857/24136580/6fa95b72-0e4a-11e7-938b-5def25d9ce2d.png)

Json:

`Blockly.Blocks['light_new'] = {
  init: function() {
    this.jsonInit({
      "id": "light_new",
      "message0": "%1 %2",
      "args0": [
        {
          "type": "field_image",
          "src": Blockly.mainWorkspace.options.pathToMedia + "icons/set-led_mystery.svg",
          "width": 40,
          "height": 40,
          "alt": ""
        },
        {
          "type": "input_value",
          "name": "CHOICE"
        }
      ],
      "inputsInline": true,
      "previousStatement": null,
      "nextStatement": null,
      "category": Blockly.Categories.looks,
      "colour": Blockly.Colours.more.primary,
      "colourSecondary": Blockly.Colours.more.secondary,
      "colourTertiary": Blockly.Colours.more.tertiary
    });
  }
};`

![15](https://cloud.githubusercontent.com/assets/16182857/24136585/74b25d58-0e4a-11e7-810a-80d76a183685.png)

Javascript code :
![16](https://cloud.githubusercontent.com/assets/16182857/24136587/79bedccc-0e4a-11e7-9cdc-0f191012b713.png)

### Operating System and Browser

windows 7 & chrome 56.0.2924.87

TheBrokenRail:
Make sure the compressed_javascript.js file is included in your website and if you are running this in node see #856

cwillisf:
Closing this for now: we're planning to work on horizontal blocks some time next year.

PR #869, opened 2017-04-14
Upload Generators To NPM

TheBrokenRail:
### Resolves

_#856 & #853 (If you are using it in Node.js)._

### Proposed Changes

_Stops ignoring Generators when uploading to NPM._

### Reason for Changes

_So generators can be used._


rschamp:
Good catch, thanks @TheBrokenRail!

Issue #535, opened 2016-07-14
Design for "create variable" button and dialog


rachel-fenichel:
Placeholders right now:

![image](https://cloud.githubusercontent.com/assets/13686399/16853924/79b4bfae-49c3-11e6-95d0-f44eadb45a47.png)

![image](https://cloud.githubusercontent.com/assets/13686399/16853949/9f7c6606-49c3-11e6-9dd6-73d1fb73c0d0.png)

carljbowman:
Todo - 
- [ ] Document 2.0 Variable Deletion Behavior
- [ ] Document 2.0 Variable Editing/Renaming Behavior
- [ ] Document 2.0 Dragging Variables to Sprites/Stage Behavior

# Scratch 2.0 - Creating a Variable Behaviors

## Clicking 'Create a Variable' opens dialog
![open-dialog mov](https://cloud.githubusercontent.com/assets/3409578/23029551/f1297550-f438-11e6-8187-9e481ce1b063.gif)

## Ability to Name a Variable
![name-var mov](https://cloud.githubusercontent.com/assets/3409578/23029651/42f14e44-f439-11e6-9a6c-c8375be6b4cf.gif)
- Variables cannot have the same name (case sensitive), if the user tries to name a Variable with the same name an error will be thrown
![name error mov](https://cloud.githubusercontent.com/assets/3409578/23031038/98a545b2-f43d-11e6-8f0c-82b324115bc4.gif)
![screen shot 2017-02-16 at 11 47 23 am](https://cloud.githubusercontent.com/assets/3409578/23031063/b15c3d4a-f43d-11e6-9378-d84f793f73e5.png)
- A user can create two 'For this sprite only' variables that have same name across two different sprites

## Ability to Choose if the Variable is 'For all sprites' or 'For this sprite only'
![var global or local mov](https://cloud.githubusercontent.com/assets/3409578/23029868/e60d2be8-f439-11e6-8c52-c64489277991.gif)

- **For all sprites (aka Global):** The Variable that is created will be accessible in the 'Data' category on all sprites
![global var mov](https://cloud.githubusercontent.com/assets/3409578/23030254/3a28ccea-f43b-11e6-9dc0-a34091e27282.gif)

- **For only this sprite (aka Local):** The Variable that is created will only be accessible in the 'Data' category on the sprite it was created. (Unless manually dragged to another block)
[...]

PR #872, opened 2017-04-20
Add back variables

paulkaplan:
Adds back the data category with variable blocks and restyles the button for adding variables.

The "new variable" creation prompt will be handled through the GUI, so it isn't part of this.

screen shot 2017-04-20 at 1 37 32 pm

### DIFF
--- a/blocks_vertical/default_toolbox.js
+++ b/blocks_vertical/default_toolbox.js
@@ -341,6 +341,8 @@ Blockly.Blocks.defaultToolbox = '<xml id="toolbox-categories" style="display: no
       '</value>'+
     '</block>'+
   '</category>'+
+  '<category name="Data" colour="#FF8C1A" secondaryColour="#DB6E00" custom="VARIABLE">' +
+  '</category>' +
   '<category name="Events" colour="#FFD500" secondaryColour="#CC9900">'+
     '<block type="event_whenflagclicked"></block>'+
     '<block type="event_whenkeypressed">'+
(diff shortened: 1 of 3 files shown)


towerofnix:
Note the button label in 2.0 is "Make a Variable", not "Create variable...", and that the 3.0 design also goes with "Make a Variable" – see @carljbowman's post [here](https://github.com/LLK/scratch-blocks/issues/535#issuecomment-280391920) and this design image:

"Make a Variable" button

That said, the required change would be in the `en.json` translation file (`NEW_VARIABLE`).

paulkaplan:
@liam4 yup I noticed that but was going to change it separately. @rachel-fenichel there are a few places where `NEW_VARIABLE` is defined, do I just change all of them?

rachel-fenichel:
To avoid merge conflicts, and since you're going to have separate buttons for data and lists, you should 
- add a new entry in msg/messages.js (such as Blockly.Msg.NEW_DATA)
- use that message in place of NEW_VARIABLE
- run build.py

msg/messages.js is used to build some files with all of our messages in a format that translatewiki can parse.
The .json files for various languages come from translatewiki and have everything they've translated for a given language.  The .js files for languages are built based on the .json files, falling back to english if a translation is lacking.

rschamp (approved):
@paulkaplan This looks great, thanks.  I think you can merge without considering that text for now, since we're going to be addressing localization soon anyway.

Issue #873, opened 2017-04-21
Scrolling toolbox upwards through dragging is broken


MichaelZawadzki:
### Expected Behavior

When dragging on the background of the toolbox, it is expected that the toolbox properly scrolls up and down. 

### Actual Behavior

Scrolling down works fine. But once the toolbox is scrolled down, it can no longer be scrolled up (by dragging on the background). Additionally, the scrolling behavior gets wonky in general and its hard to navigate the toolbox by the user. 

### Steps to Reproduce

We can reproduce this in the vertical playground on develop:

1) Checkout develop branch
2) Open the vertical playground
3) Start a drag operation on the toolbox: click-down on any empty space in the toolbox and drag the mouse down -> toolbox will properly scroll downwards. Scroll all the way down and release.
4) Now we will want to perform the same operation and simply drag back up. So: click on any empty space in the toolbox, drag upwards -> nothing happens! BUG
5) Additionally, at this point these operations, dragging the toolbox to scroll it become very wonky. 

This bug got introduced with this commit: https://github.com/LLK/scratch-blocks/commit/06fa507dbbb989df09e7f18b8dfed475d8af664d. It is the 1st commit on which this bug occurs, any previous SHA's on develop do not have this bug. Any way to tag user bigeyex on this here?

### Operating System and Browser

Mac OS, both on Safari and Chrome

rachel-fenichel:
@bigeyex 
@thisandagain I would recommend reverting that commit for now.

bigeyex:
Hmm, interesting. Sorry for introducing this bug.
I proposed another solution in [pr874](https://github.com/LLK/scratch-blocks/pull/874), 
using the scrollTop instead of viewTop when calculating scrollbar position when dragging.
This will solve the "nudge a little when refreshing" without introducing new scroll problems.
How about that? Return what belongs to Caesar back to Caesar...

thisandagain:
Thanks for reporting @MichaelZawadzki and thanks for the quick turnaround on a patch @bigeyex. 

@rschamp do you mind taking a look at this today? I am also able to reproduce the issue by attempting to scroll the blocks palette using two finger (drag) scroll on a MacBook Pro.

rschamp:
Fixed in #874

MichaelZawadzki:
Yup seems fixed now (even with that original offending line in there). So looks good from a quick test. We'll report if we see any issues as we use/test more with the latest version. Thanks!

PR #874, opened 2017-04-24
Fixes dragging failure in blocks palette

bigeyex:
as said in issue #873, which in introduced by commit 06fa507.

The change is using workspace.scroll instead of viewTop to calculate the position of the scrollbar.
Because the top of the view and the scrollbar position are two different things.

### Resolves

[issue #873](https://github.com/LLK/scratch-blocks/issues/873#issuecomment-296300791)

### Proposed Changes

Use scroll bar properties directly to calculate the scrollbar position

### DIFF
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -343,7 +343,7 @@ Blockly.VerticalFlyout.prototype.wheel_ = function(e) {
       delta *= 10;
     }
     var metrics = this.getMetrics_();
-    var pos = metrics.viewTop + delta;
+    var pos = -this.workspace_.scrollY + delta;
     var limit = metrics.contentHeight - metrics.viewHeight;
     pos = Math.min(pos, limit);
     pos = Math.max(pos, 0);
@@ -581,7 +581,7 @@ Blockly.VerticalFlyout.prototype.onMouseMove_ = function(e) {
   }
   var dy = e.clientY - this.startDragMouseY_;
   this.startDragMouseY_ = e.clientY;
-  var y = metrics.viewTop - dy;
+  var y = -this.workspace_.scrollY - dy;
   y = goog.math.clamp(y, 0, metrics.contentHeight - metrics.viewHeight);
   this.scrollbar_.set(y);
 };


rschamp (approved):
This does appear to fix the issue without introducing anything that I can find. @rachel-fenichel Does this look reasonable to you? Anything else we should consider?

core/flyout_vertical.js
```diff
@@ -343,7 +343,7 @@ Blockly.VerticalFlyout.prototype.wheel_ = function(e) {
       delta *= 10;
     }
     var metrics = this.getMetrics_();
-    var pos = metrics.viewTop + delta;
+    var pos = -this.workspace_.scrollY + delta;
```
rachel-fenichel, line 346: Can you explain why you went from a positive value (metrics.viewTop) to a negative value?

rachel-fenichel (commented):
Does this work when you resize the page?  Does it work when you scroll the page (not scroll within blockly) and then try to scroll the flyout?

rachel-fenichel:
I'm leery of scrollbar changes that aren't fully understood, because they can lead to really nasty recursion.

In particular, scrollbar.set calls scrollbar.onScroll, which calls workspace.setMetrics.

rschamp:
@rachel-fenichel It works as expected after resizes and page scrolls.  I see possible recursion if we ever end up calling scrollbar.set on metrics changes. At least, in this PR I don't see any _new_ dangers of recursion being introduced, since it's only reading `scrollY`.

Am also curious about the origins of the negative sign.

bigeyex:
Ah, this is what's happening here.
image

in the original version, viewTop is assigned from negative scrollY, which is a negative value.
But the issue is, the actual top of the view content is not strictly equal to the scrollY - the optionBox itself has a y offset. This will cause the box shifts up a little every time it's refreshed.

In commit 06fa507, I changed the way how viewTop is calculated, but it breaks scrolling, since viewTop is also used in scroll procedures where scrollY is the value that should be used.

This is where this PR comes from.

rachel-fenichel:
@rschamp LGTM, but let's keep an eye out for any scrollbar-related instability.

PR #881, opened 2017-05-08
Merge google/blockly, May 2017

paulkaplan:
**In Progress**

Merge google/blockly develop into scratch-blocks develop.

Issues that need to be sorted before merging:
- [x] webpack command does not properly build scratch-blocks for the GUI /cc @rschamp, error is `Cannot find module "./shim/blockly_compressed_vertical-blocks_compressed`
- [x] Fix duplicate function in `core/workspace.js` (`allInputsFilled`)
- [x] Fix duplicate function in `core/field.js` (`getDisplayText_`)


.travis.yml
```diff
@@ -1,35 +0,0 @@
-language: node_js
```
thisandagain: Hmm. Is this removing our entire Travis config?

README.md
```diff
@@ -1,24 +0,0 @@
-# scratch-blocks
```
thisandagain: Looks like this is removing our README

core/utils.js
```diff
@@ -788,7 +813,7 @@ Blockly.utils.wrapToText_ = function(words, wordBreaks) {
  * @param {string} text The actual text to measure
  * @return {number} Width of the text in px.
  */
-Blockly.measureText = function(fontSize, fontFamily, fontWeight, text) {
+Blockly.utils.measureText = function(fontSize, fontFamily, fontWeight, text) {
```
thisandagain, line 817: Because this and `encodeEntities` are used in Scratch Blocks but do not exist in Blockly I recommend that we add notes in the comments to help with future merges.

thisandagain (changes requested):
Quick first pass. I'll take another more in-depth look once the webpack issues are resolved.

core/block.js
```diff
@@ -1415,15 +1388,14 @@ Blockly.Block.newFieldImageFromJson_ = function(options) {
   var height =
     Number(Blockly.utils.replaceMessageReferences(options['height']));
   var alt = Blockly.utils.replaceMessageReferences(options['alt']);
-  var flip_rtl = Blockly.utils.replaceMessageReferences(options['flip_rtl']);
-  return new Blockly.FieldImage(src, width, height, alt, flip_rtl);
+  return new Blockly.FieldImage(src, width, height, alt);
```
rachel-fenichel: I think you want to keep flip_rtl

core/block_svg.js
```diff
@@ -592,6 +595,63 @@ Blockly.BlockSvg.prototype.clearTransformAttributes_ = function() {
 };
 
 /**
+ * Transforms a block by setting the translation on the transform attribute
```
rachel-fenichel: Something's wrong here.  These should not be completely new functions in scratch-blocks.

core/blockly.js
```diff
@@ -349,6 +349,20 @@ Blockly.hideChaff = function(opt_allowToolbox) {
 };
 
 /**
+ * When something in Blockly's workspace changes, call a function.
+ * @param {!Function} func Function to call.
+ * @return {!Array.<!Array>} Opaque data that can be passed to
+ *     removeChangeListener.
+ * @deprecated April 2015
+ */
+Blockly.addChangeListener = function(func) {
```
rachel-fenichel: Looks like scratch-blocks had removed this (since it's deprecated).  You probably don't want to add it back.

core/css.js
```diff
@@ -649,7 +672,7 @@ Blockly.Css.CONTENT = [
     'overflow-y: auto;',
     'position: absolute;',
     'font-family: "Helvetica Neue", Helvetica, sans-serif;',
-    'z-index: 40;', /* so blocks go over toolbox when dragging */
+    'z-index: 70;', /* so blocks go under toolbox when dragging */
```
rachel-fenichel: Scratch-blocks may have deliberately changed this.  Please double-check.

core/field.js
```diff
@@ -547,6 +554,31 @@ Blockly.Field.prototype.getDisplayText_ = function() {
 };
 
 /**
+ * Get the text from this field as displayed on screen.  May differ from getText
```
rachel-fenichel: This is not a new function.  Needs to be merged, not duplicated.

core/field_angle.js
```diff
@@ -153,7 +152,8 @@ Blockly.FieldAngle.prototype.showEditor_ = function() {
   this.line_ = Blockly.utils.createSvgElement('line',
       {'x1': Blockly.FieldAngle.HALF,
       'y1': Blockly.FieldAngle.HALF,
-      'class': 'blocklyAngleLine'}, svg);
+      'class': 'blocklyAngleLine',
```
rachel-fenichel: Looks like this change was not properly merged: https://github.com/google/blockly/pull/1044/files

rachel-fenichel (changes requested):

rachel-fenichel:
Also check for changes in the Blockly flyout.  The structure of scratch-blocks (flyout_base, flyout_horizontal, and flyout_vertical) is sufficiently different that git doesn't always catch changes.

On my todo list: switching blockly to the same structure, which will hopefully help with merges + lead to clearer code.

core/block.js
thisandagain: It looks like it was removed from upstream and it doesn't appear that we are using it. Any specific reasons to keep it?

core/block_svg.js
thisandagain: Looking at `develop` this doesn't look like the first time we've had the problem: I see three definitions for `Blockly.BlockSvg.prototype.translate`.

core/block_svg.js
```diff
@@ -592,6 +595,63 @@ Blockly.BlockSvg.prototype.clearTransformAttributes_ = function() {
...
+      'translate(' + x + ',' + y + ')');
+};
+
+/**
+ * Move this block to its workspace's drag surface, accounting for positioning.
+ * Generally should be called at the same time as setDragging_(true).
+ * Does nothing if useDragSurface_ is false.
+ * @private
+ */
+Blockly.BlockSvg.prototype.moveToDragSurface_ = function() {
```
thisandagain: We already have two of these. 👻

core/block_svg.js
```diff
@@ -592,6 +595,63 @@ Blockly.BlockSvg.prototype.clearTransformAttributes_ = function() {
...
+  this.workspace.blockDragSurface_.setBlocksAndShow(this.getSvgRoot());
+};
+
+/**
+ * Move this block back to the workspace block canvas.
+ * Generally should be called at the same time as setDragging_(false).
+ * Does nothing if useDragSurface_ is false.
+ * @private
+ */
+Blockly.BlockSvg.prototype.moveOffDragSurface_ = function() {
```
thisandagain: ... and three of these.

core/block.js
rachel-fenichel: Because you're still using it: https://github.com/LLK/scratch-blocks/blob/develop/blocks_horizontal/event.js#L49

core/block_svg.js
```diff
@@ -592,6 +595,63 @@ Blockly.BlockSvg.prototype.clearTransformAttributes_ = function() {
...
+  this.translate(xy.x, xy.y);
+  this.workspace.blockDragSurface_.clearAndHide(this.workspace.getCanvas());
+};
+
+/**
+ * Clear the block of transform="..." attributes.
+ * Used when the block is switching from 3d to 2d transform or vice versa.
+ * @private
+ */
+Blockly.BlockSvg.prototype.clearTransformAttributes_ = function() {
```
thisandagain: ... and three of these.

core/field.js
thisandagain: This also looks like a recurring issue. We currently have this already defined twice as well.

core/workspace.js
```diff
@@ -490,8 +491,8 @@ Blockly.Workspace.prototype.fireChangeListener = function(event) {
  */
 Blockly.Workspace.prototype.getBlockById = function(id) {
   var block = this.blockDB_[id];
-  if (!block && this.getFlyout() && this.getFlyout().getWorkspace()) {
-    block = this.getFlyout().getWorkspace().blockDB_[id];
+  if (!block && this.getFlyout_() && this.getFlyout_().getWorkspace()) {
```
rachel-fenichel: this.getFlyout_ will be undefined if this is an unrendered workspace.

Either leave getFlyout defined in Blockly.Workspace (and rename it to getFlyout_) or move this definition of getBlockById into WorkspaceSvg and make the Workspace definition a simple lookup in the blockDB_.

You probably aren't getting bitten by this right now because you're not using headless scratch-blocks.  I don't have a strong opinion on whether you should fix it in this merge or in a separate PR--if the latter, please make an issue about it.

core/block.js
thisandagain: Hmm. I think we should remove that and track upstream. @carljbowman or @paulkaplan any thoughts?

thisandagain: PR where it was introduced (via @rachel-fenichel ): https://github.com/LLK/scratch-blocks/commit/470d10af102319fbc1a97ef7f34f84ea599333f7

thisandagain: Actually. Ignore me. @rachel-fenichel is right. We need this to handle the direction of the repeat / forever icons in both horizontal and vertical grammar. This should be restored.

core/block.js
```diff
@@ -1388,7 +1388,8 @@ Blockly.Block.newFieldImageFromJson_ = function(options) {
   var height =
     Number(Blockly.utils.replaceMessageReferences(options['height']));
   var alt = Blockly.utils.replaceMessageReferences(options['alt']);
-  return new Blockly.FieldImage(src, width, height, alt);
+  var flip_rtl = Blockly.utils.replaceMessageReferences(options['flip_rtl']);
```
rachel-fenichel: flip_rtl should always be a boolean, not a message reference, so this should be more like [this](https://github.com/google/blockly/blob/master/core/block.js#L1227) code:

```
  if (typeof options['spellcheck'] == 'boolean') {
    field.setSpellcheck(options['spellcheck']);
  }
```

thisandagain:
This is looking really good @paulkaplan. Thanks for pulling this together. 

I am able to build and testing in both the horizontal and vertical playgrounds looks good so far. 👍

thisandagain (approved):

rachel-fenichel:
Looks good to me as well, aside from the getFlyout_ issue that I pointed out (which can be resolved in a followup).

Make sure you squash it when you merge, please

rachel-fenichel (approved):

core/field_angle.js
```diff
@@ -235,12 +235,6 @@ Blockly.FieldAngle.prototype.setText = function(text) {
     return;
   }
   this.updateGraph_();
-  // Insert degree symbol.
```
picklesrus, line 238: One last question. Should this block of code stay? In trying to update 
https://github.com/LLK/scratch-blocks/issues/843 to explain why https://github.com/google/blockly/pull/1044 didn't fix it, I noticed we accepted part of that change (this delete), but not the addition of render_. This is fine as it is now if it works as well as before as long as we update 843.  What was the js error? textElement_ is undefined? Was it here or in render_?

core/workspace.js
paulkaplan: So my solution was to replace all `getFlyout` -> `getFlyout_`, is that right?

If we do that, we should just make sure that we update `scratch-gui` (https://github.com/LLK/scratch-gui/blob/075216f585770f74a189ddaff66ac75a39c2308d/src/containers/blocks.jsx#L63-L66)

rachel-fenichel: If you're using `getFlyout` in scratch-gui, you shouldn't rename it to `getFlyout_`, because that means it should be private.

rachel-fenichel:
:tada:

Issue #883, opened 2017-05-10
Categories not cleared on dispose


ericrosenbaum:
### Expected Behavior

The array containing block categories, Blockly.Toolbox.CategoryMenu.categories_, has a dispose function that should result in it becoming empty.

### Actual Behavior

The individual categories are disposed, but the array still contains them. Toolbox updates which are meant to replace categories end up adding new ones. The excess categories don't display, but an error is triggered on switching sprites in scratch-gui.

### Steps to Reproduce

I stumbled on this error as I was prototyping, using the updateToolbox function to replace the entire toolbox (in the future we should use the category menu callback to update the toolbox instead).

Thanks to @rachel-fenichel for help with this! There's a one-line fix.

### Operating System and Browser

Mac OS 10.11.6 Chrome 57.0.2987.133

PR #884, opened 2017-05-10
Empty the categories array on dispose

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/883

### Proposed Changes

Set categories_ to the empty array

### Reason for Changes

The categories were getting disposed individually, but the array elements were not being cleared, which caused an error.


rachel-fenichel:
Fine by me, but should go in after the big merge.

PR #886, opened 2017-05-11
Give blocks ids

fsih:
### Resolves
https://github.com/LLK/scratch-gui/issues/356

### Proposed Changes
Give blocks in the flyout fixed IDs so that when they are destroyed and re-created, they are treated as the same instance as the first block. This prevents monitor blocks from tracking tons of duplicates.

(Does this change need to be applied to default_toolbox in horizontal as well? core/variables, core/procedures, etc.?)
/cc @paulkaplan @rschamp

### DIFF
--- a/blocks_vertical/default_toolbox.js
+++ b/blocks_vertical/default_toolbox.js
@@ -30,41 +30,41 @@ goog.require('Blockly.Blocks');
 
 Blockly.Blocks.defaultToolbox = '<xml id="toolbox-categories" style="display: none">'+
   '<category name="Motion" colour="#4C97FF" secondaryColour="#3373CC">'+
-    '<block type="motion_movesteps">'+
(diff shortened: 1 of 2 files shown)


fsih:
Oops I accidentally Paul's change in here https://github.com/paulkaplan/scratch-blocks/tree/emit-checkbox-event

rschamp:
@fsih I think it's ok to pull in the checkbox change here, since this is everything required for the monitors update.

@rachel-fenichel FYI this PR contains the change event we talked about earlier this week.  It's convenient because it allows us to treat the checkbox state as part of the block state and handle it with our change event handler (in the VM we also treat it as block state).

rschamp (approved):
I think we can address the other blocks as we encounter them. Soon we'll be moving the toolbox definition out of this repository anyway.

I think this is also fine to merge using the change event. We can address that too as we encounter issues with it.

PR #891, opened 2017-05-17
New dragging!  Merge from google

rachel-fenichel:
This change picks up all of my new dragging code, plus some of the new variable code.

It's ginormous but I've checked it against https://github.com/google/blockly/pull/1078 and I'm pretty happy with it.

All of the insertion marker code now lives in insertion_marker_manager.js, which has the same API as dragged_connection_manager.js.  This should help reduce conflicts between scratch and blockly codebases.


rachel-fenichel:
One conflict in flyout_vertical, related to checkboxes, which I will check out tomorrow.

fsih:
Fixes https://github.com/LLK/scratch-blocks/issues/893

paulkaplan:
@rachel-fenichel thanks for walking me through this, sorry about the delay. 

Pulling and building to use for `scratch-gui` I get this error:
image

I'm not familiar with the way the `goog.require(...)` system works, but I tried adding `goog.require('Blockly.VariableMap')` to the top of `workspace.js`, then rebuilding and using in the GUI works. Not sure if that is the correct solution?

core/workspace.js
```diff
@@ -197,6 +201,7 @@ Blockly.Workspace.prototype.clear = function() {
   if (!existingGroup) {
     Blockly.Events.setGroup(false);
   }
+  this.variableMap_.clear();
   this.variableList.length = 0;
```
paulkaplan: Looks like this `this.variableList.length = 0` was missed, I think it should be removed (line 205 of new code).

paulkaplan:
@rachel-fenichel It looks like deleting a block by dragging it to the toolbox doesn't work. I don't get any cursor style and when i drop it does not delete. 
delete-doesnt-work

Traced it back to `Blockly.InsertionMarkerManager.prototype.update` (https://github.com/rachel-fenichel/scratch-blocks/blob/81c53c3fb53f6582833389c8063bf48e9608df0c/core/insertion_marker_manager.js#L214). Because setting `this.wouldDeleteBlock_` is inside the `shouldUpdate` if block, it is not being set. Seems like maybe these lines should go outside of `shouldUpdate` or something https://github.com/rachel-fenichel/scratch-blocks/blob/81c53c3fb53f6582833389c8063bf48e9608df0c/core/insertion_marker_manager.js#L221-L227

paulkaplan (changes requested):
Left feedback as other comments. Looks like a couple minor things and one (possibly) bigger problem.

rachel-fenichel:
Fixed all three.  Thanks for the details on the deletion one.  They made it easy to fix.

rachel-fenichel:
Oh wait--now if I have an insertion marker and transition to needing a deletion hand, the insertion marker remains.  That's no good.

rachel-fenichel:
Okay, I've fixed that one as well.

paulkaplan:
Great thanks @rachel-fenichel! Looks good to me now.

paulkaplan:
Wait sorry getting a similar require error for `Uncaught TypeError: Blockly.VariableModel is not a constructor` from `VariableMap.createVariable`.

rachel-fenichel:
Fixed missing require.

rachel-fenichel:
Verbally okayed by Paul.

PR #896, opened 2017-05-19
Add ID to each XY movement shadow ~field~ block for updating

paulkaplan:
### Reason for Changes

Add explicit ID field so that they can be updated from the GUI, re:  this issue https://github.com/LLK/scratch-gui/issues/210


rachel-fenichel:
LGTM, but note that they are shadow *blocks*, not shadow *fields*.

PR #898, opened 2017-05-22
Remove duplicate code in workspace_svg

rachel-fenichel:
Gets rid of more duplicated code from bad merges, and explicitly defines lastSound_ (which disappeared during a merge a year ago).


paulkaplan:
LG

PR #902, opened 2017-05-24
Fix disappearing dropdown

### LINKED ISSUE #901: Block menu opens and immediately closes on click
### Expected Behavior

Clicking on the menu in a block in the workspace should cause it to open (and stay open).

### Actual Behavior

The menu immediately closes again.

Note that this occurs only in the workspace, not in the toolbox.

menu-bug

Weirdly, clicking for a longer time (maybe half a second) causes the expected behavior. 

### Steps to Reproduce

Drag a block with a menu onto the workspace, such as "point towards [mouse pointer]".

Click on its menu to open it.

Mac OS 10.11.6 Chrome 58.0.3029.110

rachel-fenichel:
### Resolves

Fixes #901 

### Proposed Changes

Gets rid of a spurious dropdownDiv.hide();

### Reason for Changes

That's being done in gesture by calling hideChaff

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1016,9 +1016,6 @@ Blockly.WorkspaceSvg.prototype.isDeleteArea = function(e) {
  * @private
  */
 Blockly.WorkspaceSvg.prototype.onMouseDown_ = function(e) {
-  // TODO (fenichel): Move this to gesture.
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
Looks good- thanks!

PR #908, opened 2017-05-30
Add callback for setting checkbox state

paulkaplan:
Introduces a callback that can be overridden to provide the checkbox state when creating checkboxes.


core/blocks.js
```diff
@@ -35,3 +35,12 @@ goog.provide('Blockly.Blocks');
  * @type {!Object<string,Object>}
  */
 Blockly.Blocks = new Object(null);
+
+/**
+ * Get the default checkbox state of a given block. To be overridden.
```
rachel-fenichel: I'm really leery of having the checkbox state be on the block instead of on the flyout.  For instance, what if you have the same block in multiple categories in the toolbox, with different IDs?  (We can't do much if they have the same ID.)  The state of the checkbox in the flyout isn't intrinsic to the block prototype, and not every block instance is in a flyout--not even every BlockSvg instance.

If this is just a hack for the sake of testing, please add a TODO with a linked github issue to every place where you're hacking for the sake of adding this feature, so we can find all of the places to unhack later.

paulkaplan:
@rachel-fenichel I moved it into vertical flyout. Is that what you meant?

core/flyout_vertical.js
```diff
@@ -717,3 +723,12 @@ Blockly.VerticalFlyout.prototype.reflowInternal_ = function(/* blocks */) {
   // This is a no-op because the flyout is a fixed size.
   return;
 };
+
+/**
+ * Gets the checkbox state for a block
+ * @param {string} blockId The ID of the block in question.
+ * @return {boolean} Whether the block is checked.
```
rachel-fenichel, line 730: If public, add @public.  If not, @package.

rachel-fenichel:
@paulkaplan yes, that looks better.  I had one suggestion on annotations, but LGTM.

PR #916, opened 2017-06-02
Give persistent block ids to variables in the flyout

rschamp:
### Resolves

Persists the monitoring state of variables between flyout refreshes

### Proposed Changes

Give a persistent id to each variable block in the flyout

### Reason for Changes

Otherwise, the variable blocks get a new id each time the category is switched to/from, and the monitor checkboxes become unchecked (since they are keyed by flyout block id).

### DIFF
--- a/core/variables.js
+++ b/core/variables.js
@@ -132,6 +132,7 @@ Blockly.Variables.flyoutCategory = function(workspace) {
       var block = goog.dom.createDom('block');
       block.setAttribute('type', 'data_variable');
       block.setAttribute('gap', 8);
+      block.setAttribute('id', 'VAR_' + variableModelList[i].name);
 
       var field = goog.dom.createDom('field', null, variableModelList[i].name);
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Late to the party, but: why are you using the variable name instead of the variable ID?

paulkaplan:
We had to because there is no concept of a variable ID in the VM, so if we used ID it wouldn't be stable across sprite changes

rachel-fenichel:
No concept of a variable ID in the VM--so do you just have explicit lists for each variable type?  Or do lists and variables get prefixes?  How do you distinguish between a list and a variable that are both named foo?

rachel-fenichel:
Correction: you can't name a list and a variable the same name, but you can name a variable and a message the same name.

marisaleung:
Going to chip in here - I plan to change the VM so that it will use a variableMap and will have the concept of types and variables id's.
So I think after that we can use ID.

PR #917, opened 2017-06-02
Set checkbox state after creating a variable through the prompt

### LINKED ISSUE #911: Variables should be monitored by default when created
When a variable is created, it should be monitored by default (the little blue checkbox).

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/911

### Proposed Changes

Set the checkbox state after a variable has been created through the prompt flow. This will both set the checkbox correctly and emit an event so the VM can create a monitor.

### Reason for Changes

https://github.com/LLK/scratch-blocks/issues/911

variable-checked

---

This is a pretty temporary hack to make this work for our playtests. Doesn't seem like the right way going forward, but we will need to rethink a fair amount about how monitors work to get this to work correctly.

### DIFF
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -546,13 +546,12 @@ Blockly.VerticalFlyout.prototype.createCheckbox_ = function(block, cursorX,
  * @private
  */
 Blockly.VerticalFlyout.prototype.checkboxClicked_ = function(checkboxObj) {
-  var context = this;
   return function(e) {
-    context.setCheckboxState(checkboxObj.block.id, !checkboxObj.clicked);
+    this.setCheckboxState(checkboxObj.block.id, !checkboxObj.clicked);
     // This event has been handled.  No need to bubble up to the document.
     e.stopPropagation();
     e.preventDefault();
-  };
+  }.bind(this);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


core/variables.js
```diff
@@ -381,7 +381,14 @@ Blockly.Variables.createVariable = function(workspace, opt_callback) {
                 });
           }
           else {
-            workspace.createVariable(text);
+            var variable = workspace.createVariable(text);
+
+            var fl = workspace.getFlyout();
```
rschamp: Please rename to `flyout`.

core/flyout_vertical.js
```diff
@@ -546,27 +546,45 @@ Blockly.VerticalFlyout.prototype.createCheckbox_ = function(block, cursorX,
...
-    if (checkboxObj.clicked) {
-      Blockly.utils.addClass((checkboxObj.svgRoot), 'checked');
-    } else {
-      Blockly.utils.removeClass((checkboxObj.svgRoot), 'checked');
-    }
-
-    Blockly.Events.fire(new Blockly.Events.Change(
-        checkboxObj.block, 'checkbox', null, oldValue, newValue));
-
+    context.setCheckboxState(checkboxObj.block.id, !checkboxObj.clicked);
```
rschamp: Please use the poor man's arrow function here, and `.bind(this)` on the function definition.

rschamp (changes requested):
Looks good other than a couple of nits.

rschamp (approved):

PR #924, opened 2017-06-05
Don't attempt to give the flyout negative height

### LINKED ISSUE #782: When page is short enough that the flyout doesn't fit, a js error is thrown
There's a js error thrown when the the page is small enough that the flyout doesn't fit.

Error: <rect> attribute height: A negative value is not valid. ("-3px")
Stack trace:
Blockly.VerticalFlyout.setMetrics_	@	vertical.js:2957
Blockly.Scrollbar.onScroll_	@	vertical.js:2513
Blockly.Scrollbar.resize	@	vertical.js:2501
Blockly.VerticalFlyout.position	@	vertical.js:2959
Blockly.Toolbox.position	@	vertical.js:2977
Blockly.WorkspaceSvg.resize	@	vertical.js:2565
Blockly.svgResize	@	vertical.js:3048
(anonymous)	@	vertical.js:3043

We should pin the minimum height attribute for the flyout at 0.
It is probably worth checking the width as well.

### Steps to Reproduce

Open the js console and shrink the page vertically. As the flyout disappears due to lack of space, watch the console to see the js error.

Chome on mac0S.

cesium12:
### Resolves

Fixes #782

### Proposed Changes

See bug report. The flyout subtracts some measurements to decide how much space it has. If the window is too small, it tries setting the element to <0px tall. In this case it should be safe to just set it to 0.

### DIFF
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -237,7 +237,7 @@ Blockly.VerticalFlyout.prototype.setMetrics_ = function(xyRatio) {
   this.workspace_.translate(this.workspace_.scrollX + metrics.absoluteLeft,
       this.workspace_.scrollY + metrics.absoluteTop);
 
-  this.clipRect_.setAttribute('height', metrics.viewHeight + 'px');
+  this.clipRect_.setAttribute('height', Math.max(0, metrics.viewHeight) + 'px');
   this.clipRect_.setAttribute('width', metrics.viewWidth + 'px');
 };
 
@@ -269,7 +269,7 @@ Blockly.VerticalFlyout.prototype.position = function() {
   }
 
   // Record the height for Blockly.Flyout.getMetrics_
-  this.height_ = targetWorkspaceMetrics.viewHeight - y;
+  this.height_ = Math.max(0, targetWorkspaceMetrics.viewHeight - y);
 
   this.setBackgroundPath_(this.width_, this.height_);
 


paulkaplan:
@rachel-fenichel do you think this is a good way to do this? Doesn't seem to cause any issues right now, and might make this harder to fix down-the-line because it will basically swallow the error?

fsih:
Issue #782 says
"We should pin the minimum height attribute for the flyout at 0.
It is probably worth checking the width as well."
so seems like this is the way Katelyn recommended

picklesrus:
@fsih @paulkaplan Yeah. I do think pinning the minimum height at 0 (or maybe you'd want a larger min so something is visible.) is right.  But, I think @paulkaplan is saying viewHeight should be 0 too, right?  We should understand why the viewHeight is coming out negative.  If it is just because this.height is 0 [here](https://github.com/google/blockly/blob/master/core/flyout_vertical.js#L94) and we subtract 2* scrollbarPadding, it seems fine to just set viewHeight to 0 in getMetrics.  If it is because this.height on that line is negative, we should probably track that down.

thisandagain:
@paulkaplan Bump

paulkaplan:
@picklesrus it appears that the viewHeight is negative before subtracting the padding.

paulkaplan:
I'm going to merge this since it doesn't seem to cause any new issues.

PR #925, opened 2017-06-06
Optimize bulk changes to the workspace in more situations

### LINKED ISSUE #887: Performance: Creating blocks that refer to variables
### Expected Behavior

DomToBlock on a block with a variable reference (FieldVariable or FieldVariableGetter) should not take much longer than creating a block with a FieldDropdown or FieldTextInput.

### Actual Behavior

Approximate benchmarking, by switching sprites in a project with lots of variables.  Here's a sample FieldTextInput.init call, taking 0.23ms:
image

And here is a sample FieldVariable.init call.  It takes 213ms, with 212ms of that spent refreshing the toolbox unnecessarily.
image

Slowdown mechanism:
WorkspaceSvg.createVariable is called for every field that references a variable, which causes the flyout to be refreshed.  

In Blockly, projects with variables usually have an auto-closing flyout, so it doesn't refresh.  In scratch-blocks the flyout is almost always open.  In Blockly createVariable 

Relevant code: https://github.com/LLK/scratch-blocks/blob/develop/core/workspace_svg.js#L1004
```
Blockly.WorkspaceSvg.prototype.createVariable = function(name) {
  Blockly.WorkspaceSvg.superClass_.createVariable.call(this, name);
  // Don't refresh the toolbox if there's a drag in progress.
  if (this.toolbox_ && this.toolbox_.flyout_ && !Blockly.Flyout.startFlyout_) {
    this.toolbox_.refreshSelection();
  }
};
```

refreshSelection on the toolbox can take some time due to layout passes (tracked in #879).  But the problem here is that we're calling refreshSelection too many times.

### Steps to Reproduce

Load a workspace with plenty of variable references (see https://github.com/LLK/scratch-gui/issues/329).

All

### Possible Solutions

- Only refresh the toolbox if it's open to a category with variables (data, extensions, ?).  There are some pretty easy ways to backfire here--it would make changing sprites change speed depending on which category you have open.
- Block toolbox refreshing while loading new workspaces--this could be similar to how we block workspace resizing during a load
- Batch variable creation at the beginning of a workspace load.  This should be happening with the new typed variable work.

sclements:
Resolves #887

This introduces an isUpdating_ flag to the WorkspaceSvg class and removes the resizesEnabled_ flag. This allows for optimizing more situations where the WorkspaceSvg is changing it's contents and shouldn't perform some functions it normally would during changes. I used the Paper Minecraft (10128407) project for testing since I know it has a massive number of blocks.

Pre-changes it took 48 seconds to switch between the first two sprites:
image

Post-changes it took 6.8 seconds:
image


sclements:
My git-fu is stale. I'll try to resolve this later today.

sclements:
Ready for review

core/workspace_svg.js
```diff
@@ -1686,28 +1688,34 @@ Blockly.WorkspaceSvg.setTopLevelWorkspaceMetrics_ = function(xyRatio) {
...
+ * Use to avoid resizing and other logic during a batch operation, for
+ * performance.
+ * @param {boolean} isUpdating Whether the workspace is updating.
  */
-Blockly.WorkspaceSvg.prototype.setResizesEnabled = function(enabled) {
-  var reenabled = (!this.resizesEnabled_ && enabled);
-  this.resizesEnabled_ = enabled;
-  if (reenabled) {
-    // Newly enabled.  Trigger a resize.
+Blockly.WorkspaceSvg.prototype.setIsUpdating = function(isUpdating) {
```
picklesrus, line 1698: This method is technically public so we need to be careful about changing the name so we don't break external developers.  Stuffing that logic inside of setResizesEnabled then makes setResizesEnabled misleading though so a name change does seem appropriate.  What about making another function that, for now, calls setResizesEnabled and does the refreshSelection check?  We could tag setResizesEnabled deprecated and then remove it eventually.  I'm stuck on a name for it though. pauseBulkUpdates? start/endBulkUpdate?

picklesrus (commented):
Thanks for this! Sorry for the slow response!

core/workspace_svg.js
sclements: I'm good with start/endBulkUpdate. I'll get back to this another day when I have some time.

paulkaplan:
@sclements, @picklesrus any update on this?

thisandagain:
@sclements @picklesrus @paulkaplan Hi folks. As per our discussion yesterday, should we close this for now and attempt a new PR when we have some time to push it from our side?

rachel-fenichel:
Do we have a timeline for getting this to work?  It's a giant performance improvement once we land it.

picklesrus:
We're going to look more generally at toolbox performance, especially in the context #1059, later this week. Plus, #887 is still open so I don't think this will be forgotten :)

Issue #907, opened 2017-05-30
Hit area of dropdown button under horizontal blocks is smaller than before


msintov:
### Expected Behavior

In horizontal_playground.html, when the user taps any part of the dropdown button under the envelope blocks, the dropdown should display.

### Actual Behavior

If the user taps on the white down arrow in the middle of the dropdown button under the envelope blocks, the dropdown displays. If the user taps anywhere else on the button, the dropdown does not display.

This new behavior appears to have been introduced with 6275e11, the new dragging commit. Is the smaller hit area intentional?

### Steps to Reproduce

1. Load horizontal_playground.html.
2. Tap the dropdown button under either of the envelope blocks without tapping on the white down arrow. The blocks can be either in the flyout or in the workspace.

### Operating System and Browser

_e.g. Mac OS 10.11.6 Chrome Version 58.0.3029.110

rachel-fenichel:
Nope, that's not intentional.  Thanks for noting which commit it came in on.  I'll take a look.

msintov:
This issue also applies to other dropdowns such as those containing a number with numpad dropdown.

rachel-fenichel:
Fixed by #927

PR #927, opened 2017-06-06
Fix click targets on single-field blocks

### LINKED ISSUE #920: Field selection targets are too small
### Expected Behavior

- Clicking anywhere on a selection field should trigger the drop-down

### Actual Behavior

- Only clicking in the very center of the field will produce the drop-down

### Steps to Reproduce

untitled

MacOS 10.12.4 | Chrome Version 58.0.3029.110 (64-bit)

/cc @paulkaplan @rschamp

rachel-fenichel:
### Resolves

Fixes #920 
### Proposed Changes

Reverts the change to this line: https://github.com/LLK/scratch-blocks/commit/6275e1137c9be43ff8a94dcf8139196cb384e657#diff-393b262d7ed6406aa328f2e57e089e5fR182

### Reason for Changes

Blocks with single fields should have the entire field as the click target.  That behaviour was explicitly added in [this commit](https://github.com/LLK/scratch-blocks/commit/4daf3cd16fbe11d52f18c590e4b5b2972ef6534a) and accidentally removed in the new dragging code.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -179,7 +179,7 @@ Blockly.Field.prototype.init = function() {
   this.render_();
   this.size_.width = 0;
   this.mouseDownWrapper_ =
-      Blockly.bindEventWithChecks_(this.fieldGroup_, 'mousedown', this,
+      Blockly.bindEventWithChecks_(this.getClickTarget_(), 'mousedown', this,
       this.onMouseDown_);
 };
 


rachel-fenichel:
Fixes #907

paulkaplan (approved):
Great! @thisandagain I checked this with the color fields, the dropdowns and number inputs and seems to fix all of them. 

One question @rachel-fenichel, and this doesn't block this PR, but would it be hard to change the hover cursor behavior so that it was consistent over the clickable area? I don't know if this is a big deal, but I noticed that even though the whole area is clickable, the cursor still has the drag hand and changes over the smaller inner field.

picklesrus:
@paulkaplan: about the cursor part, it should be easier now that the cursor is set (mostly) in css instead of js.  In the case of the color input, the reason the cursor stays a hand for so long is because the element with class "blocklyEditableText" (which has cursor: default to override "blocklyDraggable" cursor:grab) is quite small.  I'm not sure why those elements are different sized in scratch-blocks and blockly or if it matters. Either making the blocklyEditableText take up the whole field like it does in blockly, or adding the cursor:default to whatever element corresponds to getClickTarget() should do the trick.

rachel-fenichel:
@picklesrus in Blockly the whole colored area is the field.  In scratch-blocks the colored area is the block, and the field's group is just an empty bit of text.  The color field changes the color of its parent block on set.

Which raises the question: what does scratch-blocks do if you try to have a color field on a block with more than one field?  If that's just not allowed, it's not recorded anywhere.

@paulkaplan can you file a separate issue for the cursor?

PR #928, opened 2017-06-07
Fix #922

### LINKED ISSUE #922: Dragging stacks of blocks over other stacks disconnects them
### Expected Behavior

Dragging a stack of blocks around the screen has no lasting impact on other blocks / scripts that are not part of the drag source or target.

### Actual Behavior

As a stack of two or more blocks is dragged over other block stacks, the other stacks become disconnected (although they stay positioned as if they were still connected).

### Steps to Reproduce

Here's a screen capture of me showing the behavior of dragging two blocks over a stack in a chrome browser. Notice how 'all' blocks are detached once the drag is complete even though I didn't release the drag over the stack? This is happening on https://llk.github.io/scratch-gui/ has this page got the latest changes from github?

6cb209ba-4792-11e7-8535-7db61f9e0f1c

Windows 10 with Chrome

rachel-fenichel:
### Resolves

Fixes #922 

### Proposed Changes
Explicitly check that the candidate connection is the first in the stack besides the insertion marker, rather than relying on the presence of the insertion marker to say that this is a good connection.
I'm not convinced that this is the correct solution, since there's still a ton of code for this check.  I think it is a functional solution.

### Reason for Changes
image

The bug was happening when dragging a stack of two or more statement blocks (in this case, the `show` and `clear` blocks) over a stack of two or more statement blocks (in this case, the `stop all sounds` and `if on edge, bounce` blocks).  Scratch allows the last statement block in a dragging stack to connect in front of the first statement block in a stationary stack, and I broke the logic for that check.

The previous connection of the `show` block would try to connect to the next connection of the `stop all sounds` block.  This was legal, so an insertion marker would appear between the `stop all sounds` block and the `if on edge, bounce` block.

On the next mouse move, the code would check whether the last connection in the dragging stack (the next connection on the `clear` block) could connect to anything nearby.  The nearest available connection would be the previous connection of the `if on edge, bounce` block.  There was already an insertion marker there, so the old code said that the connection must be valid.  But the insertion marker was there based on the `show` block, not the `clear` block, and therefore had nothing to do with the validity of the connection.  It wasn't enough to check whether the connected block is an insertion marker--I had to also check whether that insertion marker was the first block in the stack.

Here's the old code:
```
if (candidate.targetConnection) {
          // If the other side of this connection is the active insertion marker
          // connection, we've obviously already decided that this is a good
          // connection.
          if (candidate.targetBlock().isInsertionMarker()) {
            return true;
          } else {
            return false;
          }
        }
```

Before my new dragging it checked for the insertion marker connection, rather than the insertion marker block:
```
if (candidate.targetConnection) {
          // If the other side of this connection is the active insertion marker
          // connection, we've obviously already decided that this is a good
          // connection.
          if (candidate.targetConnection ==
              Blockly.insertionMarkerConnection_) {
            return true;
          } else {
            return false;
          }
        }
```

My new code hid away the details of the insertion marker, but in this case I hid them too well.

### DIFF
--- a/core/connection.js
+++ b/core/connection.js
@@ -345,6 +345,60 @@ Blockly.Connection.prototype.checkConnection_ = function(target) {
  * Check if the two connections can be dragged to connect to each other.
  * This is used by the connection database when searching for the closest
  * connection.
+ * @param {!Blockly.Connection} candidate A nearby connection to check, which
+ *     must be a previous connection.
+ * @return {boolean} True if the connection is allowed, false otherwise.
+ */
+Blockly.Connection.prototype.canConnectToPrevious_ = function(candidate) {
+  if (this.targetConnection) {
+    // This connection is already occupied.
+    // A next connection will never disconnect itself mid-drag.
+    return false;
+  }
+
+  // Don't let blocks try to connect to themselves or ones they nest.
+  if (Blockly.draggingConnections_.indexOf(candidate) != -1) {
+    return false;
+  }
+
+  var firstStatementConnection =
+      this.sourceBlock_.getFirstStatementConnection();
+  // Is it a C-shaped (e.g. repeat) or E-shaped (e.g. if-else) block?
+  var isComplexStatement = firstStatementConnection != null;
+  var isFirstStatementConnection = this == firstStatementConnection;
+  var isNextConnection = this == this.sourceBlock_.nextConnection;
+
+  // Scratch-specific behaviour: can connect to the first statement input of a
+  // C-shaped or E-shaped block, or to the next connection of any statement
+  // block, but not to the second statement input of an E-shaped block.
+  if (isComplexStatement && !isFirstStatementConnection && !isNextConnection) {
+    return false;
+  }
+
+  if (isFirstStatementConnection) {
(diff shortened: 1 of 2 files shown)


core/connection.js
```diff
@@ -345,6 +345,60 @@ Blockly.Connection.prototype.checkConnection_ = function(target) {
...
+    }
+    // If it's connected to an insertion marker but that insertion marker
+    // is the first block in a stack, it's still fine.  If that insertion
+    // marker is in the middle of a stack, it won't work.
+    return !targetBlock.getPreviousBlock();
+  }
+};
+
+/**
+ * Check if the two connections can be dragged to connect to each other.
```
paulkaplan, line 399: The jsdoc for 345-347 is the same as 399-401, which doesn't seem right?

paulkaplan (approved):
Code looks good other than that duplicate jsdoc question!

core/connection.js
rachel-fenichel: Yes, that's intentional.  The first function has an extra description of the candidate parameter, saying that it must be a previous connection.

Issue #929, opened 2017-06-07
Overlapping field text


fsih:
Branched from https://github.com/LLK/scratch-gui/issues/413

In the vertical playground:
- Add the "touching mouse-pointer" block.
- Click "Export to XML"
- Remove that block so it is empty.
- Click "Hide"
- Click "Import from XML"
- Click "Show"

### Expected Behavior
Text in the blocks should not overlap.

### Actual Behavior
Field text overlaps block

![image](https://user-images.githubusercontent.com/2855464/26888519-446fc0ce-4b79-11e7-8fc5-1d1993d44963.png)

It also isn't just dropdowns. Here I typed in a large number, and did the same steps as above

![image](https://user-images.githubusercontent.com/2855464/26888531-49c58eaa-4b79-11e7-9c1c-466be34314a6.png)

fsih:
Summarizing the comments from the old issue:

"The reason this is happening is because blockly fields are initialized with isVisible set to true, which means that when we create them while the workspace is not visible they try to render themselves and cache their width incorrectly. We need to cause a re-render of the workspace when you come back to [visible]. We might also want a way for the workspace when hidden to tell initialized fields that they are hidden, so we don't waste time re-rendering"

"It looks like setVisible should be causing a re-render. Not sure why it isn't recalculating the widths. May want to file this on scratch-blocks and reach out to them?"

Issue #912, opened 2017-05-31
Can't delete blocks


thisandagain:
### Expected Behavior

- Dragging a block into the palette will cause it to be deleted

### Actual Behavior

- Occasionally, dragging a block into the palette will result in it being placed behind the pallete

### Steps to Reproduce

- I haven't been able to find a reproducible case yet. Both @ericrosenbaum and I have seen it over the past few days

### Operating System and Browser

MacOS 10.12.4 | Chrome Version 58.0.3029.110 (64-bit)

/cc @ericrosenbaum @paulkaplan @rachel-fenichel

rachel-fenichel:
You are probably grabbing the block on the right side of the block and dragging it so that the origin of the block is over the flyout, but the cursor isn't.  The deletion state depends on the position of the cursor.

We can change that if you want, but it needs to be an explicit change.

If that's not the cause of the bug, let me know and I can debug further.

thisandagain:
Nope. I've certainly seen it when both the cursor and the entire block are over the palette.

rachel-fenichel:
:+1: I'll take a look.

towerofnix:
I've noticed this as well.

I think the most reproducible cause of this is to go into a separate editor tab (e.g. "costumes" or "sounds") and then go back into "scripts". Blocks will no longer be able to be deleted (even ones newly dragged out from the palette, that haven't even been dropped into the workspace – i.e. grab a new block from the palette, then release it while over the palette - it won't be deleted).

paulkaplan:
@liam4 Yup that looks like it works to reproduce. 

1. Load https://llk.github.io/scratch-gui/ fresh
2. Drag a block out onto the workspace.
3. Click "Costumes"
4. Click "Scripts" 
5. Try to drag the block back into the toolbox, but it cannot be deleted

---

Update 1: Investigating a bit further, here are a few things that **do not** cause this to happen:
 - changing from sprite to background (which causes a whole workspace update)
 - using the show/hide buttons in the playground (setVisible is one of the things that happens when changing tabs, but doesn't seem to be solely responsible)

So far I haven't been able to reproduce in the playground.

---
[...]

PR #930, opened 2017-06-07
Do not render blocks or fields if the workspace is hidden

fsih:
### Resolves

https://github.com/LLK/scratch-blocks/issues/929

### Proposed Changes

Check if workspace is visible when rendering fields

### DIFF
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -379,7 +379,7 @@ Blockly.BlockSvg.prototype.getHeightWidth = function(opt_ignoreFields) {
  */
 Blockly.BlockSvg.prototype.render = function(opt_bubble) {
   // No need to render if the workspace is hidden.
-  if (!this.workspace.isVisible) {
+  if (!this.workspace.isVisible()) {
     return;
   }
   
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -378,6 +378,11 @@ Blockly.BlockSvg.prototype.getHeightWidth = function(opt_ignoreFields) {
  *   If true, also render block's parent, grandparent, etc.  Defaults to true.
  */
 Blockly.BlockSvg.prototype.render = function(opt_bubble) {
+  // No need to render if the workspace is hidden.
+  if (!this.workspace.isVisible) {
+    return;
+  }
+  
   Blockly.Field.startCache();
   this.rendered = true;
 
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -568,6 +568,11 @@ Blockly.BlockSvg.prototype.getHeightWidth = function() {
  *   If true, also render block's parent, grandparent, etc.  Defaults to true.
  */
 Blockly.BlockSvg.prototype.render = function(opt_bubble) {
+  // No need to render if the workspace is hidden.
+  if (!this.workspace.isVisible) {
+    return;
+  }
+  
   Blockly.Field.startCache();
   this.rendered = true;
 
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -569,7 +569,7 @@ Blockly.BlockSvg.prototype.getHeightWidth = function() {
  */
 Blockly.BlockSvg.prototype.render = function(opt_bubble) {
   // No need to render if the workspace is hidden.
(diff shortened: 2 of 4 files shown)


rachel-fenichel:
From past experience, shuffling around the timing of rendering passes usually breaks Blockly for edge and IE10/11.

I think there's a different root cause, involving our show function sometimes being too lazy, and I'd rather solve that one.

fsih:
I'm not sure what you mean by show being too lazy. Isn't the issue that show is being called too often, and it's caching a width that doesn't exist yet because the workspace is hidden?

rachel-fenichel:
@picklesrus

picklesrus:
I don't think show is being called too often, but more than render is being called when the workspace is invisible.  We could try and catch all the places that call render and check for visibility, but I had been thinking of this as show's problem. Lots of other things may have changed (window size, browser zoom level, etc.) while the workspace was hidden. #912 is essentially that problem too - we do not recalculate the position of the delete area when the workspace is set back to be visible. @rachel-fenichel just sent out a fix that might actually fix this too.

All that said, re-rendering everything on show is a heavier hammer that we might want to avoid if it caused performance problems.  If the browser is smart for us and not spending a bunch of time rendering stuff while it is invisible, doing it when the workspace is shown again seems a bit easier?

picklesrus:
Hmmm, but everything I said before doesn't make a ton of sense now that I see workspace.render is already being called in show... Do you know at what level the render pass stops when show is called?  Looks like it should get called on each block - does it get called on each field? Is the text width somehow cached?  I was under the impression the cache was clear at that point, but maybe not?

fsih:
It's this line here:
https://github.com/LLK/scratch-blocks/blob/develop/core/field.js#L489

When render happens when the workspace is hidden, the field gets its width set to something that doesn't include text width. Then when the workspace is shown, the field already has a width, so getSize returns the old width.

fsih:
@rachel-fenichel @picklesrus do you guys have advice on a better way to fix this bug?

rachel-fenichel:
@fsih sorry, I haven't yet had time to fully understand this issue.  I'm hoping to get to it later today, or on monday at the latest.

picklesrus:
sorry for the delayed response @fsih The end of last week ended up 100% filled with meetings.

I think this fix is ok, but I worry that there are other ways the workspace will be invisible that we won't know about - I'm pretty sure we ran into one with the snowflake project we did.  Let me dig a little into what the width is on that line you mentioned - I'd hoped it would have been 0, but it must not be.

picklesrus:
Ah. One interesting bit - blockly actually sets with width back to 0 when fields are invisible:
https://github.com/google/blockly/blob/master/core/field.js#L307  It doesn't look like scratch-blocks does that in field's render.  I know that isn't much different from what you've added, would adding that bit of code to scratch-blocks solve the problem?  At least then it is in the same place?

fsih:
No problem, thanks for helping look at this bug!

Adding the 
```
  if (!this.visible_) {
    this.size_.width = 0;
    return;
  }
```
clause actually doesn't work, because field.visible_ is set to true. The field "visibility" has to do with whether the block is collapsed and not whether the workspace is visible. Should I switch it to check for workspace visibility there?

picklesrus:
Ah, I see. That's unfortunate.  I do think workspace visibility might be the right longer term way to do this since I'd rather refine the signal for when we need to clear the text size cache than avoid rendering (which is sprinkled all over) whenever we think there might be a problem.  I'm not sure using the workspace visibility won't cause you more bugs right now though - this code is brittle :(
I know you have some play testing coming up though so don't hold off your merge and let me block you from getting the things back into a working state!

fsih:
Just FYI, Ray suggests we circumvent this bug from the GUI side by clearing out the workspace instead of just hiding it when the tab isn't visible, so I'm looking at that alternative.

fsih:
Just merged https://github.com/LLK/scratch-gui/pull/459 where the GUI reloads the workspace when it is set to visible. So we can feel free to pursue a longer-term better solution here.

paulkaplan:
So since this isn't blocking us anymore, @rachel-fenichel @picklesrus how do you want to handle this? Should we close the PR and reopen at a later date?

thisandagain:
@paulkaplan @fsih @rachel-fenichel @picklesrus How would you folks like to resolve this? Should we close for now?

paulkaplan:
Let's close this for now.

Issue #926, opened 2017-06-06
Duplicating blocks with shadows keep the same shadow id


paulkaplan:
When you create two of the same block by dragging them out of the toolbox twice, you get different IDs for all the blocks (including the shadow)

![different-ids](https://user-images.githubusercontent.com/654102/26850635-40f1c4a4-4ad6-11e7-8ffa-cc148bf0be45.png)

However, when you right-click and duplicate, you get a block with a different ID but the shadow has the same ID.

![same-id](https://user-images.githubusercontent.com/654102/26850664-538e6f7c-4ad6-11e7-9b45-51a912b801fc.png)

This is causing a bug in scratch-vm https://github.com/LLK/scratch-vm/issues/589 because the vm-side block cache can be fooled by blocks with the same IDs (specifically, if you delete one of them and then try to read out the block XML, you get an error). 

@rachel-fenichel is this a known / desired behavior?

RoboErikG:
That only happens when you duplicate a block where the shadow is hidden by another block. This is because the shadow is serialized to xml and not present in the model when it's replaced by another block, and duplicating it doesn't touch the serialized shadow xml.

When you remove the x position blocks the shadows should be loaded from xml back into the model and the second one will get a new id.

paulkaplan:
Ok interesting. I can't seem to find where in the code that is happening. What would be the easiest way to fix this problem?

For more context, you can see https://github.com/LLK/scratch-vm/issues/589, where the TL;DR is "it isn't safe to delete blocks if there are duplicate IDs."

paulkaplan:
It looks like one fix would be to use the `opt_noId` option in `blockToDom` while copying the XML.

RoboErikG:
I think your VM is missing a step when a shadow block is replaced by a normal block. In the [connection code](https://github.com/google/blockly/blob/06ebf52dd11da90e7407082ed241558b2a91fd86/core/connection.js#L136) the shadow gets serialized and then the model object is disposed. I'm guessing your VM doesn't dispose of the shadow object and this leads to having two copies of it when you try to delete.

The options are to add a check in the VM to remove the shadow block when another block is attached or we can start sending events for the shadow block being deleted/created. We hadn't added events for it because we originally thought only other Blockly instances would listen to events and would always perform the delete/create of the shadow block in response to the move.

paulkaplan:
My knowledge of this area of the VM is still very limited, so /cc @thisandagain and @rschamp who might have more context?

[...]

PR #936, opened 2017-06-09
Do not duplicate IDs on paste

paulkaplan:
### Resolves

In response to https://github.com/LLK/scratch-blocks/issues/926 and fixing https://github.com/LLK/scratch-vm/issues/589 and https://github.com/LLK/scratch-vm/issues/590

@rschamp and I spent a while trying to figure this out. Here is what is going on, and why we chose this (temporary) solution.

#### How scratch-blocks treats shadows
Blockly "deletes" shadows when they are obscured, sending a delete event and disposing of the model. However, the shadow is serialized for possible later use. If/when the obscuring block is removed, the shadow is deserialized, given a new ID if needed, and reinstated in the model. 

#### How scratch-vm treats shadows
Scratch-vm treats shadows as regular blocks. They are stored (by id) in a target-specific block cache, where they can be acted on by create, move and delete events. The block input stores a reference to both a `shadow` id and a `block` id, both to be retrieved from the block cache. If there is both a `shadow` and a `block` on the input, and they are different, it is an obscured shadow. When blocks are deleted, the blocks connected to their inputs (both "real" blocks and shadow blocks) are deleted. 

#### Where the trouble comes from
These two models _would_ be functionally equivalent if not for one thing: because Blockly does not consider an obscured shadow block a "real" block, it does not give it a new ID when it is duplicated (that happens when the shadow is eventually revealed). See  https://github.com/LLK/scratch-blocks/issues/926 for further description.

---

@rschamp and I tried several things to try to make the VMs model of processing events, but were not able to without a full refactor of the way shadow blocks are dealt with. I'd like to have a more full conversation with the team offline about when/if that will be required (I have some reservations about making the vm work the way scratch-blocks does). 

---

### Proposed Changes

This change rewrites the shadow IDs on paste by reaching into each block connection and setting the id attribute using `genUid`. (The rearrangement of the IE/Edge fix is just coincidental, saving a separate loop through the blocks on paste). 

I marked the change `Scratch-specific` to aid in future merging.

We tried to confirmed this fixes all of the bugs relating to blocks disappearing from the VM.

---

@rachel-fenichel and @RoboErikG thanks for talking to us about this offline and doing some investigation. I know this isn't the answer that you noted in https://github.com/LLK/scratch-blocks/issues/926. If this seems like a kosher thing to do, I think it would be best to merge this and than do a deeper dive a bit later on into bringing the vm more in line with blockly.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -901,14 +901,30 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
   Blockly.Events.disable();
   try {
     var block = Blockly.Xml.domToBlock(xmlBlock, this);
-    // Rerender to get around problem with IE and Edge not measuring text
-    // correctly when it is hidden.
-    if (goog.userAgent.IE || goog.userAgent.EDGE) {
-      var blocks = block.getDescendants();
-      for (var i = blocks.length - 1; i >= 0; i--) {
-        blocks[i].render(false);
+
+    var blocks = block.getDescendants();
+    for (var i = blocks.length - 1; i >= 0; i--) {
+      var descendant = blocks[i];
+
+      // Scratch-specific: Give shadow dom new IDs to prevent duplicating on paste
+      for(var j = 0; j < descendant.inputList.length; j++) {
+        if (descendant.inputList[j].connection) {
+          var connection = descendant.inputList[j].connection;
+          if (connection && connection.getShadowDom()) {
+            var shadowDom = connection.getShadowDom();
+            shadowDom.setAttribute('id', Blockly.utils.genUid());
+            connection.setShadowDom(shadowDom);
+          }
+        }
+      }
+
+      // Rerender to get around problem with IE and Edge not measuring text
+      // correctly when it is hidden.
+      if (goog.userAgent.IE || goog.userAgent.EDGE) {
+        descendant.render(false);
       }
     }
+
     // Move the duplicate to original position.
     var blockX = parseInt(xmlBlock.getAttribute('x'), 10);
     var blockY = parseInt(xmlBlock.getAttribute('y'), 10);
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -910,8 +910,8 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
       for(var j = 0; j < descendant.inputList.length; j++) {
         if (descendant.inputList[j].connection) {
           var connection = descendant.inputList[j].connection;
-          if (connection && connection.getShadowDom()) {
-            var shadowDom = connection.getShadowDom();
+          var shadowDom = connection.getShadowDom();
+          if (shadowDom) {
             shadowDom.setAttribute('id', Blockly.utils.genUid());
             connection.setShadowDom(shadowDom);
           }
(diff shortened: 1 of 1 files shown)


core/workspace_svg.js
```diff
@@ -901,14 +901,30 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
...
+    var blocks = block.getDescendants();
+    for (var i = blocks.length - 1; i >= 0; i--) {
+      var descendant = blocks[i];
+
+      // Scratch-specific: Give shadow dom new IDs to prevent duplicating on paste
+      for(var j = 0; j < descendant.inputList.length; j++) {
+        if (descendant.inputList[j].connection) {
+          var connection = descendant.inputList[j].connection;
+          if (connection && connection.getShadowDom()) {
+            var shadowDom = connection.getShadowDom();
```
rschamp: Could be simplified a little bit since we already know if the connection exists
```javascript
        if (descendant.inputList[j].connection) {
          var connection = descendant.inputList[j].connection;
          var shadowDom = connection.getShadowDom();
          if (shadowDom) { ...
```

paulkaplan: Fixed in  af103a2

rachel-fenichel:
Looking at this code, I see that Blockly is only emitting one Create event on a paste, regardless of how many blocks were pasted.  Does the VM code walk the full XML to find all of the children?

core/workspace_svg.js
```diff
@@ -901,14 +901,30 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
...
-        blocks[i].render(false);
+
+    var blocks = block.getDescendants();
+    for (var i = blocks.length - 1; i >= 0; i--) {
+      var descendant = blocks[i];
+
+      // Scratch-specific: Give shadow dom new IDs to prevent duplicating on paste
+      for(var j = 0; j < descendant.inputList.length; j++) {
+        if (descendant.inputList[j].connection) {
+          var connection = descendant.inputList[j].connection;
```
rachel-fenichel: Move this before the if, then do `if (connection)`

core/workspace_svg.js
```diff
@@ -901,14 +901,30 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
...
-      var blocks = block.getDescendants();
-      for (var i = blocks.length - 1; i >= 0; i--) {
-        blocks[i].render(false);
+
+    var blocks = block.getDescendants();
+    for (var i = blocks.length - 1; i >= 0; i--) {
+      var descendant = blocks[i];
+
+      // Scratch-specific: Give shadow dom new IDs to prevent duplicating on paste
+      for(var j = 0; j < descendant.inputList.length; j++) {
```
rachel-fenichel: nit: missing space between `for` and `(`

rachel-fenichel:
A few style comments, but I think this works.

RoboErikG:
This looks reasonable to me too.

rachel-fenichel:
Hold on, I have a few things to check before you merge this.

rachel-fenichel:
This breaks if you have nested shadow blocks.  

This change is safe to make for the sake of your play test because scratch-blocks doesn't have nested shadow blocks in the current base blocks.

Longer term, either this needs to be fixed correctly or you need to state that nested shadow blocks are illegal.  Defining whether statements can be shadows is probably part of that work.

paulkaplan:
Thanks @rachel-fenichel and @RoboErikG. Sounds good about nested shadows, will keep that in mind.

rschamp:
Also,

> Does the VM code walk the full XML to find all of the children?

Yes!

PR #938, opened 2017-06-10
Use extensions to share color and shape between some blocks

rachel-fenichel:
This is an example of how to use Blockly's extensions framework to share common block attributes.  In this case I'm using it for colors for a few categories, as well as for statement blocks.

I'm sure I didn't do the greatest job on naming.  I'm happy to change the extension names (and even the extension file + namespace) to the scratchier names.


paulkaplan:
WOOT 🥇

PR #945, opened 2017-06-13
Use blockly extensions for more block definitions

rachel-fenichel:
This is a continuation of #938 

I did not handle all blocks in the sensing category, because some seem to have the wrong output types.

Should be a no-op, but significantly shortens block definitions.

### DIFF
--- a/blocks_vertical/vertical_extensions.js
+++ b/blocks_vertical/vertical_extensions.js
@@ -31,7 +31,6 @@ goog.provide('Blockly.ScratchBlocks.VerticalExtensions');
 
 goog.require('Blockly.Colours');
 goog.require('Blockly.constants');
-goog.require('Blockly.Extensions');
 
 
 /**
--- a/blocks_vertical/vertical_extensions.js
+++ b/blocks_vertical/vertical_extensions.js
@@ -27,7 +27,17 @@ goog.require('Blockly.constants');
 goog.require('Blockly.Extensions');
 
 
+/**
+ * Helper function that generates an extension based on a category name.
+ * The generated function with set primary, secondary, and tertiary colours
+ * based on the category name.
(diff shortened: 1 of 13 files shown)


blocks_vertical/vertical_extensions.js
```diff
@@ -27,9 +34,29 @@ goog.require('Blockly.constants');
...
+ * Helper function that generates an extension based on a category name.
+ * The generated function with set primary, secondary, and tertiary colours
+ * based on the category name.
+ * @param {String} category The name of the category to set colours for.
+ * @return {function} An extension function that sets colours based on the given
+ *     category.
+ */
 Blockly.ScratchBlocks.VerticalExtensions.colourHelper = function(category) {
+  var colours = Blockly.Colours[category];
+  if (!(colours.primary && colours.secondary && colours.tertiary)) {
```
rachel-fenichel, line 46: Should there be an error thrown if the colours cannot be found?  This code just fails silently instead.

paulkaplan:
Just noticed that the previous extensions PR actually breaks when built because the core/extensions.js uses goog.isString which isn't compiling correctly. Not sure why?

blocks_vertical/vertical_extensions.js
paulkaplan: I think it would probably be a good idea to throw. Explicit errors > secret ones.

paulkaplan (approved):
LG 🙏 

Feel free to add that throw if you want. Also might want to just double check the fill color for shadow block checking, not sure why it is showing up as a diff here if it was merged earlier?

rachel-fenichel:
I'm merging while I've got an LGTM, and I'll follow up today by adding the throw and fixing #968.

PR #950, opened 2017-06-15
Set block color with extensions

rachel-fenichel:
New version of #938.

In #938 I added `core/extensions.js` to the set of files for building `blocks_compressed.js`, and in `blocks_vertical/vertical_extensions.js` I required `Blockly.Extensions`.

That caused build problems because `extensions.js` needed to pull in a whole bunch of other files.

I've removed the offending lines from this commit.  The code in `vertical_extensions.js` won't get optimized out during compression because it runs when it's loaded, and the calls to `Blockly.Extensions` will happen after Blockly has been loaded, while initializing blocks.  Using `blocks_compressed.js` without compressed or uncompressed Blockly is expected to fail, and that's fine.

### DIFF
--- a/blocks_vertical/control.js
+++ b/blocks_vertical/control.js
@@ -24,6 +24,7 @@ goog.provide('Blockly.Blocks.control');
 
 goog.require('Blockly.Blocks');
 goog.require('Blockly.Colours');
+goog.require('Blockly.ScratchBlocks.VerticalExtensions');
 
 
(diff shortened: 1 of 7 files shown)


rachel-fenichel:
Once this is merged #945 should stop being conflicted.

paulkaplan:
Ok, this seems to build correctly :)

Issue #959, opened 2017-06-16
FieldToDom doesn't include id and type for FieldVariableGetter 


marisaleung:
### Expected Behavior

When Blockly.Xml.blockToDom's fieldToDom is called on a block with a FieldVariableGetter inside, the variable's id and type should be included in the dom.
### Actual Behavior

The id and the type are not included, because the variable isn't even searched for.
### Steps to Reproduce

1. Open vertical playground.
2. Put a break point L110 at the very beginning of Blockly.Xml.blockToDom's fieldToDom().
3. Create a new variable by the name of 'a'.
4. Step through and see that we never enter the 'if (field instanceof Blockly.FieldVariable)' statement

### Operating System and Browser

Ubuntu Chrome Version 59.0.3071.86

Issue #958, opened 2017-06-16
FieldToDom searching flyout workspace instead of main workspace


marisaleung:
### Expected Behavior

When Blockly.Xml.blockToDom's fieldToDom is called on a block with a FieldVariable inside, the variable's id and type should be included in the dom.
### Actual Behavior

The variable is not found, because block.workspace.getVariable() searched the flyout workspace instead of the main workspace.
### Steps to Reproduce

1. Open vertical playground.
2. Put a break point in Blockly.Xml.blockToDom's fieldToDom() right before it calls getVariable().
3. Create a new variable by the name of 'a'.
4. Step through until after the L115 'var variable = block.workspace.getVariable(field.getValue());'
5. See that the variable is null. Because it wasn't found, because the flyout's workspace was checked instead of the main workspace.

### Operating System and Browser

Ubuntu Chrome Version 59.0.3071.86

PR #960, opened 2017-06-16
Develop fix fieldToDom

marisaleung:
### Resolves

https://github.com/LLK/scratch-blocks/issues/959
https://github.com/LLK/scratch-blocks/issues/958
### Proposed Changes

Checks main workspace instead of flyout workspace for a variable when getVariable is called.
Includes type and id for FieldVariableGetter field when converting from field to DOM.
### Reason for Changes

With this change, an extra variable is added to the flyout's workspace when the variable already exists in the main workspace.
This is problematic for the vm because when a new variable is created, the variable will be created in the vm with the correct id. And then the block that contains the variable in the toolbox will not be able to find it and then will fire a block create event that does not include all of the variable information.
### Test Coverage

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -278,6 +278,13 @@ Blockly.Flyout.prototype.init = function(targetWorkspace) {
   // A flyout connected to a workspace doesn't have its own current gesture.
   this.workspace_.getGesture =
       this.targetWorkspace_.getGesture.bind(this.targetWorkspace_);
+
+  // Get variables from the main workspace rather than the target workspace.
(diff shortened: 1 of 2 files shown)


marisaleung:
@rachel-fenichel @roboerikg @rschamp @paulkaplan

rachel-fenichel (approved):

rachel-fenichel:
LGTM but I'd like Paul to sanity check it.

rachel-fenichel (approved):

paulkaplan:
LG

PR #962, opened 2017-06-19
Use the secondary color for shadow block fills

rachel-fenichel:
Shadow blocks should always be filled with the secondary block color for that category.

This also means that blocks tat are always used as shadows don't need to have special color definitions.


paulkaplan (approved):
LG

paulkaplan:
Hey sorry my fault for LGing this. It seems to cause an issue with the background color of the shadow blocks, see screenshot. New on the left, old on the right.
image

Issue #964, opened 2017-06-19
Right-click on variable blocks is broken


rachel-fenichel:
### Expected Behavior

Right-clicking on a variable getter should show a menu with all of the other variable names in the workspace.

### Actual Behavior

Right-clicking does nothing.

### Steps to Reproduce

Create two variables.
Drag a variable getter block into the workspace.
Right-click on the variable getter block.

### Operating System and Browser

Chrome on Ubuntu.

carljbowman:
@rachel-fenichel - We should take a look at this as apart of the Variable spec.

rachel-fenichel:
Okay.  It used to do what Scratch 2.0 did, and the change that broke it is straightforward.

PR #965, opened 2017-06-19
Fix Variable Getter Context Menu Bug.

marisaleung:
### Resolves

https://github.com/LLK/scratch-blocks/issues/964

### Proposed Changes

Replaces workspace.variableList with workspace.getAllVariables() because variableList no longer exists. Access the variableModel's 'name' field for setting the text of the options in the Context Menu.

### Reason for Changes

The context menu was not showing up when right clicking a variable getter. This is because it was tryingto access workspace.variableList.length when variableList no longer existed and threw an error.

### Test Coverage

### DIFF
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -468,10 +468,10 @@ Blockly.Constants.Data.CUSTOM_CONTEXT_MENU_GET_VARIABLE_MIXIN = {
    */
   customContextMenu: function(options) {
     if (!this.isCollapsed()) {
-      var variablesList = this.workspace.variableList;
+      var variablesList = this.workspace.getVariablesOfType('');
       for (var i = 0; i < variablesList.length; i++) {
         var option = {enabled: true};
-        option.text = variablesList[i];
+        option.text = variablesList[i].name;
 
         option.callback =
             Blockly.Constants.Data.VARIABLE_OPTION_CALLBACK_FACTORY(this,


rachel-fenichel (approved):

blocks_vertical/data.js
```diff
@@ -468,10 +468,10 @@ Blockly.Constants.Data.CUSTOM_CONTEXT_MENU_GET_VARIABLE_MIXIN = {
    */
   customContextMenu: function(options) {
     if (!this.isCollapsed()) {
-      var variablesList = this.workspace.variableList;
+      var variablesList = this.workspace.getAllVariables();
```
RoboErikG: This should probably get all variables of '' type and then a different mixin should be added to lists to get variables of that type.

marisaleung: Done

blocks_vertical/data.js
```diff
@@ -486,6 +486,40 @@ Blockly.Extensions.registerMixin('contextMenu_getVariableBlock',
...
+Blockly.Constants.Data.CUSTOM_CONTEXT_MENU_GET_LIST_MIXIN = {
+  /**
+   * Add context menu option to create getter block for the loop's list.
+   * (customContextMenu support limited to web BlockSvg.)
+   * @param {!Array} options List of menu options to add to.
+   * @this Blockly.Block
+   */
+  customContextMenu: function(options) {
+    if (!this.isCollapsed()) {
+      var listOfLists = this.workspace.getVariablesOfType('list');
```
rachel-fenichel: This is the same as the function above except for 'list' vs '' on this line.  It looks like you want a function factory instead of duplication.

See #945 for an example: 
https://github.com/rachel-fenichel/scratch-blocks/blob/f8723d0dccb8a9746f78a503b84407bc56725a71/blocks_vertical/vertical_extensions.js#L152

https://github.com/rachel-fenichel/scratch-blocks/blob/f8723d0dccb8a9746f78a503b84407bc56725a71/blocks_vertical/vertical_extensions.js#L44

https://github.com/rachel-fenichel/scratch-blocks/blob/f8723d0dccb8a9746f78a503b84407bc56725a71/blocks_vertical/vertical_extensions.js#L58

blocks_vertical/data.js
```diff
@@ -486,6 +486,40 @@ Blockly.Extensions.registerMixin('contextMenu_getVariableBlock',
...
 
 /**
+ * Mixin to add a context menu for a data_list block.  It adds one item for
+ * each list defined on the workspace.
+ * @mixin
+ * @augments Blockly.Block
+ * @package
+ * @readonly
+ */
+Blockly.Constants.Data.CUSTOM_CONTEXT_MENU_GET_LIST_MIXIN = {
```
RoboErikG: Whoops, sorry my comment wasn't clear. This part should be future work (following Rachel's suggestion below) but since it's not used in this CL you don't need to add it yet.

marisaleung: That makes more sense. Alright, will revert to before I added the context menu mixin for lists.

blocks_vertical/data.js
marisaleung: This is really helpful and I will refer to this when we add this mixin later.

PR #966, opened 2017-06-19
Remove all audio effects except for pitch and pan

ericrosenbaum:
### Resolves

To address performance issues that cause the audio to drop out (as discussed [here](https://github.com/LLK/scratch-audio/issues/22) and [here](https://github.com/LLK/scratch-audio/issues/27)), we are removing the echo, reverb, fuzz and robot effects. 

### Proposed Changes

Remove the echo, reverb, fuzz, and robot effects.

### Reason for Changes

These effects create large numbers of webaudio nodes, which contributes to audio glitches.


rachel-fenichel (approved):
:(

paulkaplan (approved):
LG

PR #967, opened 2017-06-19
Add The Undefined Block

TheBrokenRail:
### Resolves

_Fixes #756_

### Proposed Changes

_If a block is invalid it will create a undefined block_

### Reason for Changes

_So projects with invalid blocks are not immediately broken when loaded_


towerofnix:
Looks like this was accidentally closed by the merge/06-28 (the commit that "fixes" this was targeting an issue in google/blockly, not the issue with the same ID in LLK/scratch-blocks.)

thisandagain:
@TheBrokenRail I think we are going to handle this in the project parser / validator rather than in the blocks to avoid the fatal exception.

Issue #969, opened 2017-06-19
WorkspaceSvg CreateVariable() doesn't include type and id


marisaleung:
### Expected Behavior

The created variable will have the id and type that is passed into it. 

### Actual Behavior

When workspaceSvg.createVariable() is called with a name, type, and id, only the name is used to created the variable. The id and type are ignored.

### Steps to Reproduce

You can just look at the function workspaceSvg.createVariable and see that the id and type are not parameters that can be passed in.

But for evidence of this problem in the playground, do the following steps.
1. Create a variable.
2. Drag a set block of that variable onto the workspace.
3. Click 'Export to XML.'
4. Take note of the id of the variable.
5. Refresh the page.
6. Click 'Import from XML.'
7. Click 'Export to XML.'
8. Take note that the id of the variable is different.

### Operating System and Browser

Ubuntu Chrome

PR #970, opened 2017-06-19
Develop fix workspace svg

marisaleung:
### Resolves

https://github.com/LLK/scratch-blocks/issues/969

### Proposed Changes

Includes type and id when creating a variable in workspaceSvg

### Reason for Changes

Without this change, a new id will be generated which can cause duplicates in the vm (two variables with same name and different id's.)

### Test Coverage


RoboErikG:
You seem to have pulled in some not yet merged variable event changes. Edit: Oh, that seems to be the point. ;)

rschamp (approved):

PR #972, opened 2017-06-19
Fix #968

### LINKED ISSUE #968: Text fields are the wrong colour
Left: actual behaviour.  Right: expected behaviour.
image

Directly caused by #962 
If a block doesn't have secondary and tertiary colours defined we darken the primary colour.  Blocks that use `Blockly.Colours.textField` seem to mostly just have a single colour defined, but are used as shadows.

A short-term fix is to revert #962.  

A better fix is to add a `colours_textfield` colour category and extension that uses the same colours for primary, secondary, and tertiary.  I'm not going to do that until #945 lands because #945 has a refactor of vertical_extensions.js, but it's easy to do once that's in.

rachel-fenichel:
### Resolves

#968 

### Proposed Changes

For blocks under blocks_common I'm adding secondary and tertiary colours explicitly.
In blocks_vertical/vertical_extensions.js I added an extension for setting text field colours.
For blocks under blocks_vertical I'm using that new extension.

I added an explicit error if colours are undefined.

### Reason for Changes

If blocks only have one colour defined Scratch-Blocks will darken that colour to get the secondary and tertiary colours, which makes text fields look terrible.

### Test Coverage

I checked that an extension is thrown when a single colour in a category or an entire category of colours is missing.
I loaded up scratch-gui to make sure that the built files work.

### DIFF
--- a/blocks_vertical/vertical_extensions.js
+++ b/blocks_vertical/vertical_extensions.js
@@ -35,7 +35,7 @@ goog.require('Blockly.constants');
 
 /**
  * Helper function that generates an extension based on a category name.
- * The generated function with set primary, secondary, and tertiary colours
+ * The generated function will set primary, secondary, and tertiary colours
  * based on the category name.
  * @param {String} category The name of the category to set colours for.
  * @return {function} An extension function that sets colours based on the given
@@ -43,12 +43,8 @@ goog.require('Blockly.constants');
  */
 Blockly.ScratchBlocks.VerticalExtensions.colourHelper = function(category) {
   var colours = Blockly.Colours[category];
-  if (!(colours.primary && colours.secondary && colours.tertiary)) {
-    /**
-     * Return an empty function, rather than throwing an error later.
-     * @this {Blockly.Block}
-     */
-    return function() { };
+  if (!(colours && colours.primary && colours.secondary && colours.tertiary)) {
+    throw new Error('Could not find colours for category "' + category + '"');
   }
   /**
    * Set the primary, secondary, and tertiary colours on this block for the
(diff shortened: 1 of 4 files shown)


rachel-fenichel:
I'm using explicit colours instead of the new extension in blocks_common so I'm not calling in to blocks_vertical to create the common blocks.

Future question: should the extensions I've added all be in blocks_common instead?  Do they all apply to blocks regardless of whether they are oriented horizontally or vertically?

If yes, what should the namespace be?  `Blockly.ScratchBlocks.Extensions` is going to create even more confusion about what an extension is (scratch extensions vs blockly extensions).

paulkaplan:
Yeah that naming could get confusing fast. Although, I _think_ the scratch version of extensions should be a transparent concept to scratch-blocks? On the other hand, if we go with a name like `BlockMixins` or something, it might also be confusing since at some point it will encounter the `Extensions` convention?

blocks_vertical/vertical_extensions.js
```diff
@@ -151,6 +155,11 @@ Blockly.ScratchBlocks.VerticalExtensions.registerAll = function() {
     Blockly.Extensions.register('colours_' + name,
         Blockly.ScratchBlocks.VerticalExtensions.colourHelper(name));
   }
+
+  // Text fields transcend categories.
```
paulkaplan, line 159: transcend

paulkaplan (approved):
LG

rachel-fenichel:
#980 for the followup question.


I agree that Scratch extensions shouldn't be conflicting with Blockly extensions in code, since scratch-blocks will just see Scratch extensions as normal blocks.  I do think there's a naming conflict because some of the same people will be building Scratch extensions and Scratch-Blocks extensions.

PR #974, opened 2017-06-20
Use extensions for the sensing category

rachel-fenichel:
I skipped this file on a previous pass because I think some of the output types are wrong.  On this change I've preserved the existing output types, even if they are wrong, and switched to using blockly extensions for shape and colour.

### DIFF
--- a/blocks_vertical/sensing.js
+++ b/blocks_vertical/sensing.js
@@ -67,10 +67,7 @@ Blockly.Blocks['sensing_touchingobjectmenu'] = {
             ]
           }
         ],
-        "colour": Blockly.Colours.sensing.secondary,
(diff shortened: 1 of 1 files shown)


paulkaplan:
So I'm not so familiar with the sensing category blocks and outputs. What blocks are you suspicious about?

rachel-fenichel:
"distance to" is returning a string.
"answer" is returning a number.
"video on menu one" and video on menu two" are returning strings.
"video toggle menu" is returning a string.
"username" is returning a number.

paulkaplan (approved):
Welp, at least those wrong outputs are documented, and we'll know where to go when it comes up again. Hi future us  👋

Issue #919, opened 2017-06-02
Top half of block cut off when using a statement input inside a square reporter


TheBrokenRail:
### Expected Behavior

_You see the whole block_

### Actual Behavior

_The top half is cut off, statements offset, and bottom has no padding_

### Steps to Reproduce

_Add my code to blocks_vertical/motion.js and and create the block_

### My Code

```js
Blockly.Blocks['motion_lambda'] = {
  init: function() {
    this.jsonInit({
      "message0": "do %1",
      "message1": "and return %1",
      "args0": [
        {
          "type": "input_statement",
          "name": "FUNCTION"
        }
      ],
      "args1": [
        {
          "type": "input_value",
          "name": "RETURN"
        }
      ],
      "inputsInline": false,
      "category": Blockly.Categories.motion,
      "output": "String",
      "outputShape": Blockly.OUTPUT_SHAPE_SQUARE,
      "colour": Blockly.Colours.motion.primary,
      "colourSecondary": Blockly.Colours.motion.secondary,
      "colourTertiary": Blockly.Colours.motion.tertiary
    });
  }
};
```

### Screenshot
![block](https://cloud.githubusercontent.com/assets/17478432/26744900/1c4d6284-47b6-11e7-9313-5b3b43177714.PNG)

### Operating System and Browser

_Google Chrome and Windows 10_

rachel-fenichel:
Long shot: @tmickel do you know what's going on here?

tmickel:
Hmmm... well...not precisely...but I think it is safe to say we did the rendering code for blocks with outputs (i.e. reporters) assuming that the shape would be a single row.

To start imagining reporters with multiple rows is a lot of work. It might be intuitive for a "square"-shaped reporter like the one above (just don't render notches?) but for rounded ones and "hexagonal" boolean ones, it's very strange to think about having statement inputs like above...

maybe it could be considered to add a check that a block with output contains no statement inputs and only a single row; if this is not the case, it's not currently supported by the rendering.

PR #976, opened 2017-06-20
Use only JSON for pen block definitions

rachel-fenichel:
This is a followup on my previous changes that made block definitions shorter.  

I'm not sure that scratch should go fully to using JSON block definitions, but if you did it would look like this.  I've tagged people who may have strong opinions about developer ergonomics/block definition.

I've moved the "extensions" attribute to be closer to the start of the block definition, since it's broadly defining how a block looks and interacts with other blocks.

As I mentioned before, I don't have a strong opinion about the names that I've given the extensions (`colours_pen`, `shape_statement`, `output_number`, etc).  If there are other names that are more accurate for scratch team, I'm happy to use those.

The change in vertical_extensions.js makes setting the colour also set the category attribute, which removed one line per block definition.

### DIFF
--- a/blocks_vertical/pen.js
+++ b/blocks_vertical/pen.js
@@ -23,203 +23,115 @@
 goog.provide('Blockly.Blocks.pen');
 
 goog.require('Blockly.Blocks');
-goog.require('Blockly.Colours');
-goog.require('Blockly.constants');
 goog.require('Blockly.ScratchBlocks.VerticalExtensions');
 
-
-Blockly.Blocks['pen_clear'] = {
-  /**
-   * Block to clear drawing.
-   * @this Blockly.Block
-   */
-  init: function() {
-    this.jsonInit({
-      "message0": "clear",
-      "category": Blockly.Categories.pen,
-      "extensions": ["colours_pen", "shape_statement"]
-    });
+Blockly.defineBlocksWithJsonArray([
+  // Block to clear drawing.
+  {
+    "type": "pen_clear",
+    "extensions": ["colours_pen", "shape_statement"],
+    "message0": "clear"
+  },
+  // Block to stamp a sprite
+  {
+    "type": "pen_stamp",
+    "extensions": ["colours_pen", "shape_statement"],
+    "message0": "stamp"
+  },
+  // Block to pull down the sprite's pen.
+  {
+    "type": "pen_pendown",
+    "extensions": ["colours_pen", "shape_statement"],
+    "message0": "pen down"
+  },
+  // Block to pull up the sprite's pen.
+  {
+    "type": "pen_penup",
+    "extensions": ["colours_pen", "shape_statement"],
+    "message0": "pen up"
+  },
+  // Block to set the pen's color to the value.
+  {
+    "type": "pen_setpencolortocolor",
+    "extensions": ["colours_pen", "shape_statement"],
+    "message0": "set pen color to %1",
+    "args0": [
+      {
+        "type": "input_value",
+        "name": "COLOR"
+      }
+    ]
+  },
+  // Block to change the pen's color by the value.
+  {
+    "type": "pen_changepencolorby",
+    "extensions": ["colours_pen", "shape_statement"],
+    "message0": "change pen color by %1",
+    "args0": [
+      {
+        "type": "input_value",
+        "name": "COLOR"
+      }
+    ]
+  },
+  // Block to set the pen's color to the value.
+  {
+    "type": "pen_setpencolortonum",
+    "extensions": ["colours_pen", "shape_statement"],
+    "message0": "set pen color to %1",
+    "args0": [
+      {
+        "type": "input_value",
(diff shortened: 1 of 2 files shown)


rschamp (approved):
Thanks @rachel-fenichel! This looks nice to me... I think if we can get away with defining them in JSON, we may as well for those blocks that can be?

Curious if @paulkaplan @ericrosenbaum or @cwillisf have strong opinions about the JSON format?

cwillisf (approved):
LGTM. In particular it seems like it would be fairly easy to generate JSON objects like this from the info we'll have available in the Scratch extension definition objects as proposed.

rachel-fenichel:
I think it'll be easy to generate the simple blocks, like pen up and pen down.  What about blocks with inputs?

cwillisf:
We haven't completely settled on how we'll specify inputs in our block string. The current proposal is something like `'give {MY_ARG,string} to my reporter'` but based on recent conversations that's likely to change. At any rate, it'll be something which conveys some of that info, and the rest of the necessary info will be outside the string somewhere nearby.

I think the most painful part will be replacing our placeholder with your `%1`, but that's just a bit of text manipulation and we can do it once at extension-load time. Generating your `args0` list looks like it'll be a relatively straightforward matter of dropping in the name of the arg along with a lookup into a type mapping. Also, if we end up switching out the currently-proposed complicated placeholders with stuff like `%1` then maybe the string manipulation will be trivial...

The Scratch extensions proposal doesn't currently cover it, but we'll need to add a way to indicate the kind of block being defined: hat, C/E, regular statement, and so on. If I understand correctly that (along with the block category) will influence the list of Blockly extensions to be applied, which seems easy enough.

cwillisf:
BTW, the Scratch extensions proposal currently lives here:
https://github.com/LLK/scratch-vm/wiki/Extensions-Proposal

Also potentially of interest:
https://github.com/LLK/scratch-vm/wiki/Extension-translation-straw-cat

rachel-fenichel:
You may not have an answer to this yet, but: are you going to have a list of every shape of block that may be created, and disallow all other shapes?

For instance, you can have a C or E shaped block.  What about the next step, where it has three statement inputs?  (The current best name for it is an
```
E
L
```
block.)

rachel-fenichel:
Awesome, I'll take another look at those.

Note that I'm not solely concerned about blocks created as Scratch extensions.  I'm also thinking about block definition for anyone who uses just scratch-blocks as their library.

cwillisf:
I plan to explicitly support CELL blocks ;)

Serious answer: we haven't fully resolved that yet but we're leaning toward only allowing Scratch extensions to specify Scratch's "usual" shapes, probably with an enum or something similar, plus a flag or two for things like whether or not the block terminates a stack. There's a chance we'll have a unified "choose" type which covers things like `if`, `if-else`, and `switch`, but I expect that's the craziest it'll get.

rachel-fenichel:
>we're leaning toward only allowing Scratch extensions to specify Scratch's "usual" shapes

And non-scratch extensions?

cwillisf:
> And non-scratch extensions?

Sorry, I was just using "Scratch extensions" to distinguish "bag o' blocks" extensions from Blockly's "do something to a block" extensions. As far as I'm aware, we're only planning on supporting "bag o' blocks" extensions in a scratch-gui / scratch-vm kind of context. Specifically, I've been operating under the assumption that scratch-blocks would be unaware that such a beast exists, and that it would be our problem (scratch-vm's & scratch-gui's problem) to present block definitions to scratch-blocks in a way that makes extension blocks and "core" blocks (like "move 10 steps") more or less indistinguishable. We've even discussed the possibility of implementing all Scratch 3.0 blocks in terms of the "extension" system.

In that respect, as long as scratch-blocks supports at least as many shapes as we plan to expose to Scratch extension authors, I'm happy. I don't see a need for scratch-blocks to impose restrictions beyond what Blockly does -- we'll impose those restrictions on the "client" side of this API.

rachel-fenichel:
Okay, sounds good to me.

Yes, there's a language problem with scratch extensions vs. blockly extensions.  I'm trying to find a good way to distinguish between the two in normal conversation.

RoboErikG:
And one more note on the topic of supporting block shapes, we've already seen a few cases of developers trying to make a block that isn't supported in Scratch Blocks (example: https://github.com/LLK/scratch-blocks/issues/919) but I don't think there's a full spec for what will be supported in SB yet (independent of what Scratch 3.0 supports).

fsih:
€ blocks?

paulkaplan:
@rachel-fenichel I'm concerned about using multiple different ways to define blocks, we may want to favor consistency. One difficult thing already about the block-defining process is the number of different ways one can do it. Maybe we should just wait on/work with @cwillisf on the extension definition level?

paulkaplan:
After talking offline with @rachel-fenichel, I think we should try to err on the side of consistency and leave these definitions the way they are.

PR #977, opened 2017-06-21
Don't fire events for insertion marker creation and deletion

### LINKED ISSUE #934: Undo/Redo functionality is broken
### Expected Behavior

Using undo should undo 1 user action. Pressing redo should redo the last undo action.

### Actual Behavior

Lots of strange things are happening. In the simplest example of just placing some blocks pressing undo removes some of the inner shadow blocks, and also places blocks in the upper-left.

### Steps to Reproduce

1) Launch the playground from current sources
2) Drag several blocks onto the stage: say 1 move block + 1 turn left block + 1 turn right block
3) Right-click on the workspace and click undo
4) Bug -> notice the last placed blocked is not removed but instead the inner shadow block disappears. 

Further undo actions after this create more buggy behavior:
5) Next undo seems to actually remove the block
6) Next 2 undo's behave like the 1st 2
7) 5th undo makes a block appear in the upper-left

Additionally, clicking Redo does not redo the actions (bugged or not). It might not be worth commenting more about this given undo is broken. 

Chrome on Mac

rachel-fenichel:
### Resolves

Fixes #934 
### Proposed Changes

Disable events before creating or deleting insertion markers; re-enable events after insertion marker creation or deletion.

### Reason for Changes

We were getting spurious events for the insertion markers, and also breaking undo/redo

### DIFF
--- a/core/insertion_marker_manager.js
+++ b/core/insertion_marker_manager.js
@@ -26,6 +26,7 @@
 
 goog.provide('Blockly.InsertionMarkerManager');
 
+goog.require('Blockly.Events');
 goog.require('Blockly.RenderedConnection');
 
 goog.require('goog.math.Coordinate');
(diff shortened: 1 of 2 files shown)


core/insertion_marker_manager.js
```diff
@@ -157,14 +158,21 @@ Blockly.InsertionMarkerManager.prototype.dispose = function() {
...
+  try {
+    if (this.firstMarker_) {
+      this.firstMarker_.dispose();
+      this.firstMarker_ = null;
+    }
+    if (this.lastMarker_) {
+      this.lastMarker_.dispose();
+      this.lastMarker_ = null;
+    }
+  } finally {
```
paulkaplan, line 172: TIL `try/finally`!

paulkaplan (approved):
LG! Undo/Redo looks great now.

core/insertion_marker_manager.js
rachel-fenichel: Yup!  You should use try/finally almost any time that you're disabling events in blockly, because event disable/enable isn't a boolean.  It's a counter, so you need to have symmetrical disables and enables.  If an exception is throw between the disable and the enable and there's no `finally` clause, you may never get events again.

Issue #975, opened 2017-06-20
Variable Deletion Spec


marisaleung:
There are two options for when variables should come back that we are discussing and would like your input or specification.

1. Letting it come back next time the toolbox is refreshed.

2. Explicitly comes back when a block with that variable runs.

I understand the second one is how scratch 2 works. The first option is  easier to implement since we can just do a light version of delete by just removing variable blocks from the flyout instead of removing them from the variable map.

mrjacobbloom:
Potentially relevant: https://github.com/LLK/scratch-blocks/issues/372#issuecomment-227532505
> Update on deletion: we agreed to make deleting a variable remove all instances of it in the workspace (after confirmation) but not in the backpack.

> Hmmmmm... what if, the drop-down changes to "blank"?

>That seems non-obvious, especially in a large script. Also what does it mean for the dropdown to say "blank", from a runtime perspective? ... If it were blank I would suggest also turning it red (shouldn't be too hard) but that seems counter to scratch philosophy of no errors.

RoboErikG:
The Scratch team has been revisiting that discussion. We're waiting for confirmation from Carl on the exact behavior they want to move forward with.

RoboErikG:
Per in person discussion, we'll move forward with 1. which is easier to implement and try it out. If we don't like the behavior we can move to a more involved solution.

PR #998, opened 2017-07-06
Add variable disabled attribute and interactions with it.

marisaleung:
### Resolves

Implements the spec discussed here: https://github.com/LLK/scratch-blocks/issues/975
Deleted variables reappear on toolbox refresh if they are deleted and still exist in the workspace.

### Proposed Changes

- Add a disabled attribute to VariableModel.
- When a variable is disabled, it does not show up in dropdowns or the toolbox. If a dropdown in the workspace already has that variable selected, that instance will remain.
- A variable is disabled when the user tries to delete a variable at the same time that there is a workspace instance of it.
- When the variable is enabled, it will reappear in the dropdowns and toolbox.
- A variable is enabled when the toolbox refreshes and when it is selected in a dropdown on the workspace.
- A variable is deleted if it is disabled and the last instance of it in the workspace is removed. For example:
    1. If the last instance is a block and it is deleted.
    2. If the last instance is a dropdown with that variable selected, and the a different variable is selected. 
- Undo and redo features are supported.

### Reason for Changes

Made to match the spec in the issue above. This is so variable can come back in the toolbox if a user accidentally deleted it when that variable is still used in the workspace.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -229,6 +229,8 @@ Blockly.Block.prototype.dispose = function(healStack) {
     // Already deleted.
     return;
   }
+  var variableField = this.getField('VARIABLE');
+  var workspace = this.workspace;
   // Terminate onchange event calls.
   if (this.onchangeWrapper_) {
     this.workspace.removeChangeListener(this.onchangeWrapper_);
@@ -278,6 +280,12 @@ Blockly.Block.prototype.dispose = function(healStack) {
     }
   } finally {
     Blockly.Events.enable();
+    if (variableField) {
+      var variable = workspace.getVariable(variableField.text_);
+      if (variable && !variable.isEnabled()) {
+        workspace.deleteDisabledVariableIfUnused(variable);
+      }
+    }
   }
 };
 
--- a/core/events.js
+++ b/core/events.js
@@ -121,6 +121,18 @@ Blockly.Events.VAR_DELETE = 'var_delete';
 Blockly.Events.VAR_RENAME = 'var_rename';
 
 /**
+ * Name of event that disables a variable.
+ * @const
+ */
+Blockly.Events.VAR_DISABLE = 'var_disable';
+
+/**
+ * Name of event that enables a variable.
+ * @const
+ */
+Blockly.Events.VAR_ENABLE = 'var_enable';
+
+/**
  * Name of event that records a UI change.
  * @const
  */
@@ -326,6 +338,12 @@ Blockly.Events.fromJson = function(json, workspace) {
       break;
     case Blockly.Events.VAR_RENAME:
       event = new Blockly.Events.VarRename(null);
+      break;
+    case Blockly.Events.VAR_DISABLE:
+      event = new Blockly.Events.VarDisable(null);
+      break;
+    case Blockly.Events.VAR_ENABLE:
+      event = new Blockly.Events.VarEnable(null);
       break;
     case Blockly.Events.UI:
       event = new Blockly.Events.Ui(null);
@@ -966,7 +984,11 @@ Blockly.Events.VarCreate.prototype.run = function(forward) {
   if (forward) {
     workspace.createVariable(this.varName, this.varType, this.varId);
   } else {
+    // In case there is a var_enable event in the same group, prevent the
+    // flyoutCategory from re-enabling the variable.
+    Blockly.Variables.allowEnable = false;
     workspace.deleteVariableById(this.varId);
+    Blockly.Variables.allowEnable = false;
   }
 };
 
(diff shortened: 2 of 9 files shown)


rachel-fenichel:
>A variable is enabled when the toolbox refreshes and it is selected in a dropdown.

Nit: selected in a dropdown on the workspace, right?

> A variable is deleted if it is disabled and the last instance of it in the workspace is removed. For example:

What about variables that exist across multiple workspaces?  If the last instance of the variable is deleted in the currently selected workspace but it's still in use in another workspace?

core/block.js
```diff
@@ -278,6 +280,12 @@ Blockly.Block.prototype.dispose = function(healStack) {
     }
   } finally {
     Blockly.Events.enable();
+    if (variableField) {
```
rachel-fenichel, line 283: Why does this happen after the event enable?

core/events.js
```diff
@@ -1086,6 +1095,43 @@ Blockly.Events.VarRename.prototype.run = function(forward) {
 };
 
 /**
+ * Class for a variable disable event.
```
rachel-fenichel: Add a note that this is scratch-specific/that mainline Blockly has no "disabled variable" concept.

core/events.js
```diff
@@ -1086,6 +1095,43 @@ Blockly.Events.VarRename.prototype.run = function(forward) {
...
+ */
+Blockly.Events.VarDisable.prototype.type = Blockly.Events.VAR_DISABLE;
+
+/**
+ * Run a variable disable event.
+ * @param {boolean} forward True if run forward, false if run backward (undo).
+ */
+Blockly.Events.VarDisable.prototype.run = function(forward) {
+  var workspace = this.getEventWorkspace_();
+  if (!workspace) return;
```
rachel-fenichel: Put the return on a new line and add braces around it.

The style guide technically says this is okay, but it's not the style that the rest of Blockly uses.

marisaleung:
Correct, updated comment. And this is a good point. We decided to ignore the multiple workspaces for now and implement everything else since the multiple workspaces requires a listener for the vm. If Scratch decides they like the variable reappearing on refresh, then we will continue and add that listener.

core/variable_model.js
```diff
@@ -87,6 +94,21 @@ Blockly.VariableModel.prototype.getId = function() {
...
+Blockly.VariableModel.prototype.disable = function() {
+  this.disabled = true;
+  Blockly.Events.fire(new Blockly.Events.VarDisable(this));
+};
+
+/**
+ * Enable the variable.
+ */
+Blockly.VariableModel.prototype.enable = function() {
+  this.disabled = false;
```
rachel-fenichel: No event is fired when you enable a variable?

core/field_variable.js
```diff
@@ -163,9 +163,11 @@ Blockly.FieldVariable.dropdownCreate = function() {
   }
   variableModelList.sort(Blockly.VariableModel.compareByName);
   var options = [];
-  for (var i = 0; i < variableModelList.length; i++) {
+  for (var i = 0, variable; variable = variableModelList[i]; i++) {
+    if (!variable.disabled || this.getValue() == variable.name) {
```
rachel-fenichel: You check ` if(!variable.disabled)` quite a few times in this PR.  

It would be easier to parse `if (variable.isEnabled())`.  Also (per offline discussion) switch to storing `enabled` instead of `disabled` to avoid double negatives.

core/workspace.js
```diff
@@ -401,19 +388,35 @@ Blockly.Workspace.prototype.deleteVariableById = function(id) {
...
+ * Delete a disabled variable if there are no more uses of it.
+ * @param {Blockly.VariableModel} variable Variable to delete.
+ */
+Blockly.Workspace.prototype.deleteDisabledVariableIfUnused = function(variable) {
+  var uses = this.getVariableUses(variable.name);
+  if (uses.length == 0) {
+    this.deleteVariableInternal_(variable);
+  }
+};
+
```
rachel-fenichel, line 400: Remove extra newlines.

core/workspace.js
```diff
@@ -401,19 +388,35 @@ Blockly.Workspace.prototype.deleteVariableById = function(id) {
...
  * @private
  */
 Blockly.Workspace.prototype.deleteVariableInternal_ = function(variable) {
   var uses = this.getVariableUses(variable.name);
-  Blockly.Events.setGroup(true);
-  for (var i = 0; i < uses.length; i++) {
-    uses[i].dispose(true, false);
+  // If the variable is disabled, the delete occurs with another event such as
+  // removing the last workspace instance of the variable.
+  if (!variable.disabled) Blockly.Events.setGroup(true);
```
rachel-fenichel: Add braces and move the inside of the if to a new line, here and below.

core/block.js
marisaleung: It is after the enable for two reasons:
1. I want the VAR_DELETE event to be fired if it happens.
2. deleteDisabledVariableIfUnused() checks if there are any workspace uses of the variable, after enable, this block is gone so the number of uses will be 0. If it is put before, the the disable() the number of uses will include this block and be 1.

core/events.js
marisaleung: Done.

core/events.js
marisaleung: Thanks! Done.

core/workspace.js
marisaleung: Done

core/workspace.js
marisaleung: Done.

core/field_variable.js
marisaleung: Done

core/variable_model.js
marisaleung: This is a good point. I found a bug from that thank you! Added a an enable event.

marisaleung:
Uploading more changes:
- Added a var_enable event
- Renamed 'disabled' attribute to 'enabled' to avoid double negatives.
- Added a Blockly.Variables.allowEnable attribute so when flyoutCategory is run right after a variable is disabled, it doesn't automatically re-enable the variable.
    - Also added "Blockly.Variables.allowEnable = false" in Blockly.Events.VarCreate() because if a VarDisable() occurred in the same group, the following would happen during an undo:
        1. Disable variable
        2. Blockly.Variables.allowEnable = false
        3. Refresh toolbox -> run flyoutCategory
        4. Blockly.Variables.allowEnable = true
        5. Delete Variable
        6. Refresh toolbox -> run flyoutCategory
        7. Enable variable that was just disabled.

paulkaplan:
Is it possible to pass the `enabled` flag in through XML properties in the `<variables>` section?

marisaleung:
Since this is a change specifically for scratch-blocks feel free to add one, but we will not be adding it to blockly. Scratch gui should work without it in the variables section.

paulkaplan:
I'm seeing an issue when I build this into GUI:
1. Create a variable
2. Immediately delete the variable (without dragging any blocks out of the palette)
3. Switch to the backdrop
4. Switch back to the cat scratch sprite, see the variable comes back but says "undefined"
image

Can't tell where this is problem is occurring, I'll need to track it down deeper.

rschamp:
This PR will be useful when we do implement this, but probably a bit too out of date to actually use as-is.

PR #1000, opened 2017-07-06
Correctly place blocks when the flyout and main workspace are at diferent zoom levels.

### LINKED ISSUE #880: Dragging from flyout when zoomed all the way out
### Expected Behavior

Block starts out approximately where the cursor is.

### Actual Behavior

Block is offset from the cursor by an amount that is related to the zoom level.

### Steps to Reproduce

Open llk.github.io/scratch-gui
Zoom out all the way
Drag a block from the flyout

Any and all

rachel-fenichel:
### Resolves

#880 

### Proposed Changes

Gets rid of `scroll / scale - scroll` math!  Instead, everything is expressed in terms of either pixels of workspace units/coordinates.  All of the positions are now translated to pixels, and math is done in the pixel realm.  The deltas are changed back to workspace units only when they are applied.

### Reason for Changes

The old math was bad and no one understood it, so we couldn't fix it.

### Test Coverage

Tested in RTL and LTR mode on the vertical playground.  In both cases I zoomed in and dragged in blocks, zoomed out and dragged in blocks, and scrolled both the main workspace and the flyout workspace before dragging blocks.

### DIFF
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -616,93 +616,60 @@ Blockly.VerticalFlyout.prototype.isDragTowardWorkspace = function(currentDragDel
 
 /**
  * Copy a block from the flyout to the workspace and position it correctly.
- * @param {!Blockly.Block} originBlock The flyout block to copy.
+ * @param {!Blockly.Block} oldBlock The flyout block to copy.
  * @return {!Blockly.Block} The new block in the main workspace.
  * @private
  */
-Blockly.VerticalFlyout.prototype.placeNewBlock_ = function(originBlock) {
+Blockly.VerticalFlyout.prototype.placeNewBlock_ = function(oldBlock) {
   var targetWorkspace = this.targetWorkspace_;
-  var svgRootOld = originBlock.getSvgRoot();
+  var svgRootOld = oldBlock.getSvgRoot();
   if (!svgRootOld) {
     throw 'originBlock is not rendered.';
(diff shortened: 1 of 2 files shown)


paulkaplan:
Code looks good and works. Can you quick explain how/why this is different from the same code on upstream blockly? https://github.com/google/blockly/blame/develop/core/flyout_vertical.js#L309-L383

paulkaplan (approved):

rachel-fenichel:
I'm changing upstream blockly to use reasonable math as well: https://github.com/google/blockly/pull/1215

All of the math here has been terrible and has predated my time on the project.  No one has ever been able to explain what it meant or why it worked the way it did--and every time I've tried to simplify it I've made it worse.  This math is also a large part of why making the Scratch-style toolbox was so painful--dragging blocks in would put them in completely random places on the screen.

The previous code worked as long as the scale was the same between the flyout workspace and the main workspace.  Blockly doesn't let them diverge so we never had a major problem (other than the frustration of not being able to understand our code).  Scratch Blocks allows the main workspace to zoom without the flyout workspace, which really exposed issues here.

TLDR: nope, can't explain that.

rachel-fenichel:
I'm 99% sure that the same code also works for horizontal mode, so in a followup PR I will probably move placeNewBlock into flyout_base and remove it from flyout_vertical and flyout_horizontal.

PR #1001, opened 2017-07-07
Make duplicated blocks stick to the cursor.

### LINKED ISSUE #828: Duplicated blocks should be attached to the mouse cursor

rachel-fenichel:
### Resolves

#828

### Proposed Changes

Duplicate blocks exactly on top of the existing block, then force a drag to start.

There's still a small jump, because the new block gets placed at the position of the old block instead of at the mouse position.  I'll look into that in a followup.

I think there's some cleanup I can do to decompose out code from this and #1000.  

### Reason for Changes

Matching Scratch 2.0 behaviour.

### Test Coverage

Tested in the vertical playground with categories and the simple multi_playground.

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -470,7 +470,7 @@ Blockly.BlockSvg.prototype.clearTransformAttributes_ = function() {
 /**
  * Snap this block to the nearest grid point.
  */
-Blockly.BlockSvg.prototype.snapToGrid = function() {
+Blockly.BlockSvg.prototype.snapToGrid  = function() {
   if (!this.workspace) {
     return;  // Deleted block.
   }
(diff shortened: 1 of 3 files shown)


core/block_svg.js
```diff
@@ -655,6 +655,87 @@ Blockly.BlockSvg.prototype.showHelp_ = function() {
...
+ */
+Blockly.BlockSvg.prototype.duplicateAndDragCallback_ = function() {
+  var oldBlock = this;
+  return function(e) {
+    // Give the context menu a chance to close.
+    setTimeout(function() {
+      var ws = oldBlock.workspace;
+      var svgRootOld = oldBlock.getSvgRoot();
+      if (!svgRootOld) {
+        throw 'oldBlock is not rendered.';
```
paulkaplan: The codebase seems inconsistent about this, but I would be in favor of never throwing bare strings and always throwing `new Error(<string>)`, to make sure that error catching can always consistently rely on catching an error object. Usually a linter rule enforces that.

core/block_svg.js
```diff
@@ -655,6 +655,87 @@ Blockly.BlockSvg.prototype.showHelp_ = function() {
...
+      // The target workspace would normally resize during domToBlock, which
+      // will lead to weird jumps.  Save it for terminateDrag.
+      ws.setResizesEnabled(false);
+
+      // Using domToBlock instead of domToWorkspace means that the new block
+      // will be placed at position (0, 0) in main workspace units.
+      var block = Blockly.Xml.domToBlock(xml, ws);
+      var svgRootNew = block.getSvgRoot();
+      if (!svgRootNew) {
+        throw 'block is not rendered.';
```
paulkaplan: Ditto throwing errors instead of strings

paulkaplan: Also, since I'm unfamiliar with the codebase, it would be nice if this error could be a bit more descriptive. When can someone expect a block to not have an svg root?

core/workspace_svg.js
```diff
@@ -1908,6 +1939,27 @@ Blockly.WorkspaceSvg.prototype.cancelCurrentGesture = function() {
...
+ *     drag, including clientX and clientY.
+ * @param {!Blockly.BlockSvg} block The block to start dragging.
+ * @package
+ */
+Blockly.WorkspaceSvg.prototype.startDragWithFakeEvent = function(fakeEvent,
+    block) {
+  Blockly.Touch.clearTouchIdentifier();
+  // Also sets the touch identifier
+  if (!Blockly.Touch.checkTouchIdentifier(fakeEvent)) {
+    console.log('something went wrong while setting the touch identifier');
```
paulkaplan: nit: it looks the non-test code all use `console.warn` instead of `log`, seems like this would deserve a warning. Similar to other error message comment, it might be good to include some hints (if you have any) about why that can fail.

core/block_svg.js
```diff
@@ -655,6 +655,87 @@ Blockly.BlockSvg.prototype.showHelp_ = function() {
...
+      var ws = oldBlock.workspace;
+      var svgRootOld = oldBlock.getSvgRoot();
+      if (!svgRootOld) {
+        throw 'oldBlock is not rendered.';
+      }
+
+      // Create the new block by cloning the block in the flyout (via XML).
+      var xml = Blockly.Xml.blockToDom(oldBlock);
+      // The target workspace would normally resize during domToBlock, which
+      // will lead to weird jumps.  Save it for terminateDrag.
```
paulkaplan: Do you mean that terminating a drag sets resize enabled back to true? (I can't find anything called `terminateDrag`)

paulkaplan (changes requested):
Few questions and nits

core/block_svg.js
rachel-fenichel: Do you know which eslint rule offhand?  If not I'll look into it.

I'm fine with using Errors.

core/block_svg.js
rachel-fenichel: Yeah it's now BlockDragger:endBlockDrag.

core/block_svg.js
rachel-fenichel: Only when the block is not rendered, and I don't know what would cause it to not be rendered at the end of the constructor without an error already having been thrown.  In debugging I would use this to get a stack trace and go from there.

I can make this error be specific to duplication for now, but I'm going to combine this code and the flyout code eventually.

I'd like some help with phrasing here.  I can change it to 'newBlock is not rendered' and rename `block` to `newBlock`, but it still doesn't have a lot of information.

core/workspace_svg.js
rachel-fenichel: Actually, I'm going to remove this warning entirely.  It's mostly a holdover from dragging code feeling unstable, and I'm less concerned about that right now.

core/block_svg.js
paulkaplan: Looks like `eslint no-throw-literal: "error"` http://eslint.org/docs/rules/no-throw-literal

paulkaplan (approved):

PR #1002, opened 2017-07-10
make variable id match block id in the flyout.

fsih:
### Resolves

Part of solution for https://github.com/LLK/scratch-vm/issues/623

### Proposed Changes

Change the block ID to match the variable ID when creating a variable block. This will make it easier to keep them in sync in the VM.


rachel-fenichel:
This looks reasonable but please note that this is only in the flyout, not in the main workspace.

Issue #854, opened 2017-03-21
Block detach distance threshold is too large


ericrosenbaum:
### Expected Behavior

When two blocks are connected, and you drag the bottom one down to detach, I expect to be able to release the block after dragging it beyond the edge of the insertion marker.

### Actual Behavior

In order to detach the block, I have to drag past the ~~shadow~~ insertion marker, about twice its height. This gif shows dragging the block almost this far, and it still snaps back:

![block-detach-distance](https://cloud.githubusercontent.com/assets/567844/24158635/42f090a8-0e33-11e7-946c-8d2bcc5c48c6.gif)

### Steps to Reproduce

Attempt to disconnect two blocks by dragging the bottom one downward.

### Operating System and Browser

Mac OS 10.11.6 Chrome 57

rachel-fenichel:
Code pointer:  https://github.com/LLK/scratch-blocks/blob/develop/core/block_svg.js#L1149

I would play around with this when the block is in the middle of a stack to make sure you want to change it.  I think it's protecting against jumping around as the blocks shift to accommodate the insertion marker*.

*not a shadow.  Shadow has a specific meaning in Blockly.

tmickel:
Original issue with video: https://github.com/LLK/scratch-blocks/issues/290

you might be able to come up with a more clever solution :) e.g., maybe the size could be minimized depending on the size of the insertion marker (smaller for vertical than horizontal), or maybe you could only apply this rule when the insertion marker is in the middle of a stack

rachel-fenichel:
Related: this transitions almost* directly from before the stack to wrapping the second block in the stack.

*almost because when it does wrap the whole stack, it does it for a frame or two and it's a pretty nasty flicker, but finding that space intentionally is almost impossible at normal zoom.

![63e22b70-ac8c-473f-a9b2-f28165ac3b43](https://user-images.githubusercontent.com/13686399/26859267-659659ea-4aeb-11e7-88cd-716800b8fe7d.gif)

PR #1003, opened 2017-07-10
Fix problem with insertion markers

### LINKED ISSUE #946: Insertion marker disconnects C shaped blocks when dragging with a block that has no prev or next statements
### Expected Behavior

Insertion marker shouldn't disconnect blocks when a block with no previous or next statement hovers over a number of nested C shaped blocks. 

### Actual Behavior

Insertion marker disconnects the C shaped blocks

### Steps to Reproduce

Easiest repro is to 
go to https://github.com/LLK/scratch-blocks/blob/develop/blocks_vertical/control.js#L60
control_forever, and remove "previousStatement": null,

Go to the playground and drag a number of C shaped blocks (eg: repeat 10) and nest them. 
Then drag a control -> forever block and drag that over each of the nested repeat blocks. 

insertionmarkers_issue

Mac OS 10.12.3 Chrome 58.0.3029.110

rachel-fenichel:
### Resolves

Fix #946

### Proposed Changes

If a block has a statement input and no previous connection, it is not allowed to connect in the middle of a stack of blocks.

### Reason for Changes

The old behaviour made blocks disconnect and fail to reconnect during a drag.

### Test Coverage
Tested in the vertical playground by forcing the repeat block to not have a previous connection.

The fix works, but in testing #854 was very obviously problematic.

64182cd7-5339-4202-801c-079d00d64467

### DIFF
--- a/core/connection.js
+++ b/core/connection.js
@@ -381,7 +381,9 @@ Blockly.Connection.prototype.canConnectToPrevious_ = function(candidate) {
 
   if (isFirstStatementConnection && sourceHasPreviousConn) {
     return true;
-  } else if (isNextConnection ||
+  }
+
+  if (isNextConnection ||
       (isFirstStatementConnection && !sourceHasPreviousConn)) {
     // If the candidate is the first connection in a stack, we can connect.
     if (!candidate.targetConnection) {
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):
Cool cool. 

Because the first statement is just `if... return` I think we should drop the `else` in 384 and just make a separate if block. Helps understanding because you do not need to know if there is a final `else {` to understand the code. But up to you.

rachel-fenichel:
Good point.  Updated.

PR #1015, opened 2017-07-19
Set primary, secondary and tertiary colors on setValue

### LINKED ISSUE #986: The color picker didn't set to the color selected
### Expected Behavior

The color picker 'set pen color to ' block in pen catalog didn't reponse to the color selected

### Actual Behavior

It should, but the rounded color picker just keep red. It is interesting that argValues give the correct color value.

### Steps to Reproduce
Take a try at pen catalog from vertical playground.
http://llk.github.io/scratch-blocks/playgrounds/tests/vertical_playground.html

Windows 10 and chrome 58

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/986

### Proposed Changes

Changes the `setValue` function of the `FieldColour` to set all 3 colors (primary, secondary and tertiary) instead of just the primary color.

### Reason for Changes

The secondary and tertiary colors are not used by the `FieldColour`, but the renderer expects to be able to use the secondary color as the fill for a shadow. See linked issue for more.

### DIFF
--- a/core/field_colour.js
+++ b/core/field_colour.js
@@ -106,11 +106,8 @@ Blockly.FieldColour.prototype.setValue = function(colour) {
   }
   this.colour_ = colour;
   if (this.sourceBlock_) {
-    this.sourceBlock_.setColour(
-      colour,
-      this.sourceBlock_.getColourSecondary(),
-      this.sourceBlock_.getColourTertiary()
-    );
+    // Set the primary, secondary and tertiary colour to this value.
+    this.sourceBlock_.setColour(colour, colour, colour);
   }
 };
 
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):

core/field_colour.js
```diff
@@ -106,11 +106,8 @@ Blockly.FieldColour.prototype.setValue = function(colour) {
   }
   this.colour_ = colour;
   if (this.sourceBlock_) {
-    this.sourceBlock_.setColour(
-      colour,
-      this.sourceBlock_.getColourSecondary(),
-      this.sourceBlock_.getColourTertiary()
-    );
+    // Set the primary, secondary and tertiary colour to this value.
```
rschamp, line 109: Would you explain why here please? E.g. from the PR description
>  the renderer expects to be able to use the secondary color as the fill for a shadow.

rschamp (changes requested):

core/field_colour.js
paulkaplan: @rschamp good idea 50db9134603bef71e130a911a9eb98a882cc4801

PR #1018, opened 2017-07-21
Move non-label fields to min X (fix #983)

### LINKED ISSUE #983: Min spacing not being respected for short blocks with dropdowns
image

Carl mentioned that there should be a minimum length that forces things to come after the notch, per this diagram: 
image

It looks like the relevant variable is `Blockly.BlockSvg.NO_PREVIOUS_INPUT_X_MIN` in `block_render_svg_vertical.js`. It doesn't seem to be used when there are dropdowns, but I'm not familiar enough with the blockly terminology to understand why :( halp.

tmickel:
### Resolves

#983 

### Proposed Changes

Before:
before

After:
after

screen shot 2017-07-21 at 12 15 36 am

### Reason for Changes

paying penance and finishing my unfinished business with @carljbowman 

### Test Coverage

none, but block_render_svg is pretty light in this regard


thisandagain:
\o/ Woo! Thanks @tmickel

PR #1021, opened 2017-07-21
Create data_flyout.js and add lists to the flyout.

marisaleung:
### Resolves
Adding list variables to data section
https://github.com/google/blockly/issues/957

### Proposed Changes

Adds list variables in data section, a button for creating them, and a separate file for the data flyout components
.
### Reason for Changes

We want lists in Scratch 3.0

### DIFF
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -160,7 +160,8 @@ Blockly.Blocks['data_listcontents'] = {
         {
           "type": "field_variable_getter",
           "text": "",
-          "name": "LIST"
+          "name": "LIST",
+          "variableTypes": ["list"]
         }
       ],
       "category": Blockly.Categories.data,
@@ -241,7 +242,8 @@ Blockly.Blocks['data_addtolist'] = {
         },
         {
           "type": "field_variable",
-          "name": "LIST"
+          "name": "LIST",
+          "variableTypes": ["list"]
         }
       ],
       "category": Blockly.Categories.data,
@@ -265,7 +267,8 @@ Blockly.Blocks['data_deleteoflist'] = {
         },
         {
           "type": "field_variable",
-          "name": "LIST"
+          "name": "LIST",
+          "variableTypes": ["list"]
         }
       ],
       "category": Blockly.Categories.data,
@@ -293,7 +296,8 @@ Blockly.Blocks['data_insertatlist'] = {
         },
         {
           "type": "field_variable",
-          "name": "LIST"
+          "name": "LIST",
+          "variableTypes": ["list"]
         }
       ],
       "category": Blockly.Categories.data,
@@ -317,7 +321,8 @@ Blockly.Blocks['data_replaceitemoflist'] = {
         },
         {
           "type": "field_variable",
-          "name": "LIST"
+          "name": "LIST",
+          "variableTypes": ["list"]
         },
         {
           "type": "input_value",
@@ -345,7 +350,8 @@ Blockly.Blocks['data_itemoflist'] = {
         },
         {
           "type": "field_variable",
-          "name": "LIST"
+          "name": "LIST",
+          "variableTypes": ["list"]
         }
       ],
       "output": null,
@@ -367,7 +373,8 @@ Blockly.Blocks['data_lengthoflist'] = {
       "args0": [
         {
           "type": "field_variable",
-          "name": "LIST"
+          "name": "LIST",
+          "variableTypes": ["list"]
         }
       ],
       "category": Blockly.Categories.data,
@@ -387,7 +394,8 @@ Blockly.Blocks['data_listcontainsitem'] = {
       "args0": [
         {
           "type": "field_variable",
-          "name": "LIST"
+          "name": "LIST",
+          "variableTypes": ["list"]
         },
         {
           "type": "input_value",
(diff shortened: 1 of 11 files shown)


marisaleung:
Note: This includes changes changes merged into blockly. I cherry-picked "Add VariableTypes checks". The changes to be reviewing are the ones included in the commit titled, "Adding Lists!"

blocks_vertical/data.js
```diff
@@ -160,7 +160,8 @@ Blockly.Blocks['data_listcontents'] = {
         {
           "type": "field_variable_getter",
           "text": "",
-          "name": "LIST"
+          "name": "VARIABLE",
```
rachel-fenichel: These are all list blocks.  Why are you renaming their fields to VARIABLE instead of LIST?

core/data_flyout.js
```diff
@@ -0,0 +1,379 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Utility functions for handling variables.
```
rachel-fenichel: Make this accurate.

core/data_flyout.js
```diff
@@ -0,0 +1,379 @@
...
+goog.provide('Blockly.DataFlyout');
+
+goog.require('Blockly.Blocks');
+goog.require('Blockly.VariableModel');
+goog.require('Blockly.Variables');
+goog.require('Blockly.Workspace');
+
+/**
+ * Construct the blocks required by the flyout for the variable category.
+ * @param {!Blockly.Workspace} workspace The workspace contianing variables.
```
rachel-fenichel: Fix spelling of "containing"

core/data_flyout.js
```diff
@@ -0,0 +1,379 @@
...
+ * @return {!Array.<!Element>} Array of XML block elements.
+ */
+Blockly.DataFlyout = function(workspace) {
+  var variableModelList = workspace.getVariablesOfType('');
```
rachel-fenichel: 350 line function?  Decompose into multiple function (~1 per top-level block), and figure out if any of them share the same XML other than the name of the block.

core/data_flyout.js
marisaleung: Done

core/data_flyout.js
marisaleung: Done

blocks_vertical/data.js
```diff
@@ -160,7 +160,8 @@ Blockly.Blocks['data_listcontents'] = {
         {
           "type": "field_variable_getter",
           "text": "",
-          "name": "LIST"
+          "name": "LIST",
+          "variableTypes": ['list']
```
rachel-fenichel: Double quotes for JSON.

marisaleung: Done

core/data_category.js
```diff
@@ -0,0 +1,429 @@
...
+
+  // Now add list variables to the flyout
+  Blockly.DataCategory.addCreateButton(xmlList, workspace, 'LIST');
+  variableModelList = workspace.getVariablesOfType('list');
+  variableModelList.sort(Blockly.VariableModel.compareByName);
+  for (var i = 0; i < variableModelList.length; i++) {
+    Blockly.DataCategory.addDataVariable(xmlList, variableModelList[i]);
+  }
+
+  if (variableModelList.length > 0) { // The button is always there.
```
rachel-fenichel: // the comment isn't valid

core/data_category.js
```diff
@@ -0,0 +1,429 @@
...
+  variableModelList.sort(Blockly.VariableModel.compareByName);
+  var xmlList = [];
+
+  Blockly.DataCategory.addCreateButton(xmlList, workspace, 'VARIABLE');
+
+  for (var i = 0; i < variableModelList.length; i++) {
+    Blockly.DataCategory.addDataVariable(xmlList, variableModelList[i]);
+  }
+
+  if (xmlList.length > 1) { // The button is always there.
```
rachel-fenichel: Check variableModelList.length here, the way you do below, and update the comments to say // Only show these blocks if there is at least one variable [list] defined.

rachel-fenichel (commented):
Nits then lgtm.

core/data_category.js
marisaleung: Good catch. Done.

core/data_category.js
marisaleung: Done.

paulkaplan:
@thisandagain the code here looks good, but it will not work correctly in the GUI until we update the VM to handle multiple variable types. Maybe we should merge the code to make this work but suppress it in the UI until we do the rest of list support in a few months. Thoughts on that?

thisandagain:
@paulkaplan Landing and suppressing in the UI until we have time to wrap up the other side of this work sounds good to me. I would hate for this PR to rot in the interim.

paulkaplan:
Ok, then I'll merge this and follow it up with one of my own disabling just the `Create list...` button.

PR #1025, opened 2017-07-25
Update vertical block icons to spec, fixes #526

### LINKED ISSUE #526: Final block icon assets
Currently in the vertical implementation we have the "forever" icon on loop C-blocks:
screen shot 2016-07-14 at 3 30 29 pm

The motion blocks that traditionally have icons are iconless:
screen shot 2016-07-14 at 3 31 11 pm

If we want icons here or in other places, need to figure them out.

tmickel:
### Resolves

#526

### Proposed Changes

Updates the block icons to be to spec (green flag, turn left/right, C-block arrows).

Before:
before

After:
after

### Reason for Changes

💯bringing to spec

### Test Coverage

not needed I think :)


thisandagain:
🎉  This looks great. Thanks @tmickel !

PR #1028, opened 2017-07-26
Add shadow dom ID uniquifier to 'Duplicate' code path

### LINKED ISSUE #1024: Duplicate ID problem still exists when using "Duplicate" action
A while ago we dealt with this issue https://github.com/LLK/scratch-blocks/issues/926 which we solved by manually reassigning IDs when blocks were pasted. 

The same problem still exists when you right-click => Duplicate, because that takes a different code path from the "paste" code path which was fixed by https://github.com/LLK/scratch-blocks/pull/936.

We will need to implement a similar fix for duplicating.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1024, a recurring and high priority problem that occurs when you use the "Duplicate" menu option. 

### Proposed Changes

Like the previous version of this bug, fix it by giving unique IDs to all shadows. 

### Reason for Changes

See linked issue for more background. TL;DR duplicate IDs, even for obscured shadows, breaks scratch-vm's expectations of having unique IDs.

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -686,20 +686,7 @@ Blockly.BlockSvg.prototype.duplicateAndDragCallback_ = function() {
       var newBlock = Blockly.Xml.domToBlock(xml, ws);
 
       // Scratch-specific: Give shadow dom new IDs to prevent duplicating on paste
-      var blocks = newBlock.getDescendants();
-      for (var i = blocks.length - 1; i >= 0; i--) {
-        var descendant = blocks[i];
-        for (var j = 0; j < descendant.inputList.length; j++) {
-          var connection = descendant.inputList[j].connection;
-          if (connection) {
-            var shadowDom = connection.getShadowDom();
-            if (shadowDom) {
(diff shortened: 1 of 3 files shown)


core/block_svg.js
```diff
@@ -684,6 +684,23 @@ Blockly.BlockSvg.prototype.duplicateAndDragCallback_ = function() {
       // Using domToBlock instead of domToWorkspace means that the new block
       // will be placed at position (0, 0) in main workspace units.
       var newBlock = Blockly.Xml.domToBlock(xml, ws);
+
+      // Scratch-specific: Give shadow dom new IDs to prevent duplicating on paste
```
rachel-fenichel, line 688: Make this a helper function and call it from here and the copy paste case.  Both because it's good form and because this function is already hard to understand from the coordinate work.

paulkaplan:
@rachel-fenichel fixed in 3efae4c

rachel-fenichel (approved):

PR #1030, opened 2017-07-27
Fix traversal for tab ordering

### LINKED ISSUE #170: Tab order through blocks
This feels a bit backwards to me - blocks tab through right-to-left and in-to-out. What was the logic here?

@NeilFraser

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/170

### Proposed Changes

Change the tab order traversal. Instead of always going to the parent (either output or previous connection), only go "up" to output if it exists. If there is no output, use next / previous connection as directed. 

This also fixes a bug that occurs when tabbing (and esc/entering) into fields that have a dropdown div open, namely the angle picker. 

### Reason for Changes

Because it was wrong!

tabbing

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -624,16 +624,9 @@ Blockly.BlockSvg.prototype.tab = function(start, forward) {
     if (outputBlock) {
       outputBlock.tab(this, forward);
     } else { // Otherwise, go to next / previous block, depending on value of `forward`
-      if (forward) {
-        var nextBlock = this.nextConnection && this.nextConnection.targetBlock();
-        if (nextBlock) {
-          nextBlock.tab(this, forward);
-        }
-      } else {
-        var previousBlock = this.previousConnection && this.previousConnection.targetBlock();
-        if (previousBlock) {
-          previousBlock.tab(this, forward);
-        }
+      var block = forward ? this.getNextBlock() : this.getPreviousBlock();
+      if (block) {
+        block.tab(this, forward);
       }
     }
   } else if (target instanceof Blockly.Field) {
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -619,9 +619,22 @@ Blockly.BlockSvg.prototype.tab = function(start, forward) {
   var target = list[forward ? i + 1 : i - 1];
   if (!target) {
     // Ran off of list.
-    var parent = this.getParent();
-    if (parent) {
-      parent.tab(this, forward);
+    // If there is an output, tab up to that block.
+    var outputBlock = this.outputConnection && this.outputConnection.targetBlock();
+    if (outputBlock) {
+      outputBlock.tab(this, forward);
+    } else { // Otherwise, go to next / previous block, depending on value of `forward`
+      if (forward) {
+        var nextBlock = this.nextConnection && this.nextConnection.targetBlock();
+        if (nextBlock) {
+          nextBlock.tab(this, forward);
+        }
+      } else {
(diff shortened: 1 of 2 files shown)


core/block_svg.js
```diff
@@ -619,9 +619,22 @@ Blockly.BlockSvg.prototype.tab = function(start, forward) {
   var target = list[forward ? i + 1 : i - 1];
```
rachel-fenichel, line 619: This tab function broadly needs cleanup, but I've recorded that in Blockly: https://github.com/google/blockly/issues/1249

core/block_svg.js
```diff
@@ -619,9 +619,22 @@ Blockly.BlockSvg.prototype.tab = function(start, forward) {
...
   if (!target) {
     // Ran off of list.
-    var parent = this.getParent();
-    if (parent) {
-      parent.tab(this, forward);
+    // If there is an output, tab up to that block.
+    var outputBlock = this.outputConnection && this.outputConnection.targetBlock();
+    if (outputBlock) {
```
rachel-fenichel, line 624: Why does this not depend on the value of `forward`?

core/block_svg.js
```diff
@@ -619,9 +619,22 @@ Blockly.BlockSvg.prototype.tab = function(start, forward) {
...
+      outputBlock.tab(this, forward);
+    } else { // Otherwise, go to next / previous block, depending on value of `forward`
+      if (forward) {
```
rachel-fenichel: This can be
```
var block = forward ? this.getNextBlock() : this.getPreviousBlock();
if (block) {
  block.tab(this, forward);
}
```

core/block_svg.js
```diff
@@ -619,9 +619,22 @@ Blockly.BlockSvg.prototype.tab = function(start, forward) {
...
+        var nextBlock = this.nextConnection && this.nextConnection.targetBlock();
+        if (nextBlock) {
+          nextBlock.tab(this, forward);
+        }
+      } else {
+        var previousBlock = this.previousConnection && this.previousConnection.targetBlock();
+        if (previousBlock) {
+          previousBlock.tab(this, forward);
+        }
+      }
```
rachel-fenichel: What happens when you get to the end of a stack of blocks inside an if/else?

core/block_svg.js
paulkaplan: I believe because you've basically "bottomed out", and the only way to go is back "up" to the output connection.

core/block_svg.js
paulkaplan: :( not the right thing :(

paulkaplan: At that point, I guess what we have to do is traverse back up the previous connections until we find an input connection, then go to the next connection of that block?

RoboErikG:
I think what you really want for this change is a full depth first traversal, with it popping back up to the parent when it reaches the end. For both directions you'll also have to consider the next connection as part of the list to iterate through so that when you reach the end of a stack in a loop it can recursively pop up the stack until it gets to the input and can move to the next item in the list.

core/block_svg.js
rachel-fenichel: Yeah that's what I figured.

I'm totally down with you merging this as-is, because it's an improvement over the existing code for any short-term uses of tabbing.  But it doesn't fix the problem all the way.  See what Erik said below.

There's probably a clever way to do it without using any external state, which would be preferable, but fixing it fully isn't a high priority for me right now.

paulkaplan: Sounds good. I'll merge this, and we can return to @RoboErikG if/when it comes up again.

PR #1034, opened 2017-08-01
Correctly apply flipRTL for images in vertical RTL

tmickel:
### Resolves
--
### Proposed Changes

Before (see arrows, GF):
before
After:
after

### Reason for Changes

Image fields that have the "flipRTL" property on them are intended to be flipped in RTL mode. This works for the horizontal blocks rendering, but was never implemented for vertical rendering. With this change, it is.

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -633,21 +633,32 @@ Blockly.BlockSvg.prototype.renderFields_ =
     // Offset the field upward by half its height.
     // This vertically centers the fields around cursorY.
     var yOffset = -field.getSize().height / 2;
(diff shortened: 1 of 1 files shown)


thisandagain:
Thanks @tmickel. This LGTM but I'd like @rachel-fenichel to take a second look.

rachel-fenichel (approved):

paulkaplan:
So code for this looks good, but I'm going to follow it up with a PR to disable `flip_rtl` on the green flag after talking with @carljbowman, since it should not actually be flipped. The remaining `field_image`'s that should will be flipped are only the repeat loop arrow at the bottom of repeat blocks.

PR #1035, opened 2017-08-01
Center variable reporter; fix field centering to work for it

tmickel:
### Resolves
--
### Proposed Changes

Before:
before_center

After:
after_center
screen shot 2017-08-01 at 12 19 08 am
screen shot 2017-08-01 at 12 19 30 am

### Reason for Changes

Variables named with "text with a small width" (e.g., a single character) were not aligned in the center of the reporter. By adding the CENTRE alignment (and fixing its implementation), they are.

### Test Coverage
--

### DIFF
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -37,6 +37,7 @@ Blockly.Blocks['data_variable'] = {
   init: function() {
     this.jsonInit({
       "message0": "%1",
+      "lastDummyAlign0": "CENTRE",
       "args0": [
         {
           "type": "field_variable_getter",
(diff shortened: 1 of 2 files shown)


mrjacobbloom:
aaah emoji support

thisandagain (approved):

thisandagain:
🎉 🐱 🌮 

@carljbowman This looks good functionally. Do you want to give it a look before landing?

carljbowman (approved):
@tmickel - Yes!!! This has been driving me nuts for a while. Haha also a bit dangerous that I now know I can email you requests. 

Thanks for fixing!

rachel-fenichel:
So should all single-field blocks have center align for their field?

Also gets back to the conversation we were having about alignment.

PR #1051, opened 2017-08-22
Exclude image field from positioning fix

ericrosenbaum:
### Resolves

For extensions, we'll be adding an icon to the start of each block. Currently those icons are positioned too far to the right, due to an existing positioning fix related to the notch in the block:

screen shot 2017-08-22 at 11 51 49 am

### Proposed Changes

Correct the positioning of leading icons by excluding image fields from the notch fix:

screen shot 2017-08-22 at 11 48 08 am

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -623,8 +623,9 @@ Blockly.BlockSvg.prototype.renderFields_ =
     if (!root) {
       continue;
     }
-    // In blocks with a notch, non-label fields should be bumped to a min X,
-    // to avoid overlapping with the notch.
+    // In blocks with a notch, fields should be bumped to a min X,
+    // to avoid overlapping with the notch. Label and image fields are
+    // excluded.
     if (this.previousConnection && !(field instanceof Blockly.FieldLabel) &&
         !(field instanceof Blockly.FieldImage)) {
       cursorX = this.RTL ?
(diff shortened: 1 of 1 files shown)


core/block_render_svg_vertical.js
```diff
@@ -625,7 +625,7 @@ Blockly.BlockSvg.prototype.renderFields_ =
     }
     // In blocks with a notch, non-label fields should be bumped to a min X,
     // to avoid overlapping with the notch.
-    if (this.previousConnection && !(field instanceof Blockly.FieldLabel)) {
+    if (this.previousConnection && !(field instanceof Blockly.FieldLabel) && !(field instanceof Blockly.FieldImage)) {
```
rachel-fenichel: Nit: line is too long.  Please break:

```
if (this.previousConnection && !(field instanceof Blockly.FieldLabel) && 
    !(field instanceof Blockly.FieldImage)) {
```

core/block_render_svg_vertical.js
```diff
@@ -625,7 +625,7 @@ Blockly.BlockSvg.prototype.renderFields_ =
     }
     // In blocks with a notch, non-label fields should be bumped to a min X,
```
rachel-fenichel: Update the comment to say that images are also excluded.

rachel-fenichel:
Two nits and then it looks good.  To be clear, this is applied to all images, not just icons for extensions.  Is that the intended behaviour?

ericrosenbaum:
Thanks for the feedback @rachel-fenichel! You're right, this will apply to all images, and that's the intended behavior.

Issue #1039, opened 2017-08-03
Vertical category menu


ericrosenbaum:
We are shifting to a new design for the blocks palette, with a vertical category menu.

Steps for implementing:

- [x] Put all the categories into a single column (like [this](https://github.com/ericrosenbaum/scratch-blocks/commit/7a5b9467a7a31441a4b8138fe7bba11b474bcf43)) 
- [x] Styling (like [this](https://github.com/ericrosenbaum/scratch-blocks/commit/258a20d59cd47e06377e1a553ceff2ac2ad2c8f5)).  
- [x] Fix the resulting problem with the delete area. Workspace_svg appears to have a bug. In recordDeleteAreas, it checks for a flyout using this.flyout_, which is undefined. We fixed this by instead using this.getFlyout(), and using this.getFlyout().getClientRect(). (like [this](https://github.com/ericrosenbaum/scratch-blocks/commit/bad48f06c608820b2434bc3199dba4fda44b1bcf)).

---

Questions:
- [x] Should we maintain support for the stacked category menu and add two column as an option, or just replace stacked with two column?

PR #1052, opened 2017-08-22
Vertical category menu

ericrosenbaum:
### Resolves

Closes https://github.com/LLK/scratch-blocks/issues/1039

### Proposed Changes

Switch to a vertical layout for the blocks category menu. We replaced the stacked category layout because we do not plan on supporting it any more.

One area we got tripped up on is how to specify the different widths involved. We took out the line that had the flyouts inherit their width from the toolbox, instead always having them use their `DEFAULT_WIDTH`. @rachel-fenichel do you have any thoughts about the way this was done?


rachel-fenichel:
Since this is a decently large in functionality (though not in code), here are the standard questions:

- [x] Does the horizontal toolbox work?
- [x] Does the simple horizontal flyout work?
- [x] Does the vertical toolbox work?
- [x] Does the simple vertical flyout work?
- [x] Are the delete areas correct?
- [x] Does it work when you zoom far in?
- [x] Does it work when you zoom far out?

Have you tested 

- [x] In LTR mode
- [x] In RTL mode
- [x] In the multi-playground
- [x] On Android
- [x] On iOS
- [x] In Safari
- [x] In Edge
- [x] In IE11
- [x] In Firefox
- [x] In Chrome

TheBrokenRail:
Will there be an option to use the "classic" scratch toolbox?

ericrosenbaum:
Thanks for keeping us on our toes @rachel-fenichel. The current state on the develop branch of our horizontal toolbox is pretty broken, as are all the top and bottom layouts for vertical and horizontal, and this PR doesn't change that. Delete areas, zooming, and RTL are working fine and not broken by the PR. AFAIK we're not doing full cross-browser testing until after our alpha release, but chrome, safari and firefox look fine.

rachel-fenichel:
@ericrosenbaum the horizontal toolbox is currently in use (cozmo) so at least one horizontal toolbox configuration currently works.  I'm worried about breaking that.

Not doing full cross-browser testing: I don't want to wait until December to find out that a change in August broke touch on iOS, because then there's a lot more work to fix it.  Please do a cursory test on iOS, Android, IE11, and Edge.  

If it's broken it doesn't need to block this PR, but you should file a bug so we know that this change broke it.

What about the simple vertical toolbox?  (Flyout only.)

ericrosenbaum:
Sounds good, we're doing some of that testing now. 

A question from @paulkaplan  that came up when testing touch devices: because of `Blockly.bindEvent` (which calls `preventDefault` on the event) on the categories, one cannot scroll a long category list. Do you know any way of getting around that?

Aside from that things are looking fine. We noticed a bunch of layout bugs in Edge and IE11, but those are unrelated to this PR and I will file them separately.

paulkaplan:
Added a commit to allow a horizontal layout of the category menu also.

image

ericrosenbaum:
@TheBrokenRail sorry, we don't currently plan to keep supporting the "classic" toolbox style.

msintov:
@ericrosenbaum If I set the flyout to autoclose in flyout_vertical.js, I find that once a block is dragged out and the flyout autocloses, the area where the flyout used to be is still an active delete area. Also, the workspace is only to the right of the flyout would be, whether the flyout is opened or closed. Is this behavior unintentional? Thanks in advance.

ericrosenbaum:
Sorry @msintov- that's unintentional! We neglected to test with autoclose. Sounds like we'll need to fix the workspace positioning and the delete area. Would you please submit an issue? Thanks!

msintov:
@ericrosenbaum Thanks for the note. Logged it here: https://github.com/LLK/scratch-blocks/issues/1067

PR #1055, opened 2017-08-25
Add pen transparency blocks

Kenny2github:
### Resolves

Resolves LLK/scratch-gui#602

### Proposed Changes

Adds the two new pen transparency blocks

Comes with LLK/scratch-vm#669


Kenny2github:
@ericrosenbaum if I understand this correctly you merge VM pulls before scratch-blocks pulls for the same block, yes?

cwillisf (approved):
Thanks, @Kenny2github!

cwillisf:
By the way, you're correct: when a feature crosses multiple repositories, we typically try to merge (for example) the `scratch-vm` part before the `scratch-gui` part. That's because, technically speaking, `scratch-gui` depends on `scratch-vm` rather than the other way around. You can tell which repositories depend on each other by checking the `dependencies` and `devDependencies` sections in the `package.json` file in the root of the repository.

Kenny2github:
@cwillisf Yeah that’s what I thought - it wouldn’t make sense to merge something that didn’t have its dependency merged yet.

Alright, thanks! Merge when you will.

Issue #1040, opened 2017-08-03
Blocks menu with all the blocks in one flyout


ericrosenbaum:
The new design for the blocks palette includes a single vertical scrolling flyout containing the blocks from all the categories, with labels and separators. This design works well with the vertical category menu (https://github.com/LLK/scratch-blocks/issues/1039).

Here's a [prototype](https://ericrosenbaum.github.io/scratch-gui/extension-categories/) version that includes both the vertical category menu and flyout containing all the blocks. And here's a gif:
 
![vert-toolbox-demo](https://user-images.githubusercontent.com/567844/28943848-4a8d3db6-786e-11e7-8f82-c61abd9585f0.gif) 

Implementation notes:

- [x] **Concatenate all the blocks** (like [this](https://github.com/ericrosenbaum/scratch-blocks/blob/78cd82f033429999b7617cc95049d0f6df27359a/core/toolbox.js#L191))

- [x] When you select a category by clicking the button, **auto-scroll** to the category, quickly but with easing so it slows down as it arrives. Note that the easing requires a magic number, which is the fraction of the remaining distance to cover at each animation step (0.3 seems pretty good). Also, the target position for the animation is cleared by scrollbar move events, so that you can interrupt the auto-scrolling. Commit [here](https://github.com/ericrosenbaum/scratch-blocks/commit/91bff9a5eb2f326430e33aae0faa5ea1bb437ba9).

- [x] **Update the selected item** in the category menu as the flyout scrolls into a new category. I implemented this by storing an array of objects with category position and category name fields, in flyout.show(). At each scroll event, I walk backwards down this array (i.e. starting with the bottom-most category) and check if the scroll position is below it, then set the selection state using the category name. Commit [here](https://github.com/ericrosenbaum/scratch-blocks/commit/a516a7e278eec9c418c9b9d61cd18eff30f475cf). Perhaps there's a better way to do this? 

[...]

PR #1059, opened 2017-08-28
Single auto-scrolling flyout containing all of the blocks

ericrosenbaum:
### Resolves

Resolves part of https://github.com/LLK/scratch-blocks/issues/1040

### Proposed Changes

Concatenate all the blocks into a single flyout, with category labels, so that you can scroll continuously through them. When you select a category in the category menu, auto-scroll to that category.

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -534,19 +534,19 @@ Blockly.Flyout.prototype.show = function(xmlList) {
 };
 
 /**
- * Store an array of with category names and scrollbar positions.
+ * Store an array of category names and scrollbar positions.
  * This is used when scrolling the flyout to cause a category to be selected.
  * @private
  */
 Blockly.Flyout.prototype.recordCategoryScrollPositions_ = function() {
   this.categoryScrollPositions = [];
   for (var i = 0; i < this.buttons_.length; i++) {
-    if (this.buttons_[i].isCategoryLabel_) {
+    if (this.buttons_[i].getIsCategoryLabel()) {
       var categoryLabel = this.buttons_[i];
       this.categoryScrollPositions.push({
-        categoryName: categoryLabel.text_,
+        categoryName: categoryLabel.getText(),
         position: this.horizontalLayout_ ?
-          categoryLabel.position_.x : categoryLabel.position_.y
+          categoryLabel.getPosition().x : categoryLabel.getPosition().y
       });
     }
   }
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -547,6 +547,7 @@ Blockly.Flyout.prototype.recordCategoryScrollPositions_ = function() {
 /**
  * Step the scrolling animation by scrolling a fraction of the way to
  * a scroll target, and request the next frame if necessary.
+ * @package
  */
 Blockly.Flyout.prototype.stepScrollAnimation = function() {
   if (!this.scrollTarget) {
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -124,7 +124,7 @@ Blockly.Flyout = function(workspaceOptions) {
  * Does the flyout automatically close when a block is created?
  * @type {boolean}
  */
-Blockly.Flyout.prototype.autoClose = true;
+Blockly.Flyout.prototype.autoClose = false;
 
 /**
  * Whether the flyout is visible.
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -520,6 +520,19 @@ Blockly.Flyout.prototype.show = function(xmlList) {
 
   this.reflowWrapper_ = this.reflow.bind(this);
   this.workspace_.addChangeListener(this.reflowWrapper_);
+
+  // make a map of scrollbar positions to category names
+  // so that scrolling the flyout can cause a category to be selected
+  this.categoryScrollPositions = [];
+  for (var i=0; i<this.buttons_.length; i++) {
+    var category = this.buttons_[i];
+    if (category.isLabel_) {
+      this.categoryScrollPositions.push({
+        categoryName:category.text_,
+        position:category.position_.y
+      });
+    }
+  }
 };
 
 /**
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -456,8 +456,8 @@ Blockly.Flyout.prototype.show = function(xmlList) {
       var newList = fnToApply(this.workspace_.targetWorkspace);
       // Insert the new list of blocks in the middle of the list.
       // We use splice to insert at index i, and remove a single element
-      // (the placeholder string). Because the spread operator (...) is not available,
-      // use apply and concat the array.
+      // (the placeholder string). Because the spread operator (...) is not
+      // available, use apply and concat the array.
       xmlList.splice.apply(xmlList, [i, 1].concat(newList));
       xml = xmlList[i];
     }
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -532,7 +532,7 @@ Blockly.Flyout.prototype.show = function(xmlList) {
  */
 Blockly.Flyout.prototype.recordCategoryScrollPositions_ = function() {
   this.categoryScrollPositions = [];
-  for (var i=0; i<this.buttons_.length; i++) {
+  for (var i = 0; i < this.buttons_.length; i++) {
     if (this.buttons_[i].isCategoryLabel_) {
       var categoryLabel = this.buttons_[i];
       this.categoryScrollPositions.push({
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -553,6 +553,23 @@ Blockly.Flyout.prototype.recordCategoryScrollPositions_ = function() {
 };
 
 /**
+ * Select a category using the scroll position.
+ * @param  {number} pos The scroll position.
+ * @package
+ */
+Blockly.Flyout.prototype.selectCategoryByScrollPosition = function(pos) {
+  var scaledPos = pos / this.workspace_.scale;
+  // Traverse the array of scroll positions in reverse, so we can select the furthest
+  // category that the scroll position is beyond
+  for (var i = this.categoryScrollPositions.length - 1; i >= 0; i--) {
+    if (scaledPos > this.categoryScrollPositions[i].position) {
+      this.parentToolbox_.selectCategoryByName(this.categoryScrollPositions[i].categoryName);
+      return;
+    }
+  }
+};
+
+/**
  * Step the scrolling animation by scrolling a fraction of the way to
  * a scroll target, and request the next frame if necessary.
  * @package
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -447,16 +447,17 @@ Blockly.Flyout.prototype.show = function(xmlList) {
   var gaps = [];
   this.permanentlyDisabled_.length = 0;
   for (var i = 0, xml; xml = xmlList[i]; i++) {
-    // Handle the dynamic category for variables, represented by a name instead of a list of XML.
+    // Handle dynamic categories, represented by a name instead of a list of XML.
     // Look up the correct category generation function and call that to get a
     // valid XML list.
     if (typeof xmlList[i] === 'string') {
       var fnToApply = this.workspace_.targetWorkspace.getToolboxCategoryCallback(
           xmlList[i]);
       var newList = fnToApply(this.workspace_.targetWorkspace);
-      // insert the new list of variable blocks in the middle of the list
-      // we use splice to insert at index i, and remove a single element (the placeholder string)
-      // because the spread operator (...) is not available, use apply and concat the array
+      // Insert the new list of blocks in the middle of the list.
+      // We use splice to insert at index i, and remove a single element
+      // (the placeholder string). Because the spread operator (...) is not available,
+      // use apply and concat the array.
       xmlList.splice.apply(xmlList, [i, 1].concat(newList));
       xml = xmlList[i];
     }
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -441,25 +441,25 @@ Blockly.Flyout.prototype.show = function(xmlList) {
   this.hide();
   this.clearOldBlocks_();
 
-  // Handle dynamic categories, represented by a name instead of a list of XML.
-  // Look up the correct category generation function and call that to get a
-  // valid XML list.
-  if (typeof xmlList == 'string') {
-    var fnToApply = this.workspace_.targetWorkspace.getToolboxCategoryCallback(
-        xmlList);
-    goog.asserts.assert(goog.isFunction(fnToApply),
-        'Couldn\'t find a callback function when opening a toolbox category.');
-    xmlList = fnToApply(this.workspace_.targetWorkspace);
-    goog.asserts.assert(goog.isArray(xmlList),
-        'The result of a toolbox category callback must be an array.');
-  }
-
   this.setVisible(true);
   // Create the blocks to be shown in this flyout.
   var contents = [];
   var gaps = [];
   this.permanentlyDisabled_.length = 0;
   for (var i = 0, xml; xml = xmlList[i]; i++) {
+    // Handle the dynamic category for variables, represented by a name instead of a list of XML.
+    // Look up the correct category generation function and call that to get a
+    // valid XML list.
+    if (typeof xmlList[i] === 'string') {
+      var fnToApply = this.workspace_.targetWorkspace.getToolboxCategoryCallback(
+          xmlList[i]);
+      var newList = fnToApply(this.workspace_.targetWorkspace);
+      // insert the new list of variable blocks in the middle of the list
+      // we use splice to insert at index i, and remove a single element (the placeholder string)
+      // because the spread operator (...) is not available, use apply and concat the array
+      xmlList.splice.apply(xmlList, [i, 1].concat(newList));
+      xml = xmlList[i];
+    }
     if (xml.tagName) {
       var tagName = xml.tagName.toUpperCase();
       var default_gap = this.horizontalLayout_ ? this.GAP_X : this.GAP_Y;
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -522,15 +522,23 @@ Blockly.Flyout.prototype.show = function(xmlList) {
   this.reflowWrapper_ = this.reflow.bind(this);
   this.workspace_.addChangeListener(this.reflowWrapper_);
 
-  // make a map of scrollbar positions to category names
-  // so that scrolling the flyout can cause a category to be selected
+  this.recordCategoryScrollPositions_();
+};
+
+/**
+ * Store an array of with category names and scrollbar positions.
+ * This is used when scrolling the flyout to cause a category to be selected.
+ * @private
+ */
+Blockly.Flyout.prototype.recordCategoryScrollPositions_ = function() {
   this.categoryScrollPositions = [];
   for (var i=0; i<this.buttons_.length; i++) {
-    var category = this.buttons_[i];
-    if (category.isLabel_) {
+    if (this.buttons_[i].isCategoryLabel_) {
+      var categoryLabel = this.buttons_[i];
       this.categoryScrollPositions.push({
-        categoryName:category.text_,
-        position:this.horizontalLayout_ ? category.position_.x : category.position_.y
+        categoryName: categoryLabel.text_,
+        position: this.horizontalLayout_ ?
+          categoryLabel.position_.x : categoryLabel.position_.y
       });
     }
   }
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -529,7 +529,7 @@ Blockly.Flyout.prototype.show = function(xmlList) {
     if (category.isLabel_) {
       this.categoryScrollPositions.push({
         categoryName:category.text_,
-        position:category.position_.y
+        position:this.horizontalLayout_ ? category.position_.x : category.position_.y
       });
     }
   }
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -229,6 +229,14 @@ Blockly.Flyout.prototype.verticalOffset_ = 0;
 Blockly.Flyout.prototype.dragAngleRange_ = 70;
 
 /**
+ * The fraction of the distance to the scroll target to move the flyout on
+ * each animation frame, when auto-scrolling. Values closer to 1.0 will make
+ * the scroll animation complete faster. Use 1.0 for no animation.
+ * @type {number}
+ */
+Blockly.Flyout.prototype.scrollAnimationFraction = 0.3;
+
+/**
  * Creates the flyout's DOM.  Only needs to be called once. The flyout can
  * either exist as its own svg element or be a g element nested inside a
  * separate svg element.
@@ -560,7 +568,7 @@ Blockly.Flyout.prototype.stepScrollAnimation = function() {
     this.scrollbar_.set(this.scrollTarget);
     return;
   }
-  this.scrollbar_.set(scrollPos + diff * 0.3);
+  this.scrollbar_.set(scrollPos + diff * this.scrollAnimationFraction);
 
   // Polyfilled by goog.dom.animationFrame.polyfill
   requestAnimationFrame(this.stepScrollAnimation.bind(this));
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -118,6 +118,14 @@ Blockly.Flyout = function(workspaceOptions) {
    * @private
    */
   this.parentToolbox_ = null;
+
+  /**
+   * The target position for the flyout scroll animation in pixels.
+   * Is a number while animating, null otherwise.
+   * @type {?number}
+   * @package
+   */
+  this.scrollTarget = null;
 };
 
 /**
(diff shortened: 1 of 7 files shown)


core/flyout_base.js
```diff
@@ -441,25 +441,25 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
-        'The result of a toolbox category callback must be an array.');
-  }
-
   this.setVisible(true);
   // Create the blocks to be shown in this flyout.
   var contents = [];
   var gaps = [];
   this.permanentlyDisabled_.length = 0;
   for (var i = 0, xml; xml = xmlList[i]; i++) {
+    // Handle the dynamic category for variables, represented by a name instead of a list of XML.
```
rachel-fenichel: Variables are not the only possible dynamic category now.  Change the comment back to "Handle dynamic categories."

core/flyout_base.js
```diff
@@ -441,25 +441,25 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
+    // Look up the correct category generation function and call that to get a
+    // valid XML list.
+    if (typeof xmlList[i] === 'string') {
+      var fnToApply = this.workspace_.targetWorkspace.getToolboxCategoryCallback(
+          xmlList[i]);
+      var newList = fnToApply(this.workspace_.targetWorkspace);
+      // insert the new list of variable blocks in the middle of the list
```
rachel-fenichel: Capitalize the beginnings of sentences and end with periods in comments.

core/flyout_base.js
```diff
@@ -441,25 +441,25 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
+      // we use splice to insert at index i, and remove a single element (the placeholder string)
```
rachel-fenichel: Line wrap at 80 (eyeballing, I think this exceeds 80).

core/flyout_base.js
```diff
@@ -520,6 +520,19 @@ Blockly.Flyout.prototype.show = function(xmlList) {
 
   this.reflowWrapper_ = this.reflow.bind(this);
   this.workspace_.addChangeListener(this.reflowWrapper_);
+
+  // make a map of scrollbar positions to category names
+  // so that scrolling the flyout can cause a category to be selected
+  this.categoryScrollPositions = [];
+  for (var i=0; i<this.buttons_.length; i++) {
```
rachel-fenichel: Spaces around `=` and `<`

core/flyout_base.js
```diff
@@ -520,6 +520,19 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
+    var category = this.buttons_[i];
+    if (category.isLabel_) {
+      this.categoryScrollPositions.push({
+        categoryName:category.text_,
```
rachel-fenichel: Space after colon, here and on the next line.

core/flyout_base.js
```diff
@@ -520,6 +520,19 @@ Blockly.Flyout.prototype.show = function(xmlList) {
 
   this.reflowWrapper_ = this.reflow.bind(this);
   this.workspace_.addChangeListener(this.reflowWrapper_);
+
+  // make a map of scrollbar positions to category names
+  // so that scrolling the flyout can cause a category to be selected
+  this.categoryScrollPositions = [];
+  for (var i=0; i<this.buttons_.length; i++) {
+    var category = this.buttons_[i];
+    if (category.isLabel_) {
```
rachel-fenichel: You're assuming that all labels are category labels.  That's not necessarily true.

core/flyout_base.js
```diff
@@ -520,6 +520,19 @@ Blockly.Flyout.prototype.show = function(xmlList) {
 
   this.reflowWrapper_ = this.reflow.bind(this);
   this.workspace_.addChangeListener(this.reflowWrapper_);
+
+  // make a map of scrollbar positions to category names
+  // so that scrolling the flyout can cause a category to be selected
+  this.categoryScrollPositions = [];
```
rachel-fenichel: Make all of this added code into a new function: `this.initCategoryScrollPositions_()`

core/flyout_button.js
```diff
@@ -155,13 +155,14 @@ Blockly.FlyoutButton.prototype.createDom = function() {
       this.svgGroup_);
   svgText.textContent = this.text_;
 
-  this.width = svgText.getComputedTextLength() +
-      2 * Blockly.FlyoutButton.MARGIN;
+  this.width = svgText.getComputedTextLength();
```
rachel-fenichel, line 165: Hm, this looks like a change that should go upstream as well.

core/flyout_horizontal.js
```diff
@@ -245,16 +245,60 @@ Blockly.HorizontalFlyout.prototype.scrollToStart = function() {
 };
 
 /**
+ * Scroll the flyout to a position.
+ * @param {Number} pos The targeted scroll position.
```
rachel-fenichel: Use `number`, not `Number`.

core/flyout_horizontal.js
```diff
@@ -245,16 +245,60 @@ Blockly.HorizontalFlyout.prototype.scrollToStart = function() {
...
+ */
+Blockly.HorizontalFlyout.prototype.scrollTo = function(pos) {
+  this.scrollTarget = pos * this.workspace_.scale;
+
+  // Make sure not to set the scroll target below the lowest point we can
+  // scroll to, i.e. the content height minus the view height
+  var contentWidth = this.workspace_.getMetrics().contentWidth;
```
rachel-fenichel: `getMetrics()` can be tremendously expensive.  Always cache the result if you want to use it twice.

core/flyout_horizontal.js
```diff
@@ -245,16 +245,60 @@ Blockly.HorizontalFlyout.prototype.scrollToStart = function() {
 };
 
 /**
+ * Scroll the flyout to a position.
+ * @param {Number} pos The targeted scroll position.
+ */
+Blockly.HorizontalFlyout.prototype.scrollTo = function(pos) {
+  this.scrollTarget = pos * this.workspace_.scale;
+
+  // Make sure not to set the scroll target below the lowest point we can
```
rachel-fenichel: Update this comment for horizontal instead of vertical (...past the farthest point we can scroll to, i.e. the content width minus the view width)

core/flyout_horizontal.js
```diff
@@ -245,16 +245,60 @@ Blockly.HorizontalFlyout.prototype.scrollToStart = function() {
...
+  this.scrollTarget = Math.min(this.scrollTarget, contentWidth - viewWidth);
+
+  this.step();
+};
+
+/**
+ * Step the scrolling animation by scrolling a fraction of the way to
+ * a scroll target, and request the next frame if necessary.
+ */
+Blockly.HorizontalFlyout.prototype.step = function() {
```
rachel-fenichel: This looks like a private function.  Annotate it as such, and add an underscore to the end of the name.

core/flyout_horizontal.js
```diff
@@ -245,16 +245,60 @@ Blockly.HorizontalFlyout.prototype.scrollToStart = function() {
...
+  if (!this.scrollTarget) return;
```
rachel-fenichel: Put the return on a new line and add braces.

core/flyout_horizontal.js
```diff
@@ -245,16 +245,60 @@ Blockly.HorizontalFlyout.prototype.scrollToStart = function() {
...
+  var scrollXPos = -this.workspace_.scrollX;
+  var diff = this.scrollTarget - scrollXPos;
+  if (Math.abs(diff) < 1) {
+    this.scrollbar_.set(this.scrollTarget);
+    return;
+  }
+  this.scrollbar_.set(scrollXPos + diff * 0.3);
```
rachel-fenichel: Make 0.3 a constant in this file or in flyout_base.

core/flyout_horizontal.js
```diff
@@ -245,16 +245,60 @@ Blockly.HorizontalFlyout.prototype.scrollToStart = function() {
...
  * @param {!Event} e Mouse wheel scroll event.
  * @private
  */
 Blockly.HorizontalFlyout.prototype.wheel_ = function(e) {
+  // remove scrollTarget to stop auto scrolling in step()
+  this.scrollTarget = null;
+
   var delta = e.deltaX;
 
+  // If we're scrolling more vertically than horizontally,
```
rachel-fenichel: ?  When are you scrolling vertically in a horizontal flyout?

core/flyout_horizontal.js
```diff
@@ -245,16 +245,60 @@ Blockly.HorizontalFlyout.prototype.scrollToStart = function() {
...
+  if (Math.abs(e.deltaY) > Math.abs(delta)) {
+    delta = e.deltaY;
+  }
+
   if (delta) {
-    if (goog.userAgent.GECKO) {
-      // Firefox's deltas are a tenth that of Chrome/Safari.
+    // Firefox's mouse wheel deltas are a tenth that of Chrome/Safari.
+    // DeltaMode is 1 for a mouse wheel, but not for a trackpad scroll event
+    if (goog.userAgent.GECKO && (e.deltaMode === 1)) {
```
rachel-fenichel, line 289: This also looks like a change that needs to go upstream.

core/flyout_horizontal.js
```diff
@@ -275,6 +321,22 @@ Blockly.HorizontalFlyout.prototype.wheel_ = function(e) {
 };
 
 /**
+ * Select a category using the scroll position
+ * @param  {number} pos The scroll position.
+ */
+Blockly.HorizontalFlyout.prototype.selectCategoryByScrollPosition = function(pos) {
```
rachel-fenichel: This function is the same for horizontal and vertical. Move it to flyout_base, and mark it package unless you're planning to use it externally.

core/flyout_horizontal.js
```diff
@@ -275,6 +321,22 @@ Blockly.HorizontalFlyout.prototype.wheel_ = function(e) {
...
+  var scaledPos = pos / this.workspace_.scale;
+  // Traverse the array of scroll positions in reverse, so we can select the lowest
```
rachel-fenichel: Update this to refer to "before", "past", and "furthest" (or similar terms) so that it's not specific to horizontal or vertical.

core/flyout_horizontal.js
```diff
@@ -275,6 +321,22 @@ Blockly.HorizontalFlyout.prototype.wheel_ = function(e) {
...
+  // category that the scroll position is below
+  for (var i=this.categoryScrollPositions.length-1; i>=0; i--) {
```
rachel-fenichel: Spaces around operators, here and everywhere else.

core/flyout_vertical.js
```diff
@@ -332,16 +333,54 @@ Blockly.VerticalFlyout.prototype.scrollToStart = function() {
 };
 
 /**
+ * Scroll the flyout to a position.
+ * @param {Number} pos The targeted scroll position.
```
rachel-fenichel: Use `number` instead of `Number`.

core/flyout_vertical.js
```diff
@@ -332,16 +333,54 @@ Blockly.VerticalFlyout.prototype.scrollToStart = function() {
...
+  this.scrollTarget = Math.min(this.scrollTarget, contentHeight - viewHeight);
+
+  this.step();
+};
+
+/**
+ * Step the scrolling animation by scrolling a fraction of the way to
+ * a scroll target, and request the next frame if necessary.
+ */
+Blockly.VerticalFlyout.prototype.step = function() {
```
rachel-fenichel: Same comments as on the horizontal version.  it looks like you may just be able to combine the horizontal and vertical versions into a helper function in flyout_base.  You would pass in either `this.workspace_.scrollY` or `this.workspace_.scrollX` in the horizontal and vertical implementations.

core/toolbox.js
```diff
@@ -179,10 +179,30 @@ Blockly.Toolbox.prototype.createFlyout_ = function() {
  */
 Blockly.Toolbox.prototype.populate_ = function(newTree) {
   this.categoryMenu_.populate(newTree);
+  this.showAll();
   this.setSelectedItem(this.categoryMenu_.categories_[0]);
 };
 
 /**
+ * Show all blocks for all categories in the flyout
```
rachel-fenichel, line 187: Make this function private if possible.

core/toolbox.js
```diff
@@ -273,7 +293,7 @@ Blockly.Toolbox.prototype.getClientRect = function() {
   var x = toolboxRect.left;
   var y = toolboxRect.top;
   var width = this.getWidth();
-  var height = toolboxRect.height;
+  var height = toolboxRect.height + this.flyout_.height_;
```
rachel-fenichel: Please explain this change and the change on line 289.

core/toolbox.js
```diff
@@ -296,10 +316,7 @@ Blockly.Toolbox.prototype.getClientRect = function() {
  * procedures.
  */
 Blockly.Toolbox.prototype.refreshSelection = function() {
-  var selectedItem = this.getSelectedItem();
-  if (selectedItem && selectedItem.getContents()) {
-    this.flyout_.show(selectedItem.getContents());
-  }
+  this.showAll();
```
rachel-fenichel: This means you'll scroll to the top of the toolbox every time you add a variable.  That seems wrong.  You still need to keep track of the selected category and scroll back to it, at a minimum.

core/toolbox.js
```diff
@@ -315,18 +332,39 @@ Blockly.Toolbox.prototype.getSelectedItem = function() {
...
+        }
+        return;
+      }
+    }
+  }
+};
+
+/**
+ * Select a category by name.
+ * @param {string} name The name of the category to select.
```
rachel-fenichel, line 364: Mark this @package.

core/toolbox.js
```diff
@@ -315,18 +332,39 @@ Blockly.Toolbox.prototype.getSelectedItem = function() {
...
   if (this.selectedItem_ != null) {
     this.selectedItem_.setSelected(true);
-    this.flyout_.show(item.getContents());
-    this.flyout_.scrollToStart();
+    // Scroll flyout to the top of the selected category
+    var categoryName = item.name_;
+    for (var i=0; i<this.flyout_.buttons_.length; i++) {
+      if (this.flyout_.buttons_[i].text_ === categoryName) {
+        if (this.horizontalLayout_) {
+          this.flyout_.scrollTo(this.flyout_.buttons_[i].position_.x);
```
rachel-fenichel: You're looking at a private member (position_) of a private member (buttons_), and you're doing it all over the place in this chunk of code.  Please figure out what information you want made accessible where, and then try to get it without just reaching through arbitrary private pieces of code.  If you need a position and button stored for every category, maybe you should store those in the Blockly.Toolbox.Category instances and keep them updated.

rachel-fenichel (changes requested):
- High level, this code is assuming that all labels are category labels, which means you aren't going to allow developers to put labels in their categories.  That seems like a mistake to me.  There's also a lot of code that reaches into private members, especially of buttons, which isn't okay.

- Please make all of your functions as private as possible.  If they cannot be private but should only be used by the rest of the library, mark them @package.

- Make sure you're following the style guide: https://developers.google.com/blockly/guides/modify/web/style-guide

- In particular, comments should start with capital letters and end with periods.  Also, put spaces around operators such as `=` and `<`.  (I'm surprised that the linter doesn't catch that, actually.)

core/flyout_base.js
ericrosenbaum: got it

core/flyout_base.js
ericrosenbaum: got it

core/flyout_base.js
ericrosenbaum: gotcha

core/flyout_base.js
ericrosenbaum: All set

core/flyout_base.js
ericrosenbaum: yup

core/flyout_base.js
ericrosenbaum: Good point. I've added an XML property `category-label="true"`, which flyout_button uses to set a boolean property isCategoryLabel_, and then I check for that in recordCategoryScrollPositions_().

core/flyout_base.js
ericrosenbaum: got it.

core/flyout_button.js
ericrosenbaum: Submitted

core/flyout_horizontal.js
ericrosenbaum: oops thanks

core/flyout_horizontal.js
ericrosenbaum: good catch

core/flyout_horizontal.js
ericrosenbaum: all set

core/flyout_vertical.js
ericrosenbaum: Yes! Did it. I used the horizontalLayout_ property in flyout_base to switch between scrollX and scrollY.

core/flyout_horizontal.js
ericrosenbaum: yep

core/flyout_horizontal.js
ericrosenbaum: done

core/flyout_horizontal.js
ericrosenbaum: Great. I called it scrollAnimationFraction.

core/flyout_horizontal.js
ericrosenbaum: I'm referring here to the amount of scrolling coming from the event (a mouse wheel or touchpad scroll event). On a touchpad, you might be scrolling vertically, even when viewing a horizontal flyout. On a mouse wheel, you can _only_ scroll vertically. This code handles these two cases, using the vertical scroll amount to scroll the flyout horizontally. I'll try to clarify the comment in the code!

core/flyout_horizontal.js
ericrosenbaum: Cool- made a PR

core/flyout_horizontal.js
ericrosenbaum: nice- done!

core/flyout_horizontal.js
ericrosenbaum: all set

core/flyout_horizontal.js
ericrosenbaum: Fixed

core/flyout_vertical.js
ericrosenbaum: all set

core/toolbox.js
ericrosenbaum: yup

core/toolbox.js
ericrosenbaum: I've made a comment in the code explaining that this lets us correctly calculate the delete area for the horizontal toolbox.

core/toolbox.js
ericrosenbaum: Weird- I'm not seeing this problem. ShowAll is not causing the scroll position to change.

core/toolbox.js
ericrosenbaum: This is fixed

core/toolbox.js
ericrosenbaum: done

core/flyout_base.js
```diff
@@ -441,25 +449,26 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
   this.setVisible(true);
   // Create the blocks to be shown in this flyout.
   var contents = [];
   var gaps = [];
   this.permanentlyDisabled_.length = 0;
   for (var i = 0, xml; xml = xmlList[i]; i++) {
+    // Handle dynamic categories, represented by a name instead of a list of XML.
+    // Look up the correct category generation function and call that to get a
+    // valid XML list.
+    if (typeof xmlList[i] === 'string') {
```
rachel-fenichel: nit: this can be `typeof xml`

core/flyout_base.js
```diff
@@ -520,6 +529,66 @@ Blockly.Flyout.prototype.show = function(xmlList) {
 
   this.reflowWrapper_ = this.reflow.bind(this);
   this.workspace_.addChangeListener(this.reflowWrapper_);
+
+  this.recordCategoryScrollPositions_();
+};
+
+/**
+ * Store an array of with category names and scrollbar positions.
```
rachel-fenichel: Typo: "array of with"

core/flyout_base.js
```diff
@@ -520,6 +529,66 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
+ * This is used when scrolling the flyout to cause a category to be selected.
+ * @private
+ */
+Blockly.Flyout.prototype.recordCategoryScrollPositions_ = function() {
+  this.categoryScrollPositions = [];
+  for (var i = 0; i < this.buttons_.length; i++) {
+    if (this.buttons_[i].isCategoryLabel_) {
```
rachel-fenichel: You're still directly accessing internals of the buttons (`isCategoryLabel_`, `position_`, `text_`).  Please don't.

core/flyout_base.js
```diff
@@ -520,6 +529,66 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
+        position: this.horizontalLayout_ ?
+          categoryLabel.position_.x : categoryLabel.position_.y
+      });
+    }
+  }
+};
+
+/**
+ * Select a category using the scroll position.
+ * @param  {number} pos The scroll position.
```
rachel-fenichel: What are the units for pos?

If you're not sure, add a TODO about figuring out the units, so I can follow up on that.

`scaledPos` below seems wrong, though.  Usually `scaledPos` would be a position multiplied by the workspace scale, and this is the inverse.  It's probably the position in workspace units, with pos originally being in pixels, but you should check.

core/flyout_button.js
rachel-fenichel: Thanks!

core/flyout_horizontal.js
```diff
@@ -245,16 +245,46 @@ Blockly.HorizontalFlyout.prototype.scrollToStart = function() {
...
+ */
+Blockly.HorizontalFlyout.prototype.scrollTo = function(pos) {
+  this.scrollTarget = pos * this.workspace_.scale;
+
+  // Make sure not to set the scroll target past the farthest point we can
+  // scroll to, i.e. the content width minus the view width
+  var metrics = this.workspace_.getMetrics();
+  var contentWidth = metrics.contentWidth;
+  var viewWidth = metrics.viewWidth;
+  this.scrollTarget = Math.min(this.scrollTarget, contentWidth - viewWidth);
```
rachel-fenichel, line 261: Add `scrollTarget` to `Flyout`, initialized to `null` in the constructor, and add jsdoc saying what it is and what its possible values are (null or a number, as far as I can tell).

If it's only used in flyout and its children, mark it package.  I'd prefer private but I need to check if the closure compiler would throw warnings as a result.

core/flyout_horizontal.js
```diff
@@ -245,16 +245,46 @@ Blockly.HorizontalFlyout.prototype.scrollToStart = function() {
 };
 
 /**
+ * Scroll the flyout to a position.
+ * @param {number} pos The targeted scroll position.
+ */
```
rachel-fenichel: @package

core/flyout_vertical.js
```diff
@@ -37,6 +37,7 @@ goog.require('goog.dom');
 goog.require('goog.events');
 goog.require('goog.math.Rect');
 goog.require('goog.userAgent');
+goog.require('goog.dom.animationFrame.polyfill');
```
rachel-fenichel: Nit: move this just above goog.events.  They're alphabetical.

core/toolbox.js
```diff
@@ -273,7 +296,10 @@ Blockly.Toolbox.prototype.getClientRect = function() {
   var x = toolboxRect.left;
   var y = toolboxRect.top;
   var width = this.getWidth();
-  var height = toolboxRect.height;
+  // The height here is the combined height of the category menu and flyout.
```
rachel-fenichel: This should not include the height of the flyout.  This is the deletion rect for the toolbox.  The flyout is responsible for its own deletion rect.

core/toolbox.js
```diff
@@ -286,7 +312,7 @@ Blockly.Toolbox.prototype.getClientRect = function() {
     return new goog.math.Rect(-BIG_NUM, -BIG_NUM, 2 * BIG_NUM,
         BIG_NUM + y + height);
   } else {  // Bottom
-    return new goog.math.Rect(0, y, 2 * BIG_NUM, BIG_NUM + width);
+    return new goog.math.Rect(0, y - height, 2 * BIG_NUM, BIG_NUM);
```
rachel-fenichel: I agree that the + width is unnecessary, but subtracting height doesn't work here. 

Scribble skills on point:

image

core/flyout_button.js
```diff
@@ -223,6 +223,31 @@ Blockly.FlyoutButton.prototype.getTargetWorkspace = function() {
 };
 
 /**
+ * Get whether this button is a label at the top of a category.
+ * @return {boolean} True if it is a category label.
```
rachel-fenichel, line 227: @package

core/flyout_button.js
```diff
@@ -223,6 +223,31 @@ Blockly.FlyoutButton.prototype.getTargetWorkspace = function() {
...
+ */
+Blockly.FlyoutButton.prototype.getIsCategoryLabel = function() {
+  return this.isCategoryLabel_;
+};
+
+/**
+ * Get the text of this button.
+ * @return {string} The text on the button.
+ */
+Blockly.FlyoutButton.prototype.getText = function() {
```
rachel-fenichel: @package

core/flyout_button.js
```diff
@@ -223,6 +223,31 @@ Blockly.FlyoutButton.prototype.getTargetWorkspace = function() {
...
+  return this.text_;
+};
+
+
+/**
+ * Get the position of this button.
+ * @return {!goog.math.Coordinate} The button position.
```
rachel-fenichel: @package, and remove the extra newline above this comment.

rachel-fenichel (commented):
Three nits and you're good to go.  I don't need to do another review--it's just a few missing annotations.

PR #1062, opened 2017-08-29
Add eyedropper button and callback

### LINKED ISSUE #1057: Eyedropper callback hooks for color editor
We want to create a way to add an external "eyedropper" tool for the color picker. For Scratch 3, the use-case would be to add a way to pick colors from the stage. I imagine what we need is to define a callback, like `Blockly.activateEyedropper`, that is `null` in scratch-blocks but could be defined as a function externally. Then in the `field_colour#showEditor` function, we could check if `activateEyedropper` is defined, and if so, to add a button for activating the eyedropper. When clicked, the button would call `activateEyedropper` with a function like `onChooseColor` that?sets the field value when called. This way the eyedropper could return a color to the field. 

Does this sound like a reasonable approach @picklesrus @rachel-fenichel ?

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/1057

### Proposed Changes

This is the code implementation of the idea from the linked issue. Figured it would be easier to read code than an abstract idea. This PR adds a new property, `Blockly.FieldColour.activateEyedropper` which is initialized to `null`. In the `showEditor` function, if `activateEyedropper` has been defined, a button is added to the editor that will call `activateEyedropper(callback)`. The callback is a function that will set the value of the field. 

### Reason for Changes

This is a proposal for how to add an optional eyedropper tool that is managed external to the blocks.


core/field_colour.js
```diff
@@ -216,6 +224,19 @@ Blockly.FieldColour.prototype.showEditor_ = function() {
   Blockly.WidgetDiv.position(xy.x, xy.y, windowSize, scrollOffset,
                              this.sourceBlock_.RTL);
 
+  if (Blockly.FieldColour.activateEyedropper) {
+    var button = new goog.ui.Button('Eyedropper');
```
picklesrus: Neil was always very careful about the dependencies closure UI components he added.  As long as the size of goog.Ui.Button isn't big compared to the functionality it provides over using a built in like <button>, it is fine. I want to make sure we make that explicit choice though.

core/field_colour.js
```diff
@@ -148,6 +149,13 @@ Blockly.FieldColour.COLOURS = goog.ui.ColorPicker.SIMPLE_GRID_COLORS;
 Blockly.FieldColour.COLUMNS = 7;
 
 /**
+ * Function to be called if eyedropper can be activated.
+ * If defined, an eyedropper button will be added to the color picker.
+ * The button calls this function with a callback to update the field value.
+ */
+Blockly.FieldColour.activateEyedropper = null;
```
picklesrus: Would this be global for all color picker fields, no matter what block they're attached to? I'm worried that we don't actually want that. It might be fine in the context of Scratch, but for scratch-blocks, it'd be nice if we could be more flexible. I think it is unfortunate that the colors themselves are also global across all color fields and I'd like to avoid adding more if we can.

picklesrus (commented):
Thought about this a little more and there's one point (the one about global stuff below) that's probably worth discussing.

core/field_colour.js
paulkaplan: Ah good point. No this doesn't require the closure UI element, I can do it with basic html elements. Changed in 2d3c915

core/field_colour.js
paulkaplan: Recapping what we discussed offline: I think we need to take a scratch-3-first approach here, since we can't really figure out any situations that would require this. Once they arise we may be able to re-evaluate. 

One switch we may experiment with is a color field that goes directly to color picking mode, but we can add that later if we find it is useful.

paulkaplan:
@picklesrus can you give this another look through with my updates?

core/field_colour.js
picklesrus: The problem with this approach is once you expose this "api" publicly, you can't really get rid of it because other developers may be using it and we have no way of knowing.

But redesigning the field API isn't happening quickly... This isn't great either but what do you think about you naming Blockly.FieldColour.activateEyedropper with a trailing underscore and marking it private with a note?  You'll then be violating private by using it in Scratch gui or vm, but it at least gives you the flexibility to rip it out later without pulling the rug out from other developers.

paulkaplan:
@picklesrus that is a reasonable sounding plan. done in b388e70

picklesrus:
One minor comment, but LGTM!

PR #1075, opened 2017-09-06
Slider color picker

### LINKED ISSUE #460: Specialized Menu — Color Picker
Breaking out from [Design - Specialized field editors - Issue #149](https://github.com/LLK/scratch-blocks/issues/149)

For example, this will be used in places like 'color touching [  ] ?'
screen shot 2016-06-29 at 9 26 51 am

paulkaplan:
Only the last commit here is new, the others are the currently unmerged eyedropper pr, so ignore those. 

### Resolves

This PR implements the hue, saturation, brightness color sliders instead of the color palette. 

sliders

^ gifs quantize color space a lot, here is a still of those backgrounds

image

Resolves https://github.com/LLK/scratch-blocks/issues/460

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -823,18 +823,18 @@ Blockly.Css.CONTENT = [
     'cursor: pointer;',
   '}',
 
-  '.scratchColorPickerLabel {',
+  '.scratchColourPickerLabel {',
     'font-family: "Helvetica Neue", Helvetica, sans-serif;',
     'font-size: 0.65rem;',
     'color: $colour_toolboxText;',
     'margin: 8px;',
   '}',
 
-  '.scratchColorPickerLabelText {',
+  '.scratchColourPickerLabelText {',
     'font-weight: bold;',
   '}',
 
-  '.scratchColorPickerReadout {',
+  '.scratchColourPickerReadout {',
     'margin-left: 10px;',
   '}',
 
--- a/core/css.js
+++ b/core/css.js
@@ -790,54 +790,52 @@ Blockly.Css.CONTENT = [
     'color: #fff;',
   '}',
 
-  /* Copied from: goog/css/colorpicker-simplegrid.css */
-  /*
-   * Copyright 2007 The Closure Library Authors. All Rights Reserved.
-   *
-   * Use of this source code is governed by the Apache License, Version 2.0.
-   * See the COPYING file for details.
-   */
-
-  /* Author: pupius@google.com (Daniel Pupius) */
-
-  /*
-    Styles to make the colorpicker look like the old gmail color picker
-    NOTE: without CSS scoping this will override styles defined in palette.css
-  */
-  '.blocklyWidgetDiv .goog-palette {',
-    'outline: none;',
-    'cursor: default;',
-  '}',
-
-  '.blocklyWidgetDiv .goog-palette-table {',
-    'border-collapse: collapse;',
-  '}',
-
-  '.blocklyWidgetDiv .goog-palette-cell {',
-    'height: 13px;',
-    'width: 15px;',
-    'margin: 0;',
-    'border: 0;',
-    'text-align: center;',
-    'vertical-align: middle;',
-    'font-size: 1px;',
-  '}',
-
-  '.blocklyWidgetDiv .goog-palette-colorswatch {',
+  '.blocklyDropDownDiv .goog-slider-horizontal {',
+    'margin: 8px;',
+    'height: 22px;',
+    'width: 150px;',
     'position: relative;',
-    'height: 13px;',
-    'width: 15px;',
+    'outline: none;',
+    'border-radius: 11px;',
+    'margin-bottom: 20px;',
   '}',
 
-  '.blocklyWidgetDiv .goog-palette-cell-hover .goog-palette-colorswatch {',
-    'border: 1px solid #FFF;',
-    'box-sizing: border-box;',
+  '.blocklyDropDownDiv .goog-slider-horizontal .goog-slider-thumb {',
+    'width: 26px;',
+    'height: 26px;',
+    'margin-top: -1px;',
+    'position: absolute;',
+    'background-color: white;',
+    'border-radius: 100%;',
+    '-webkit-box-shadow: 0 0 0 4px rgba(0, 0, 0, 0.15);',
+    '-moz-box-shadow: 0 0 0 4px rgba(0, 0, 0, 0.15);',
+    'box-shadow: 0 0 0 4px rgba(0, 0, 0, 0.15);',
   '}',
 
-  '.blocklyWidgetDiv .goog-palette-cell-selected .goog-palette-colorswatch {',
-    'border: 1px solid #000;',
-    'box-sizing: border-box;',
-    'color: #fff;',
+  '.scratchEyedropper {',
+    'background: none;',
+    'outline: none;',
+    'border: none;',
+    'width: 100%;',
+    'text-align: center;',
+    'border-top: 1px solid #ddd;',
+    'padding-top: 5px;',
+    'cursor: pointer;',
+  '}',
+
+  '.scratchColorPickerLabel {',
+    'font-family: "Helvetica Neue", Helvetica, sans-serif;',
+    'font-size: 0.65rem;',
+    'color: $colour_toolboxText;',
+    'margin: 8px;',
+  '}',
+
+  '.scratchColorPickerLabelText {',
+    'font-weight: bold;',
+  '}',
+
+  '.scratchColorPickerReadout {',
+    'margin-left: 10px;',
   '}',
 
   /* Copied from: goog/css/menu.css */
--- a/core/field_colour.js
+++ b/core/field_colour.js
@@ -27,12 +27,10 @@
 goog.provide('Blockly.FieldColour');
 
 goog.require('Blockly.Field');
-goog.require('Blockly.DropDownDiv');
 goog.require('goog.dom');
 goog.require('goog.events');
 goog.require('goog.style');
-goog.require('goog.color');
-goog.require('goog.ui.Slider');
+goog.require('goog.ui.ColorPicker');
 
 /**
  * Class for a colour input field.
@@ -66,24 +64,25 @@ Blockly.FieldColour.prototype.colours_ = null;
 Blockly.FieldColour.prototype.columns_ = 0;
 
 /**
- * Function to be called if eyedropper can be activated.
- * If defined, an eyedropper button will be added to the color picker.
- * The button calls this function with a callback to update the field value.
- */
-Blockly.FieldColour.activateEyedropper = null;
-
-/**
- * Path to the eyedropper svg icon.
- */
-Blockly.FieldColour.EYEDROPPER_PATH = 'eyedropper.svg';
-
-/**
  * Install this field on a block.
  * @param {!Blockly.Block} block The block containing this field.
  */
 Blockly.FieldColour.prototype.init = function(block) {
   Blockly.FieldColour.superClass_.init.call(this, block);
   this.setValue(this.getValue());
+};
+
+/**
+ * Mouse cursor style when over the hotspot that initiates the editor.
+ */
+Blockly.FieldColour.prototype.CURSOR = 'default';
+
+/**
+ * Close the colour picker if this input is being deleted.
+ */
+Blockly.FieldColour.prototype.dispose = function() {
+  Blockly.WidgetDiv.hideIfOwner(this);
+  Blockly.FieldColour.superClass_.dispose.call(this);
 };
 
 /**
@@ -95,11 +94,10 @@ Blockly.FieldColour.prototype.getValue = function() {
 };
 
 /**
- * Set the colour. If opt_fromSliders is true, do not update the sliders.
+ * Set the colour.
  * @param {string} colour The new colour in '#rrggbb' format.
- * @param {boolea} opt_fromSliders Flag to prevent sliders from recursing on themselves.
  */
-Blockly.FieldColour.prototype.setValue = function(colour, opt_fromSliders) {
+Blockly.FieldColour.prototype.setValue = function(colour) {
   if (this.sourceBlock_ && Blockly.Events.isEnabled() &&
       this.colour_ != colour) {
     Blockly.Events.fire(new Blockly.Events.BlockChange(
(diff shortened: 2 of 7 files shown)


core/css.js
```diff
@@ -790,54 +790,52 @@ Blockly.Css.CONTENT = [
...
+  '.scratchColorPickerLabelText {',
+    'font-weight: bold;',
   '}',
 
-  '.blocklyWidgetDiv .goog-palette-cell-selected .goog-palette-colorswatch {',
-    'border: 1px solid #000;',
-    'box-sizing: border-box;',
-    'color: #fff;',
+  '.scratchColorPickerReadout {',
+    'margin-left: 10px;',
```
picklesrus: margin-left/right usually cause problems in RTL. Have you checked this one?

core/css.js
```diff
@@ -790,54 +790,52 @@ Blockly.Css.CONTENT = [
...
+    'padding-top: 5px;',
+    'cursor: pointer;',
   '}',
 
-  '.blocklyWidgetDiv .goog-palette-colorswatch {',
-    'position: relative;',
-    'height: 13px;',
-    'width: 15px;',
+  '.scratchColorPickerLabel {',
+    'font-family: "Helvetica Neue", Helvetica, sans-serif;',
```
picklesrus: is this font different than the rest of the page?

core/css.js
```diff
@@ -790,54 +790,52 @@ Blockly.Css.CONTENT = [
...
+    'box-shadow: 0 0 0 4px rgba(0, 0, 0, 0.15);',
   '}',
 
-  '.blocklyWidgetDiv .goog-palette-cell {',
-    'height: 13px;',
-    'width: 15px;',
-    'margin: 0;',
-    'border: 0;',
+  '.scratchEyedropper {',
+    'background: none;',
```
picklesrus, line 816: I might find this out later when I get farther in, so apologies, but aren't background, outline, and border none by default?

core/css.js
```diff
@@ -790,54 +790,52 @@ Blockly.Css.CONTENT = [
...
+    'border-radius: 11px;',
+    'margin-bottom: 20px;',
   '}',
 
-  '.blocklyWidgetDiv .goog-palette-table {',
-    'border-collapse: collapse;',
+  '.blocklyDropDownDiv .goog-slider-horizontal .goog-slider-thumb {',
+    'width: 26px;',
+    'height: 26px;',
+    'margin-top: -1px;',
```
picklesrus, line 806: Why does the margin need to be negative? I'm ok with it but always ask since I made some really bad mistakes with this in the past (the IE 6 era past).

core/field_colour.js
```diff
@@ -64,25 +66,24 @@ Blockly.FieldColour.prototype.colours_ = null;
...
-  this.setValue(this.getValue());
-};
+Blockly.FieldColour.activateEyedropper = null;
 
 /**
- * Mouse cursor style when over the hotspot that initiates the editor.
+ * Path to the eyedropper svg icon.
  */
-Blockly.FieldColour.prototype.CURSOR = 'default';
+Blockly.FieldColour.EYEDROPPER_PATH = 'eyedropper.svg';
```
picklesrus: Should this icon live in the media directory with the others?

core/field_colour.js
```diff
@@ -64,25 +66,24 @@ Blockly.FieldColour.prototype.colours_ = null;
...
 
 /**
- * Close the colour picker if this input is being deleted.
+ * Install this field on a block.
+ * @param {!Blockly.Block} block The block containing this field.
  */
-Blockly.FieldColour.prototype.dispose = function() {
```
picklesrus: Did dispose go away? Or maybe I'm looking at the wrong diff?

core/field_colour.js
```diff
@@ -94,10 +95,11 @@ Blockly.FieldColour.prototype.getValue = function() {
 };
 
 /**
- * Set the colour.
+ * Set the colour. If opt_fromSliders is true, do not update the sliders.
  * @param {string} colour The new colour in '#rrggbb' format.
+ * @param {boolea} opt_fromSliders Flag to prevent sliders from recursing on themselves.
  */
-Blockly.FieldColour.prototype.setValue = function(colour) {
+Blockly.FieldColour.prototype.setValue = function(colour, opt_fromSliders) {
```
picklesrus: Given the mess that is currently text vs value in the field code and the amount of inheritance going on, I'm worried about changing the parameters of setValue.  Is there any harm in calling updateSliderHandles_? Or maybe you could check whether the value has changed and not call it it if it hasn't?

core/field_colour.js
```diff
@@ -107,44 +109,100 @@ Blockly.FieldColour.prototype.setValue = function(colour) {
...
+  }
+  this.updateDom_();
 };
 
 /**
- * Get the text from this field.  Used when the block is collapsed.
- * @return {string} Current text.
+ * Create the hue, saturation or value CSS gradient for the slide backgrounds.
+ * @param {string} channel – Either "hue", "saturation" or "value".
+ * @return {string} Array color hex color stops for the given channel
```
picklesrus: this should have the private tag. Same for all the private methods you're adding.

core/field_colour.js
```diff
@@ -107,44 +109,100 @@ Blockly.FieldColour.prototype.setValue = function(colour) {
...
+    }
   }
-  return colour;
+  return stops;
+};
+
+/**
+ * Set the gradient CSS properties for the given node and channel
+ * @param {Node} node - The DOM node the gradient will be set on.
+ * @param {string} channel – Either "hue", "saturation" or "value".
```
picklesrus: this should have the private tag

core/field_colour.js
```diff
@@ -107,44 +109,100 @@ Blockly.FieldColour.prototype.setValue = function(colour) {
...
+  goog.style.setStyle(node, 'background',
+        '-moz-linear-gradient(left, ' + stops.join(',') + ')');
+  goog.style.setStyle(node, 'background',
+        '-webkit-linear-gradient(left, ' + stops.join(',') + ')');
+  goog.style.setStyle(node, 'background',
+        '-o-linear-gradient(left, ' + stops.join(',') + ')');
+  goog.style.setStyle(node, 'background',
+        '-ms-linear-gradient(left, ' + stops.join(',') + ')');
+  goog.style.setStyle(node, 'background',
+        'linear-gradient(left, ' + stops.join(',') + ')');
```
picklesrus: perhaps consider constructing this part left, ' + stops.join(',') + ') and the beginning of the function. 
Something like this? goog.style.setStyle(node, 'background', '-moz-linear-gradient(' + gradient;
I'm not actually sure how much you save by not doing the join over and over again though.

core/field_colour.js
```diff
@@ -160,25 +218,54 @@ Blockly.FieldColour.activateEyedropper_ = null;
...
-  this.colours_ = colours;
-  return this;
+Blockly.FieldColour.prototype.createLabelDom_ = function(labelText) {
+  var labelContainer = document.createElement('div');
+  labelContainer.setAttribute('class', 'scratchColorPickerLabel');
+  var readout = document.createElement('span');
+  readout.setAttribute('class', 'scratchColorPickerReadout');
+  var label = document.createElement('span');
+  label.setAttribute('class', 'scratchColorPickerLabelText');
+  label.innerHTML = labelText;
```
picklesrus: We have avoid using innerHtml since technically you can put whatever malicious bits you want in there. As the dom gets built up, it can become difficult to track these down. Would you mind changing your uses out here and above? I haven't tracked down all the types of elements you're calling this on, but setting the textContent should work for spans at least

core/field_colour.js
```diff
@@ -160,25 +218,54 @@ Blockly.FieldColour.activateEyedropper_ = null;
...
+    switch (channel) {
+      case 'hue':
+        hsv[0] = channelValue;
+        break;
+      case 'saturation':
+        hsv[1] = channelValue;
+        break;
+      case 'brightness':
+        hsv[2] = channelValue;
+        break;
```
picklesrus: Do you need a default case?

core/field_colour.js
```diff
@@ -160,25 +218,54 @@ Blockly.FieldColour.activateEyedropper_ = null;
...
+ * @param {string} labelText - Text for the label
+ * @return {Array} The container node and the readout node.
+ * @private
  */
-Blockly.FieldColour.prototype.setColours = function(colours) {
-  this.colours_ = colours;
-  return this;
+Blockly.FieldColour.prototype.createLabelDom_ = function(labelText) {
+  var labelContainer = document.createElement('div');
+  labelContainer.setAttribute('class', 'scratchColorPickerLabel');
```
picklesrus: most things in the api are spelled colour (with the 'u'). I don't think you want to start mixing them.

core/field_colour.js
```diff
@@ -197,47 +284,47 @@ Blockly.FieldColour.prototype.activateEyedropperInternal_ = function() {
...
+  var brightnessElements = this.createLabelDom_('Brightness');
+  div.appendChild(brightnessElements[0]);
+  this.brightnessReadout_ = brightnessElements[1];
+  this.brightnessSlider_ = new goog.ui.Slider();
+  this.brightnessSlider_.setUnitIncrement(2);
+  this.brightnessSlider_.setMinimum(5);
+  this.brightnessSlider_.setMaximum(255);
+  this.brightnessSlider_.render(div);
+
+  Blockly.FieldColour.hueChangeEventKey_ = goog.events.listen(this.hueSlider_,
```
picklesrus: This might be fine given the changeListener is handled by closure, but I would carefully test, especially on touch devices, what happens when you start a slider change inside the editor then drag your mouse/finger outside of the element and mouse up somewhere else in the blockly window.

picklesrus (commented):
I think you should consider putting this in a separate class (maybe FieldColorSlider or something) for two reasons.
1. FieldColour in upstream can't change to this because too many developers already use the color picker as is.  That means this file is going to be a pain to merge.
2. Color picking is a relatively common thing for developers to want and not all developers are going to want to give this much control to users. e.g. if you want a block that controls LEDs, folks probably want a discrete list similar to the existing picker.

Another consideration: Do you want to parameterize the colors you can represent with the sliders so developers have more control?

core/field_colour.js
paulkaplan: Hm the problem is that updating the sliders causes a change event to fire, which would cause a loop, that is why I don't update the sliders if the update came from the sliders. Is there a different way to handle that?

core/css.js
paulkaplan: It is a button element, so their are several default styles to be overridden.

core/css.js
paulkaplan: It is used to bring the slider handles "up" because they are taller than their parent container. I'll check some other browsers to confirm.

paulkaplan:
@picklesrus updated according to your comments!

core/field_colour_slider.js
```diff
@@ -0,0 +1,346 @@
...
+ * @private
+ */
+Blockly.FieldColourSlider.prototype.createLabelDom_ = function(labelText) {
+  var labelContainer = document.createElement('div');
+  labelContainer.setAttribute('class', 'scratchColourPickerLabel');
+  var readout = document.createElement('span');
+  readout.setAttribute('class', 'scratchColourPickerReadout');
+  var label = document.createElement('span');
+  label.setAttribute('class', 'scratchColourPickerLabelText');
+  label.innerHTML = labelText;
```
picklesrus: I think this can be .textContent = labelText  instead to avoid arbitrary things being dumped in via .innerHTML

core/field_colour_slider.js
```diff
@@ -0,0 +1,346 @@
...
+  labelContainer.appendChild(readout);
+  return [labelContainer, readout];
+};
+
+/**
+ * Factory for creating the different slider callbacks
+ * @param {string} channel - One of "hue", "saturation" or "brightness"
+ * @return {function} the callback for slider update
+ */
+Blockly.FieldColourSlider.prototype.sliderCallbackFactory = function(channel) {
```
picklesrus: private?

core/field_colour_slider.js
```diff
@@ -0,0 +1,346 @@
...
+    this.hueReadout_.textContent = Math.floor(100 * hsv[0] / 360).toFixed(0);
+    this.saturationReadout_.textContent = Math.floor(100 * hsv[1]).toFixed(0);
+    this.brightnessReadout_.textContent = Math.floor(100 * hsv[2] / 255).toFixed(0);
+  }
+};
+
+/**
+ * Update the slider handle positions
+ */
+Blockly.FieldColourSlider.prototype.updateSliderHandles_ = function() {
```
picklesrus: private tag

core/field_colour_slider.js
```diff
@@ -0,0 +1,346 @@
...
+ */
+Blockly.FieldColourSlider.prototype.activateEyedropperInternal_ = function() {
+  var thisField = this;
+  Blockly.FieldColourSlider.activateEyedropper_(function(value) {
+    thisField.setValue(value);
+  });
+};
+
+/**
+ * Create a palette under the colour field.
```
picklesrus: Update this comment to reflect new behavior

core/field_colour_slider.js
```diff
@@ -0,0 +1,346 @@
...
+  }
+
+  Blockly.DropDownDiv.setColour('#ffffff', '#dddddd');
+  Blockly.DropDownDiv.setCategory(this.sourceBlock_.parentBlock_.getCategory());
+  Blockly.DropDownDiv.showPositionedByBlock(this, this.sourceBlock_);
+
+  this.setValue(this.getValue());
+};
+
+Blockly.FieldColourSlider.prototype.dispose = function() {
```
picklesrus, line 330: do you need the widgetdiv hideIfOwner call?

core/field_colour.js
picklesrus: if the update came from the sliders, will the values be the same? in other words, rather than passing the isFromSliders into here, could you instead check if they value you're about to set them to is the same as what they already are and then short circuit that way?

core/field_colour_slider.js
paulkaplan: Oh it doesn't use the widgetdiv anymore (uses the dropdowndiv)

paulkaplan:
@picklesrus thanks for the follow-up. I fixed that issue with the infinite loop in slider setting and remove the `opt_fromSliders`.

core/field_colour_slider.js
paulkaplan: Fixed 5f8184d

core/field_colour_slider.js
paulkaplan: Fixed 5f8184d

core/field_colour_slider.js
paulkaplan: Fixed 5f8184d, looks like i missed one sorry

core/field_colour_slider.js
paulkaplan: Fixed 5f8184d

picklesrus:
Awesome. Thanks! LGTM

PR #1093, opened 2017-09-22
Add a function to duplicate a block and drag the duplicate

### LINKED ISSUE #1092: Duplicate-on-drag for custom procedure arguments
### Expected Behavior
39e68b4f-03bb-4395-a5a3-fa763ab5bf86

### Actual Behavior

Not yet implemented

### Steps to Reproduce

Create a custom block with arguments.  Drag one of the inputs out of the custom block definition.

Any

rachel-fenichel:
### Resolves

#1092 

### Proposed Changes

Add a function that duplicates the gesture's target block and drags the new block instead of the original block.

I tested the function by using it on every drag, but in this PR I removed all calls to it.

### Reason for Changes

Custom procedure arguments need to have this behaviours.

### Test Coverage

Manually tested in chrome in the vertical playground by enabling it for all drags:
ac1e9f4f-b0fa-430c-b23f-e7f1c0cd6965

To enable it for all drags I inserted `this.duplicateOnDrag_()` as the first line of `Blockly.Gesture.prototype.startDraggingBlock_`.

I also checked that I can drag the duplicated block directly to the flyout to delete it with no issues.

### DIFF
--- a/core/gesture.js
+++ b/core/gesture.js
@@ -793,6 +793,8 @@ Blockly.Gesture.prototype.hasStarted = function() {
   return this.hasStarted_;
 };
 
+/* Scratch-specific */
+
 /**
  * Don't even think about using this function before talking to rachel-fenichel.
  *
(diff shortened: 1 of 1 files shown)


paulkaplan:
Couple questions:

1. Could you add to the PR description how you used that function to implement what we are seeing in the GIF for posterity?
2. Should their be any explicit checks on how this gets used (for example, making it throw if it is being used in the flyout, or enforcing that comment about not using it on blocks with children)

rachel-fenichel:
1. Comment updated.
2. It's possible that there should be checks, but I'm not yet sure what will trigger this action.  I'm going to hold off on additional logic until I know when this function should actually be called.

paulkaplan (approved):
Looks good

PR #1106, opened 2017-10-02
Extract code for drawing the right side of a statement input.

rachel-fenichel:
### Resolves

None

### Proposed Changes

Cleanup of code in renderDrawRight_

### Reason for Changes

Code readability: breaks rendering code into named pieces with explanations of what each piece of code does and where it expects to start drawing.  There's no change in behaviour.

### Test Coverage

Visually confirmed that all blocks render and look good in the vertical playground.


rachel-fenichel:
Friendly ping.

paulkaplan (approved):
Looks good!

PR #1107, opened 2017-10-02
Extract code for rendering the right side of a reporter

rachel-fenichel:
This PR builds on #1106 and is intended to be merged after the other.  The new code is in the second commit.

### Resolves

None

### Proposed Changes

Cleanup of code in renderDrawRight_.  This creates a separate function for drawing the correct right-side edge shape for a reporter.

### Reason for Changes

Code readability: breaks rendering code into named pieces with explanations of what each piece of code does and where it expects to start drawing. There's no change in behaviour.

### Test Coverage

Visually confirmed that all blocks render and look good in the vertical playground.

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -1265,18 +1265,7 @@ Blockly.BlockSvg.prototype.renderDrawRight_ = function(steps,
     }
     cursorY += row.height;
   }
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Friendly ping.

paulkaplan (approved):
LG!

Issue #1070, opened 2017-09-01
Update category menu selection when dragging flyout or scrollbar 


ericrosenbaum:
### Expected Behavior

In the auto-scrolling flyout with all the blocks ([PR](https://github.com/LLK/scratch-blocks/pull/1059)), the selected category should update when you scroll the flyout in any way. 

### Actual Behavior

Currently, the category selection is only updated on a "wheel" scroll event (via mouse wheel or trackpad scrolling gesture). It does not update when you scroll it by dragging the flyout or dragging the scrollbar.

### Operating System and Browser

Mac OS 10.11.6 Chrome 60.0

ericrosenbaum:
Any pointers on this one @picklesrus or @rachel-fenichel?

rachel-fenichel:
Flyout's `setMetrics` is called when it's being scrolled ([vertical](https://github.com/LLK/scratch-blocks/blob/develop/core/flyout_vertical.js#L228) and [horizontal](https://github.com/LLK/scratch-blocks/blob/develop/core/flyout_horizontal.js#L123)).

That's because scrollbar's `set` calls `onScroll_`, which calls `setMetrics` on the relevant workspace:

```

/**
 * Called when scrollbar is moved.
 * @private
 */
Blockly.Scrollbar.prototype.onScroll_ = function() {
  var ratio = this.handlePosition_ / this.scrollViewSize_;
  if (isNaN(ratio)) {
    ratio = 0;
  }
  var xyRatio = {};
  if (this.horizontal_) {
    xyRatio.x = ratio;
  } else {
    xyRatio.y = ratio;
  }
  this.workspace_.setMetrics(xyRatio);
};

/**
 * Set the scrollbar slider's position.
 * @param {number} value The distance from the top/left end of the bar.
 */
Blockly.Scrollbar.prototype.set = function(value) {
  this.setHandlePosition(this.constrainHandle_(value * this.ratio_));
  this.onScroll_();
};
```

`wheel_` also calls `scrollbar.set()`, so you'll want to remove that code from `wheel_` and only have it in `setMetrics`.  Beware of accidentally setting up loops between scrolling and setMetrics.  That could happen if setting the selected category on the toolbox makes it scroll the flyout to that category, for instance.

PR #1108, opened 2017-10-04
Update category menu selection when dragging flyout or scrollbar

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/1070

### Proposed Changes

Move the call to selectCategoryByScrollPosition into setMetrics, so that it is called on all the different types of scroll event (mouse wheel / touchpad scroll, dragging the flyout, dragging the scrollbar). 

This change also requires disabling the updates to category selection during auto-scrolling, so that when you click on a category button, it highlights and stays that way (rather than selecting the categories in between as the scrolling animates).

### Reason for Changes

Without this change, the category menu selection state gets out of sync with the flyout position when you scroll by dragging the flyout or the scrollbar.

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -566,6 +566,11 @@ Blockly.Flyout.prototype.recordCategoryScrollPositions_ = function() {
  * @package
  */
 Blockly.Flyout.prototype.selectCategoryByScrollPosition = function(pos) {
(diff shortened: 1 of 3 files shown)


ericrosenbaum:
Gotta do horizontal flyout too

rachel-fenichel (approved):

Issue #1102, opened 2017-10-02
Implement "Define" blocks


rschamp:
The Define block should look like this: 
![image](https://user-images.githubusercontent.com/1512021/31085338-d3a1cfb6-a764-11e7-94b0-d38abfa8852d.png)

Behavior reference from Scratch 2.0 procedures (unless otherwise noted, 3.0 behavior should match 2.0):

Procedure with no arguments:
![image](https://user-images.githubusercontent.com/1512021/31085418-0be8a5ca-a765-11e7-8582-06d02fff0609.png)

Procedure with only arguments (no label):
![image](https://user-images.githubusercontent.com/1512021/31085424-104a281e-a765-11e7-93c5-15f47324f7ed.png)

Procedure with a label and three arguments, one of each type:
![image](https://user-images.githubusercontent.com/1512021/31085433-18f96164-a765-11e7-9d3e-c36e4ca5d65c.png)

Right-click on “define”:
![image](https://user-images.githubusercontent.com/1512021/31085450-2323d73c-a765-11e7-9c75-852e6fd3649b.png)

Right-click on “foo”:
![image](https://user-images.githubusercontent.com/1512021/31085459-2c525108-a765-11e7-816b-82bcee35b77c.png)

Right-click on empty space next to “foo” (inside the statement block outline):
![image](https://user-images.githubusercontent.com/1512021/31085470-33a1f3c8-a765-11e7-9927-9d9498473251.png)

Right-click on “number1” does nothing (**but in 3.0 should pop up the same context menu as above**): 
![image](https://user-images.githubusercontent.com/1512021/31085481-3e2bbfe0-a765-11e7-8d3d-e3da32462998.png)

Dragging the arguments out duplicates them:
![image14](https://user-images.githubusercontent.com/1512021/31085517-63237dba-a765-11e7-9bd6-7dbb47431a5f.gif)

Dragging from the outline of the defined block (here, “foobar”) moves the “define” block.  It doesn’t create a copy of the new block.
![image9](https://user-images.githubusercontent.com/1512021/31085565-8e07aaf6-a765-11e7-8096-cf93a835c0b2.gif)

rachel-fenichel:
The "define" text is vertically aligned with the text and fields on the inner block, correct?

Are there types of inner blocks that would be taller than the one rendered above?

PR #1109, opened 2017-10-04
Redefine procedures_defnoreturn block and make it render scratch-style

rachel-fenichel:
### Resolves

Part of #1102

### Proposed Changes

The procedures_defnoreturn block now has a block rendered inside it, scratch-style.  It resizes dynamically with the size of the internally rendered block.

The inner block is a shadow block, so dragging it drags the outer block.

The specialized rendering code is *only* for the procedures_defnoreturn block.

image

### Reason for Changes

Support scratch-style custom procedures.

### Test Coverage

Visually--see above--and by making sure that no preexisting blocks look weird in the flyout.

### Future work
You can currently put any statement inside the block: 
image

I need to make sure that only the internally rendered custom command block can go there, and that you can't connect anything (including an instance of the custom command block) by dragging it over.  My plan for this is to add a type check on the input of the define block and the previous and next connections of the internally rendered custom command block, and make sure that nothing else has that same type.

In turn, that means that all of the other blocks need to have a type set on their previous and output connections, since they can otherwise connect to anything.  More [info](https://developers.google.com/blockly/guides/create-custom-blocks/type-checks#statement_stacks) is on the blockly developers site.

As part of this I need to define the custom command block and the argument reporters, and set up their behaviours.

There are probably associated scratch-vm changes related to reading and writing projects with procedures.

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -457,6 +457,34 @@ Blockly.BlockSvg.SHAPE_IN_SHAPE_PADDING = {
   }
 };
 
+Blockly.BlockSvg.DEFINE_HAT_CORNER_RADIUS = 4 * Blockly.BlockSvg.CORNER_RADIUS;
+
+/**
+ * SVG path for drawing the rounded top-left corner.
+ * @const
+ */
+Blockly.BlockSvg.TOP_LEFT_CORNER_DEFINE_HAT =
+    'a ' + Blockly.BlockSvg.DEFINE_HAT_CORNER_RADIUS + ',' +
+    Blockly.BlockSvg.DEFINE_HAT_CORNER_RADIUS + ' 0 0,1 ' +
+    Blockly.BlockSvg.DEFINE_HAT_CORNER_RADIUS + ',-' +
+    Blockly.BlockSvg.DEFINE_HAT_CORNER_RADIUS;
+
+/**
+ * SVG path for drawing the rounded top-left corner.
+ * @const
+ */
+Blockly.BlockSvg.TOP_RIGHT_CORNER_DEFINE_HAT =
+    'a ' + Blockly.BlockSvg.DEFINE_HAT_CORNER_RADIUS + ',' +
+    Blockly.BlockSvg.DEFINE_HAT_CORNER_RADIUS + ' 0 0,1 ' +
+    Blockly.BlockSvg.DEFINE_HAT_CORNER_RADIUS + ',' +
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
cc @carljbowman

core/block_render_svg_vertical.js
```diff
@@ -1113,35 +1141,40 @@ Blockly.BlockSvg.prototype.renderClassify_ = function() {
...
-    steps.push('m 0,0');
-    if (this.startHat_) {
-      steps.push(Blockly.BlockSvg.START_HAT_PATH);
-    }
-    // Skip space for the output shape
-    if (this.edgeShapeWidth_) {
-      steps.push('m ' + this.edgeShapeWidth_ + ',0');
-    }
+  if (this.type == Blockly.BlockSvg.DEFINE_BLOCK_TYPE) {
+    steps.push('m 0, 0');
```
paulkaplan, line 1145: What does the "move relative 0 0" do? seems like a noop?

paulkaplan:
This is looking great! Should we define a new block type for this instead of overwriting `procedures_defnoreturn` just for merge conflicts sake? I guess there is no reason to keep the existing `procedures_defnoreturn`, so maybe not. Just wondering.

@rschamp we haven't talked about the data structure changes that we'll need to make for custom procedures yet, but from this it looks like we won't be using the mutation stuff to store custom procedure data, instead it will go in a statement input to the define block? I am totally new to that part of the code, I just wanted to make sure @rschamp you have this on the radar. 

Other than my curiosity about what `m0,0` does, this LG.

paulkaplan (approved):

core/block_render_svg_vertical.js
rachel-fenichel: er yes that looks like a noop.

rachel-fenichel:
To be clear, you are still going to be using mutations to store custom procedure data.  All of the data is on the inner block (here "move 10 steps") which will be the only block that can attach to this input on the define block.

The define block's only job is to render nicely around the internal block.  The internal block is the one that actually mutates according to the block definition.

rachel-fenichel:
As to which block name to use--you tell me.  I don't have an opinion.  If you want me to leave the old block and just add a new one for the scratch-style definition, that's easy to do.

core/block_render_svg_vertical.js
rachel-fenichel: Actually, it's necessary.  All SVG paths have to starts with a move command to say where the drawing starts--it's not set to 0,0 by default.

Issue #1110, opened 2017-10-04
Implement argument reporters for custom block definition


rachel-fenichel:
![image](https://user-images.githubusercontent.com/13686399/31201410-70729c80-a913-11e7-8251-bbc45eb7a9ed.png)

"number1", "string1", and "boolean1" above, in Scratch 2.0

Scratch 3.0 define block example has "x" as an argument reporter:
![image](https://user-images.githubusercontent.com/13686399/31201463-9c423370-a913-11e7-9f98-959a2745f355.png)

With the current implementation of the define block these have to be shadow blocks (you can't have a non-shadow inside a shadow).

More details:
- [x] Like variable_getters, they only have one field, and it’s a label
- [x] They come in round (string/number) and hex (boolean) versions
- [x] String/number ones are combined
- [x] Dragging from the main <define> block is a duplication, not a disconnection (tracked in #1092)
- [x] Dragging from any other block is a disconnection
- [x] Deleting the define block that they came from does not impact these blocks
- [x] They are never in the toolbox
- [x] They cannot be replaced on the internally rendered custom command block.
  - This might require some cleverness with the type checks.
  - Type check cleverness would also mean that the blocks on the internally rendered custom command block would have to be different from the blocks that are created on duplication
  - Doing these as different types of blocks would be nicer than checking the grandparent of the block to figure out whether  to duplicate or disconnect
- [ ] If used outside of a custom block with that argument name, they evaluate to zero (needs to be done in scratch-vm)
- [x] They may be used inside other custom blocks (other than the one they came from).  
- [ ] If the other block has an argument with the same name, they act like local variables (needs to be done in scratch-vm)
- [x] They may have the same names as global variables
- [x] Rendering: they are rendered with non-shadow colors (for now) to indicate that they are draggable
- [x] Clicking on an argument reporter inside a block definition runs the custom block (done in scratch-blocks)
- [x] Right-clicking when on the internally rendered custom command block gives the context menu for the define block.
- [x] Right-clicking anywhere else gives you a standard context menu
- [x] It’s not the normal variable_getter context menu: there’s no option to select between these blocks

rachel-fenichel:
> They cannot be replaced on the internally rendered custom command block.
> This might require some cleverness with the type checks.
> Type check cleverness would also mean that the blocks on the internally rendered custom command block would have to be different from the blocks that are created on duplication
[...]

PR #1112, opened 2017-10-05
Add a way to specify which blocks should be duplicated on drag, and e…

rachel-fenichel:
…nable it for the new argument_reporter_boolean block when it's a shadow

### Resolves

Part of #1110 

### Proposed Changes

Add a "shouldDuplicateOnDrag" property to gestures, and set it while setting the start block.
Set the target block to the start block (even if it's a shadow) if it should be duplicated on drag, and call gesture.duplicateOnDrag_ to update the drag target when a block drag starts.

Deciding which blocks to duplicate is currently done by inspecting the block type, but could later be generalized.

### Reason for Changes

Argument reporters have special behaviour.

### Test Coverage

Tested in the playground with the new argument_reporter_boolean block.

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -179,3 +179,12 @@ Blockly.Blocks['procedures_param'] = {
     }
   }
 };
+
+Blockly.Blocks['argument_reporter_boolean'] = {
+  init: function() {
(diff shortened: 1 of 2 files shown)


paulkaplan (approved):
LG!

PR #1114, opened 2017-10-05
Fix variables with XML names not appearing

Scimonster:
### Resolves

LLK/scratch-gui#665

Variables with XML names do not work. Two variables named `test` and `<b>test</b>` will both display identically as `test`, and only the one without XML will be updated. Additionally, anything with an `&` in it won't display at all, simply fail with no error (not even in the console).

### Proposed Changes

Escapes the variable name when using it in an XML context.

### Reason for Changes

To fix stated bug. The bug is actually in this codebase, despite being reported on the other one.

### Test Coverage

I didn't see a test file for the `variable.js` file, so there wasn't a natural place to add a test. Please let me know if and where i should add one.

### DIFF
--- a/core/variables.js
+++ b/core/variables.js
@@ -378,7 +378,7 @@ Blockly.Variables.generateVariableFieldXml_ = function(variableModel, opt_name) 
   var name = opt_name || "VARIABLE";
   var xmlString = '<field name="' + name + '" ' + 'variableType="' +
       variableModel.type + '" id="' + variableModel.getId() + '">'+
-      variableModel.name +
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Fixed upstream, waiting for it to be pulled in: https://github.com/google/blockly/pull/1279

rachel-fenichel:
@thisandagain I'm waiting for the upstream fix to be pulled in.

paulkaplan:
Upstream merge solved this! Thanks for the patience.

PR #1122, opened 2017-10-09
Add a bulkUpdate flag to avoid expensive operations (such as toolbox …

### LINKED ISSUE #887: Performance: Creating blocks that refer to variables

rachel-fenichel:
…refresh) during workspace loading.

### Resolves

This is a new version of #925, which fixes #887.

### Proposed Changes

Add a bulkUpdate flag to workspaceSvg.  Use it to toggle resizeEnabled as well as other behaviours, such as refreshing the toolbox selection.

### Reason for Changes

The toolbox is getting thrashed by relayout every time a new variable is added during a workspace load.  It's better to wait until the end and do a single large update.

### Test Coverage

Tested by running sprinkles, exporting to xml, clearing the workspace, and importing from XML.  It felt faster.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1818,7 +1818,7 @@ Blockly.WorkspaceSvg.prototype.setBulkUpdate = function(enabled) {
   // This will trigger a resize if necessary.
   this.setResizesEnabled(enabled);
   var stoppedUpdating = (this.isBulkUpdating_ && !enabled);
-  this.isBulkUpdating_ = stoppedUpdating;
+  this.isBulkUpdating_ = enabled;
   if (stoppedUpdating) {
     // Refresh the toolbox.
     if (this.toolbox_) {
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -151,6 +151,14 @@ Blockly.WorkspaceSvg.prototype.isMutator = false;
 Blockly.WorkspaceSvg.prototype.resizesEnabled_ = true;
 
 /**
+ * Whether this workspace is in the middle of a bulk update.
+ * Turn on during batch operations for a performance improvement.
+ * @type {boolean}
+ * @private
+ */
+Blockly.WorkspaceSvg.prototype.isBulkUpdating_ = false;
+
+/**
  * Current horizontal scrolling offset in pixel units.
  * @type {number}
  */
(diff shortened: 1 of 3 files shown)


core/workspace_svg.js
```diff
@@ -1797,12 +1806,34 @@ Blockly.WorkspaceSvg.prototype.setResizesEnabled = function(enabled) {
...
+ * removed.
+ * @param {boolean} enabled True if a bulk update is starting, false if a bulk
+ *     update is ending.
+ * @package
+ */
+Blockly.WorkspaceSvg.prototype.setBulkUpdate = function(enabled) {
+  // This will trigger a resize if necessary.
+  this.setResizesEnabled(enabled);
+  var stoppedUpdating = (this.isBulkUpdating_ && !enabled);
+  this.isBulkUpdating_ = stoppedUpdating;
```
paulkaplan: I don't think this will work correctly. if you call `setBulkUpdate(true)` when `isBulkUpdating_` is false, `stoppedUpdating` will be false, so `isBulkUpdating` will never be set to true. Seems like it should be set to `enabled` instead?

paulkaplan (changes requested):

core/workspace_svg.js
rachel-fenichel: Good catch.

paulkaplan (approved):

PR #1123, opened 2017-10-09
Add correct padding on argument reporters

### LINKED ISSUE #1111: Shadow blocks with only labels are too small
### Expected Behavior
The first boolean1 block should be the same height as the second boolean1 block:
image

Same for these two string1 blocks: 
image

### Actual Behavior

The first block is shorter, because it's a shadow.  There are currently no shadow blocks with only labels on the playground.  The argument reporters for custom blocks will be shadows.

The same problem happens for strings as for booleans, but it's less noticeable because the text doesn't run into the edge as clearly.

rachel-fenichel:
### Resolves

#1111 

### Proposed Changes

Moves `shouldDuplicateOnDrag` into Blockly.utils and uses it to decide how much padding a field gets in render.

### Reason for Changes

Argument reporters weren't getting enough padding, because they are single-argument blocks that are shadows.  The padding was changing when the blocks were duplicated and unshadowed, which was very noticeable.

### Test Coverage

Checked visually in the playground.  Other blocks still look right.

image

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -873,7 +873,8 @@ Blockly.BlockSvg.prototype.computeInputWidth_ = function(input) {
  */
 Blockly.BlockSvg.prototype.computeInputHeight_ = function(input, row,
     previousRow) {
-  if (this.inputList.length === 1 && this.isShadow() && this.outputConnection) {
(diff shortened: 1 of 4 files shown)


paulkaplan (approved):
Hm, interesting. I'm learning a lot by reviewing these! Code looks good.

PR #1131, opened 2017-10-11
Context menus on procedure definition blocks

### LINKED ISSUE #1118: Context menu for define blocks
This is necessary for https://github.com/LLK/scratch-blocks/issues/1102

The context menu for the `define` block and the internally rendered custom command block has the following options:

image

delete
add comment
help
edit

There is no duplicate option.

The edit option opens the custom block editor modal.
Delete doesn't work if there are any calls to this function--it pops up a warning instead.

Currently it gives a standard context menu:
image

rachel-fenichel:
### Resolves

#1118 

### Proposed Changes

Gives the `procedure_defnoreturn` block a custom context menu.  It has no duplicate option, but has an edit option.  

Function editing and deletion are not implemented, but there are placeholder callbacks for them.

### Reason for Changes

To make custom procedures work!

### Test Coverage

image

### DIFF
--- a/msg/messages.js
+++ b/msg/messages.js
@@ -109,8 +109,8 @@ Blockly.Msg.HELP = 'Help';
 Blockly.Msg.UNDO = 'Undo';
 /// context menu - Undo the previous undo action.\n{{Identical|Redo}}
 Blockly.Msg.REDO = 'Redo';
-/// context menu - Edit the currently selected function
-Blockly.Msg.EDIT_FUNCTION = 'Edit';
+/// context menu - Edit the currently selected procedure.
+Blockly.Msg.EDIT_PROCEDURE = 'Edit';
 
 // Variable renaming.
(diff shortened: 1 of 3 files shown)


msg/messages.js
```diff
@@ -109,6 +109,8 @@ Blockly.Msg.HELP = 'Help';
 Blockly.Msg.UNDO = 'Undo';
 /// context menu - Undo the previous undo action.\n{{Identical|Redo}}
 Blockly.Msg.REDO = 'Redo';
+/// context menu - Edit the currently selected function
+Blockly.Msg.EDIT_FUNCTION = 'Edit';
```
paulkaplan: Should this be `EDIT_PROCEDURE`, I don't think we use "function" much if at all?

paulkaplan (approved):
One minor nit/question. Other than that LG.

Issue #432, opened 2016-06-17
Re-implement "Move connections at the end of render, instead of in the middle of drawing"


tmickel:
See: https://github.com/google/blockly/pull/389

> Related: This change https://github.com/google/blockly/pull/389/files was merged into scratch-blocks, as I recall, but it didn't really do anything in scratch-blocks because the rendering code has diverged significantly. One of the goals of the change was to make all of the connected blocks move at the same time so we can see the code path better. This would be good to have in scratch-blocks as well.

PR #1132, opened 2017-10-12
Move connections at the end of render

rachel-fenichel:
### Resolves

https://github.com/LLK/scratch-blocks/issues/432

### Proposed Changes

Move all connections once at the end of render, instead of one at a time in the middle of drawing.

### Reason for Changes

Code cleanup.  I did this in Blockly a while ago but we missed it on a merge to scratch-blocks.

### Test Coverage

All blocks seem to still connect in the right places in the playground, including insertion markers.


paulkaplan (approved):
LG

PR #1134, opened 2017-10-12
Tinker with constants for drawing the define block.

rachel-fenichel:
### Resolves

Last piece of https://github.com/LLK/scratch-blocks/issues/1102

### Proposed Changes

Get the size of the hat right on the define block.

### Reason for Changes

Matching the spec.

### Test Coverage

image
image

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -458,7 +458,11 @@ Blockly.BlockSvg.SHAPE_IN_SHAPE_PADDING = {
   }
 };
 
-Blockly.BlockSvg.DEFINE_HAT_CORNER_RADIUS = 4 * Blockly.BlockSvg.CORNER_RADIUS;
+/**
+ * Corner radius of the hat on the define block.
+ * @const
+ */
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
cc @carljbowman I *think* this is right now.

core/block_render_svg_vertical.js
```diff
@@ -1146,7 +1156,7 @@ Blockly.BlockSvg.prototype.renderDrawTop_ = function(steps, rightEdge) {
   /* eslint-disable indent */
   if (this.type == Blockly.BlockSvg.DEFINE_BLOCK_TYPE) {
     steps.push('m 0, 0');
-    steps.push('v', -2 * Blockly.BlockSvg.GRID_UNIT);
+    //steps.push('v', -2 * Blockly.BlockSvg.GRID_UNIT);
```
paulkaplan: Should this line be removed instead of commented out?

paulkaplan (approved):
Code LG (other than that one commented line).

core/block_render_svg_vertical.js
rachel-fenichel: Good catch.  Done.

Issue #1135, opened 2017-10-12
Context menu for custom command blocks


rachel-fenichel:
![image](https://user-images.githubusercontent.com/13686399/31515885-a6730f20-af4b-11e7-997f-c63e2a0b2aa5.png)

6 options, in order:
- [x] duplicate
- [x] delete
- [ ] add comment (tracked in #1120)
- [x] help
- [x] edit (tracked in https://github.com/LLK/scratch-blocks/issues/607)
- [x] define (tracked in https://github.com/LLK/scratch-blocks/issues/1136)

rachel-fenichel:
#1137 added all options except for comments.

PR #1137, opened 2017-10-12
Custom context menu for procedure calls

rachel-fenichel:
### Resolves

https://github.com/LLK/scratch-blocks/issues/1135, except for comments

### Proposed Changes

Add "edit" and "go to definition" options to the context menu for custom command blocks.
Share code for the "edit" option for both types of procedure blocks.

### Reason for Changes

Match the Scratch 2.0 behaviour for custom command blocks.

### Test Coverage

Tested that I get the right alerts when I click 'edit' on both types of procedure blocks and 'go to definition' on custom command blocks.  Tested that `block` is correct in the callbacks, and that the way I'm getting the procCode works.

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -54,13 +54,10 @@ Blockly.Blocks['procedures_callnoreturn'] = {
    * @this Blockly.Block
    */
   init: function() {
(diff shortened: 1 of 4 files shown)


paulkaplan (approved):
LG

Issue #1133, opened 2017-10-12
Duplication is very slow with variable reporters


paulkaplan:
The same problem we had when refreshing the workspace when variable reporters are visible is making duplication of stacks with variable reporters very slow also. Reported over on https://github.com/LLK/scratch-gui/issues/748#issuecomment-336046866

Looks like the `bulkUpdating` flag should be set during that operation as well?

PR #1138, opened 2017-10-12
Enable bulk updating during duplication

rachel-fenichel:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1133

### Proposed Changes

Set the bulk updating flag when duplicating blocks, instead of just disabling workspace resizing.

### Reason for Changes

Duplicating blocks with variable references can lead to toolbox layout thrashing otherwise.

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -682,14 +682,12 @@ Blockly.BlockSvg.prototype.duplicateAndDragCallback_ = function() {
 
       // Create the new block by cloning the block in the flyout (via XML).
       var xml = Blockly.Xml.blockToDom(oldBlock);
-      // The target workspace would normally resize during domToBlock, which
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Problem: this means that workspace resizing is enabled again immediately, because workspace resizing is enable when I turn off bulk update.  That will cause a resizing of the scrollbars.  Probably not as expensive as redoing the toolbox in its current state, but it's not great.

The older way of doing this disabled workspace resizing and left it off until the drag was done, but didn't stop toolbox updates.

paulkaplan (approved):
Great thanks for this. Yeah I see that problem, I guess it would be good to make the bulk updating/resize flags act more like the start/stop cache, where you push/pop instead of enable/disable. But sounds good for now.

Issue #1129, opened 2017-10-11
Argument reporters render wrong when the text is short


rachel-fenichel:
### Expected Behavior

Argument reporter text should be centered, regardless of length.

### Actual Behavior

![image](https://user-images.githubusercontent.com/13686399/31468873-812bb468-ae94-11e7-9dc1-976769a8f148.png)

### Steps to Reproduce

Load this xml in the playground:
```
<xml xmlns="http://www.w3.org/1999/xhtml">
<variables>
</variables>
<block id="]){{Y!7N9ezN+j@Vr`8p" type="procedures_defnoreturn" x="47.175" y="91.96000000000001">
    <statement name="custom_block">
        <shadow type="procedures_callnoreturn_internal">
            <mutation proccode="say %s %n times if %b" argumentnames="[&quot;a&quot;,&quot;this many&quot;,&quot;this is true&quot;]" argumentdefaults="[&quot;&quot;,1,false]" warp="false"/>
        </shadow>
    </statement>
    <next>
        <block id="tIVT2u+!X}IS8z4nu1x9" type="procedures_callnoreturn">
            <mutation proccode="say %s %n times if %b"/>
            <value name="input0">
                <shadow id="V0~M:TxRk0Ua%osjGzh," type="text">
                    <field name="TEXT"/>
                </shadow>
            </value>
            <value name="input1">
                <shadow id="uPx3si(KkbL1)-XpbXoQ" type="math_number">
                    <field name="NUM">1</field>
                </shadow>
            </value>
        </block>
    </next>
</block>
</xml>
```
### Operating System and Browser

Any

rachel-fenichel:
And when loading a non-shadow argument reporter from XML I see this: 
![image](https://user-images.githubusercontent.com/13686399/31468984-e825d662-ae94-11e7-971d-dab36d008259.png)

Even though it's fine if I force it to rerender.

PR #1140, opened 2017-10-13
Fix argument reporter rendering with short text

rachel-fenichel:
### Resolves

https://github.com/LLK/scratch-blocks/issues/1129

### Proposed Changes

Fix argument reporter width when the text is short:
1) give the text area a minimum width
2) use the normal min block width, not the shadow min block width

### Reason for Changes

Fixing #1129 

### Test Coverage

image

is a rendering of the following xml immediately after a load:
```
<xml xmlns="http://www.w3.org/1999/xhtml">
  <variables></variables>
  <block type="argument_reporter_string_number" id="dt3k(J)yIyA4%iipD2#=" x="121" y="1">
    <field name="VALUE">a</field>
  </block>
  <block type="procedures_defnoreturn" id="]){{Y!7N9ezN+j@Vr`8p" x="20" y="138">
    <statement name="custom_block">
      <shadow type="procedures_callnoreturn_internal" id="_m7RK@PTyvbq*RR:+u~^">
        <mutation proccode="do a barrel roll %n times" argumentnames="[&quot;x&quot;]" argumentdefaults="[1]" warp="false"></mutation>
        <value name="input0">
          <shadow type="argument_reporter_string_number" id="@r$9K1DqSuWg%e/)@i*1">
            <field name="VALUE">a</field>
          </shadow>
        </value>
      </shadow>
    </statement>
  </block>
</xml>
```

### DIFF
--- a/core/field_label_editable.js
+++ b/core/field_label_editable.js
@@ -61,3 +61,38 @@ Blockly.FieldLabelEditable.prototype.updateWidth = function() {
   // Unlike the base Field class, this doesn't add space to editable fields.
   this.size_.width = Blockly.Field.getCachedWidth(this.textElement_);
 };
+
+/**
+ * Draws the border with the correct width.
+ * Saves the computed width in a property.
+ * @private
+ */
+Blockly.FieldLabelEditable.prototype.render_ = function() {
+  if (this.visible_ && this.textElement_) {
+    // Replace the text.
+    goog.dom.removeChildren(/** @type {!Element} */ (this.textElement_));
+    var textNode = document.createTextNode(this.getDisplayText_());
+    this.textElement_.appendChild(textNode);
+    this.updateWidth();
+
+    // Update text centering, based on newly calculated width.
+    var centerTextX = this.size_.width / 2;
+
+    // If half the text length is not at least center of
+    // visible field (FIELD_WIDTH), center it there instead.
+    var minOffset = Blockly.BlockSvg.FIELD_WIDTH / 2;
+    if (this.sourceBlock_.RTL) {
+      // X position starts at the left edge of the block, in both RTL and LTR.
+      // First offset by the width of the block to move to the right edge,
+      // and then subtract to move to the same position as LTR.
+      var minCenter = this.size_.width - minOffset;
+      centerTextX = Math.min(minCenter, centerTextX);
+    } else {
+      // (width / 2) should exceed Blockly.BlockSvg.FIELD_WIDTH / 2
+      // if the text is longer.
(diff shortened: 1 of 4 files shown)


core/gesture.js
```diff
@@ -836,8 +836,11 @@ Blockly.Gesture.prototype.duplicateOnDrag_ = function() {
   try {
     // Note: targetBlock_ should have no children.  If it has children we would
     // need to update shadow block IDs to avoid problems in the VM.
+    // Resizes will be reenabled at the end of the drag.
+    this.startWorkspace_.setResizesEnabled(false);
```
paulkaplan, line 840: Is this a different fix?

core/field_label_editable.js
```diff
@@ -61,3 +61,38 @@ Blockly.FieldLabelEditable.prototype.updateWidth = function() {
   // Unlike the base Field class, this doesn't add space to editable fields.
   this.size_.width = Blockly.Field.getCachedWidth(this.textElement_);
 };
+
+/**
+ * Draws the border with the correct width.
+ * Saves the computed width in a property.
+ * @private
+ */
+Blockly.FieldLabelEditable.prototype.render_ = function() {
```
paulkaplan, line 70: So this is very similar to the render function of the base `Field` right? Why doesn't the base class rendering function center the label correctly? Could you explain a bit more about why this was needed?

paulkaplan:
Question in the code. Everything works as expected, I just wanted to learn a bit more about why it was needed.

core/gesture.js
rachel-fenichel: Yup, just avoiding a resize during the user action.

core/field_label_editable.js
rachel-fenichel: The base class has the following code around the centering:
```

    // In a text-editing shadow block's field,
    // if half the text length is not at least center of
    // visible field (FIELD_WIDTH), center it there instead,
    // unless there is a drop-down arrow.
    if (this.sourceBlock_.isShadow() && !this.positionArrow) {
```

So it won't center the label correctly if it's not a shadow block.  The check in the base class seems really suspect to me anyway--it claims to be for a "text-editing shadow block's field", and `Field` is probably the wrong place for that code.  It could probably go somewhere much more specific.

Rather than add more checks in Field, I just overwrote this to get the centering right for this particular field.  I removed some code that I don't need--mostly around the arrow and the box.

paulkaplan: Ok sounds good.

paulkaplan (approved):

Issue #879, opened 2017-05-05
Performance: opening the toolbox


rachel-fenichel:
Switching toolbox categories takes far longer than it ought to, given that we have a mostly static set of blocks.  This isn't really surprising since we've never tried to optimize that code.

Some ideas and potential low hanging fruit: 
- Better caching of computed text length for strings
- Avoid spurious rerendering of fields
- For fully static categories, cache block coordinates instead of needing a layout pass
  - This might not actually be an improvement; I need to look into it.

thisandagain:
@rachel-fenichel and @picklesrus is this something you folks could help us with?

rachel-fenichel:
Yes, it's something that I'm thinking about but don't have any patches yet.

picklesrus:
Yeah, as @rachel-fenichel mentioned, we haven't tried to optimize it and there is likely lots of low hanging fruit.  It might be a while til I could look at this in depth, but I'll try and keep it top of mind. Here's one thing I noticed while in the code today:

Flyout's getMetrics calls getBBox() on the block canvas which, iirc, causes a layout pass. We should be able to replace that with something like getBlocksBoundingBox which uses cached blocks sizes.  I haven't looked into how often we call Flyout's getMetrics though so it might just be a minor win.

samelhusseini:
We took a brief look into the "spurious rerendering of fields". It looks like field::init, the size calculations are causing the field render_ function to be called three times (once in getSize, once in render(), and once because the width is set to zero).

Although, this all looks intentional..

picklesrus:
Minor update

Not calling getComputedTextLength all.the.time. is still the biggest win we can get. Plus it is applicable across almost every block/field. 

Despite that, I dove down some other rabbit holes anyway so here are some notes:

I tested out the previously mentioned BBox change with #1059  patched in and it does remove a re-calculate style on every animation frame when choosing a new category. It should also remove one on every mouse move while scrolling within the flyout as well as anytime we call flyout.show. I have a few details (vertical vs horizontal, rtl, etc.) to sort out, but they seem straightforward enough that I should just push ahead and do it.  

These things also struck me as odd:
- Why does disposing of a flyout category worth of blocks take 1/2 the frame budget (8ms) despite there being no apparent redraws/reflows happening? (disclaimer: these numbers come from my aging, fan running, macbook air)
[...]

PR #1141, opened 2017-10-13
Block rendering performance

paulkaplan:
### Resolves

Implements the suggestions from the end of https://github.com/LLK/scratch-blocks/issues/879

On my machine, knocks the time for `workspace.refreshToolboxSelection_` down from 200ms to around 140ms. ¯\_(ツ)_/¯

### Proposed Changes

The various changes are in individual commits. They are pretty unrelated to each other, so I wouldn't be opposed to splitting them into different PRs. Individually they are only a couple lines each. 

### Reason for Changes

Rendering the full flyout of blocks, which has to be done fairly frequently, can take a very long time. We need it to... not.

### Test Coverage

Rendering appears unchanged in the playground. 

Before 
image

After

screen shot 2017-10-13 at 3 34 44 pm

No more red!

### DIFF
--- a/core/field_dropdown.js
+++ b/core/field_dropdown.js
@@ -122,6 +122,7 @@ Blockly.FieldDropdown.prototype.init = function() {
   });
   this.arrow_.setAttributeNS('http://www.w3.org/1999/xlink',
       'xlink:href', Blockly.mainWorkspace.options.pathToMedia + 'dropdown-arrow.svg');
+  this.className_ += ' blocklyDropdownText';
 
   Blockly.FieldDropdown.superClass_.init.call(this);
   // If not in a shadow block, draw a box.
@@ -431,11 +432,6 @@ Blockly.FieldDropdown.prototype.setText = function(text) {
   this.updateTextNode_();
 
   if (this.textElement_) {
-    // Update class for dropdown text.
-    // This class is reset every time updateTextNode_ is called.
-    this.textElement_.setAttribute('class',
-        this.textElement_.getAttribute('class') + ' blocklyDropdownText'
-    );
     this.textElement_.parentNode.appendChild(this.arrow_);
   }
   if (this.sourceBlock_ && this.sourceBlock_.rendered) {
--- a/core/flyout_button.js
+++ b/core/flyout_button.js
@@ -162,7 +162,7 @@ Blockly.FlyoutButton.prototype.createDom = function() {
       this.svgGroup_);
   svgText.textContent = this.text_;
 
-  this.width = svgText.getComputedTextLength();
+  this.width = Blockly.Field.getCachedWidth(svgText);
 
   if (!this.isLabel_) {
     this.width += 2 * Blockly.FlyoutButton.MARGIN;
--- a/core/flyout_button.js
+++ b/core/flyout_button.js
@@ -178,8 +178,6 @@ Blockly.FlyoutButton.prototype.createDom = function() {
   svgText.setAttribute('x', this.width / 2);
   svgText.setAttribute('y', this.height / 2);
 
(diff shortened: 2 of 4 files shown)


core/flyout_base.js
```diff
@@ -610,6 +610,7 @@ Blockly.Flyout.prototype.stepScrollAnimation = function() {
  * @private
  */
 Blockly.Flyout.prototype.clearOldBlocks_ = function() {
+  Blockly.Events.disable();
```
rachel-fenichel: If you don't need events when clearing the blocks, you also don't need them when creating the blocks.  In this PR or a followup, wrap block creation (in `show`) in a disable/enable pair.

core/flyout_base.js
```diff
@@ -627,6 +628,7 @@ Blockly.Flyout.prototype.clearOldBlocks_ = function() {
     button.dispose();
   }
   this.buttons_.length = 0;
+  Blockly.Events.disable();
```
rachel-fenichel: This should be `enable()`.

Wrap everything between `disable` and here in a try/catch, and reenable events in a finally clause.  If something goes wrong you don't want to end up with events perma-disabled.

core/flyout_button.js
```diff
@@ -178,8 +178,6 @@ Blockly.FlyoutButton.prototype.createDom = function() {
   svgText.setAttribute('x', this.width / 2);
   svgText.setAttribute('y', this.height / 2);
 
-  this.updateTransform_();
```
rachel-fenichel, line 181: You're removing this because position is always (0, 0) here, right?

core/flyout_button.js
```diff
@@ -162,7 +162,7 @@ Blockly.FlyoutButton.prototype.createDom = function() {
       this.svgGroup_);
   svgText.textContent = this.text_;
 
-  this.width = svgText.getComputedTextLength();
+  this.width = Blockly.Field.getCachedWidth(svgText);
```
rachel-fenichel, line 165: Please make this change in Blockly as well.

core/tooltip.js
```diff
@@ -248,7 +248,9 @@ Blockly.Tooltip.hide = function() {
       Blockly.Tooltip.DIV.style.display = 'none';
     }
   }
-  clearTimeout(Blockly.Tooltip.showPid_);
+  if (Blockly.Tooltip.showPid_) {
```
rachel-fenichel, line 251: Please make this change upstream as well.

core/field_dropdown.js
```diff
@@ -122,6 +122,7 @@ Blockly.FieldDropdown.prototype.init = function() {
   });
   this.arrow_.setAttributeNS('http://www.w3.org/1999/xlink',
       'xlink:href', Blockly.mainWorkspace.options.pathToMedia + 'dropdown-arrow.svg');
```
rachel-fenichel, line 124: This looks good.  Please make this change upstream as well.

rachel-fenichel (commented):
Everything but the disable looks good, and some of these changes should definitely go upstream.

core/flyout_base.js
paulkaplan: Actually I'm going to leave this out for now, I have the feeling some of the monitor related stuff may depend on events from the flyout.

core/flyout_base.js
paulkaplan: Fixed by removing that piece.

core/field_dropdown.js
paulkaplan: I do not think this applies to blockly, which seems to use the same class name on dropdown fields as other fields.

rachel-fenichel (approved):
LG!

PR #1144, opened 2017-10-14
Fix tests

rachel-fenichel:
Fix tests, and rebuild (caused changes in en.js and en.json).

Before: 27/121 failing.
Now: 120/120 passing.
One test deleted because it tests a feature that scratch-blocks doesn't have (limiting the number of blocks on the workspace).
Most tests just needed tweaks to continue working.

I did discover that blockly and scratch-blocks do different things when calling `setValue(undefined)` on a `FieldVariable`.  In blockly the value ends up being `"undefined"` because it calls `Field.setText`.  In scratch-blocks it's `undefined` because it's calling `FieldDropdown.setText`.


paulkaplan (approved):
Great. For posterity, this only fixes the `vertical_tests.html` way of running the tests. 

@rschamp it would be great for us to get this running on travis. We'll have to update the `test_runner.js` to point to the vertical/horizontal tests, and update the travis.yml/package.json.

Issue #1143, opened 2017-10-13
When a procedure definition is changed, find and update all its callers


rachel-fenichel:
When a procedure definition is edited (see https://github.com/LLK/scratch-blocks/issues/607) all of its callers should be updated to use the new definition.

`Blockly.Procedures.mutateCallers` is the function that does this, but it's currently implemented for Blockly procedures rather than scratch-blocks procedures.

Issue #1130, opened 2017-10-11
Custom procedure deletion


rachel-fenichel:
### Expected Behavior

Scratch 2.0 behaviour:

When to delete:
- [x] Right-clicking on a custom procedure definition shows a "delete" option:
![image](https://user-images.githubusercontent.com/13686399/31470970-50a8e4ee-ae9b-11e7-880a-d3515ffebaad.png)
- [x] If there are no calls to the custom procedure, or if all of the calls are inside the definition, the procedure is deleted.
![image](https://user-images.githubusercontent.com/13686399/31471059-c854ae10-ae9b-11e7-9e34-2e911b1ddb6d.png)
_This function may be deleted_
- [x] If there are any calls to the custom procedure it cannot be deleted.  An error is shown:
![image](https://user-images.githubusercontent.com/13686399/31471099-e7e7a46c-ae9b-11e7-929e-17ecb3b19037.png)

- [ ] Right-clicking on a custom command block shows a "delete" option.  Selecting that option deletes the selected block, but not the custom procedure definition:
![image](https://user-images.githubusercontent.com/13686399/31470977-63b43534-ae9b-11e7-9e7a-989d80685013.png)

How to delete:
- [x] The define block and all of its children are removed from the workspace.
- [ ] The call block is removed from the toolbox.
- [ ] Some work is done in the VM (https://github.com/LLK/scratch-vm/issues/715)

thisandagain:
Tracking remaining work in #1287

PR #1146, opened 2017-10-18
Find all callers of a procedure by procCode

rachel-fenichel:
### Resolves

https://github.com/LLK/scratch-blocks/issues/1143
https://github.com/LLK/scratch-blocks/issues/1130

This finds (but does not update or delete) callers of a custom procedure.

### Proposed Changes

Finds callers of a custom procedure based on `procCode`, and returns a list.
`findCallers` now takes a `definitionRoot` and optionally skips the stack starting with `definitionRoot`--behaviour needed for #1130.  

Adds tests for `getCallers`.
Reorders test loading (alphabetical order).

### Reason for Changes
Part of implementing custom procedure deletion and updates.

### Test Coverage

`procedure_test.js`

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -60,6 +60,15 @@ Blockly.Blocks['procedures_callnoreturn'] = {
     this.procCode_ = '';
   },
   /**
+   * Returns the name of the procedure this block calls, or the empty string if
+   * it has not yet been set.
+   * @return {string} Procedure name.
+   * @this Blockly.Block
+   */
+  getProcCode: function() {
+    return this.procCode_;
+  },
+  /**
    * Create XML to represent the (non-editable) name and arguments.
    * @return {!Element} XML storage element.
    * @this Blockly.Block
--- a/tests/jsunit/horizontal_tests.html
+++ b/tests/jsunit/horizontal_tests.html
@@ -7,7 +7,7 @@
     <script>goog.require('goog.testing.jsunit');</script>
   </head>
   <body>
-    <script src="utils_test.js"></script>
+    <script src="test_utilities.js"></script>
     <script src="blockly_test.js"></script>
     <script src="block_test.js"></script>
     <script src="connection_test.js"></script>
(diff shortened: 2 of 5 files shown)


tests/jsunit/horizontal_tests.html
```diff
@@ -7,7 +7,7 @@
     <script>goog.require('goog.testing.jsunit');</script>
   </head>
   <body>
-    <script src="utils_test.js"></script>
+    <script src="test_utilities.js"></script>
     <script src="blockly_test.js"></script>
```
paulkaplan, line 11: Not related to this PR, but just noticed this file doesn't actually exist.

blocks_vertical/procedures.js
```diff
@@ -60,6 +60,15 @@ Blockly.Blocks['procedures_callnoreturn'] = {
     this.procCode_ = '';
   },
   /**
+   * Returns the name of the procedure this block calls, or the empty string if
+   * it has not yet been set.
+   * @return {string} Procedure name.
+   * @this Blockly.Block
+   */
+  getProcCode: function() {
+    return this.procCode_;
```
paulkaplan, line 69: Question: I see procCode is defined in the mutation XML as well as stored in this variable. Which is the "source of truth"? What is the lifecycle of the mutation XML? Is that only used when loading/serializing the block?

paulkaplan (approved):
Code looks good, just had one question.

tests/jsunit/horizontal_tests.html
rachel-fenichel: Yeah.  I removed it in vertical_tests in the other testing PR.  I'm not sure what happened there.

blocks_vertical/procedures.js
rachel-fenichel: https://developers.google.com/blockly/guides/create-custom-blocks/web/mutators

The XML is used when loading and serializing the block.  We don't keep a handle on it while the block is alive.  In this case the XML is how scratch-vm is telling scratch-blocks what to load.

Issue #1151, opened 2017-10-18
Show an error when a custom procedure can't be deleted


rachel-fenichel:
In Scratch 2.0 it looks like this:
![image](https://user-images.githubusercontent.com/13686399/31747811-d94dd376-b423-11e7-8ad7-6924086847f7.png)

PR #1152, opened 2017-10-19
Feature/delete procedure

rachel-fenichel:
### Resolves

https://github.com/LLK/scratch-blocks/issues/1130

### Proposed Changes

Add a callback to check for uses of the custom procedure, and delete it if allowed.

This just uses an alert for the message when a procedure can't be deleted.  Implementing that is #1151.

### Reason for Changes

Implementing custom procedure deletion.

### Test Coverage

Tests added in `procedure_test.js`


paulkaplan (approved):
Looks good. Tested with GUI as well, all works as it should.

PR #1153, opened 2017-10-19
Move connection to the correct location

### LINKED ISSUE #1150: Define block overlaps at the bottom
Looks like a small rendering issue with the define block, the bottom edge seems to be farther down than it should be/the connection is rendered higher than it should be?

define-render-issue

rachel-fenichel:
### Resolves

Fixes #1150 

### Proposed Changes

Add a bit more height to the row with the `custom_block`.

### Reason for Changes

Row height was wrong for this special-case block.

### Test Coverage

Tested in the playground.


paulkaplan (approved):
LG thanks!

Issue #1127, opened 2017-10-10
Internally rendered custom command blocks can be overwritten


rachel-fenichel:
### Expected Behavior

The internal blocks in custom procedure definitions shouldn't be replaceable (move blocks in this image).

![image](https://user-images.githubusercontent.com/13686399/31415832-a48094c0-adda-11e7-94cf-f0deba274106.png)

### Actual Behavior

![image](https://user-images.githubusercontent.com/13686399/31415842-b3ca93fe-adda-11e7-8bb2-1f2dfcdbed24.png)

The problem occurs because all of the previous and next connections on existing blocks have the `null` type.  Adding a type check on the `custom_block` input is insufficient because the null type can connect to anything.

### Solutions
One solution is to add type information to every single next and previous connection, and give the define block's `custom_block` connection a different type.

I don't like that solution because if anyone fails to set their corrections correctly (e.g. when making blocks for extensions) they will see this problem reemerge.

Another solution is to add a `strict` or `no-wildcard` tag to connections, and use that tag here.

paulkaplan:
@rachel-fenichel is the plan to wait on the upstream feature needed for this?

rachel-fenichel:
I had been planning to chat about it in the meeting this week, which I forgot was cancelled.

What are your thoughts on the simple solution that I mentioned above?

> One solution is to add type information to every single next and previous connection, and give the define block's custom_block connection a different type.

paulkaplan:
Well, I think the concern you listed is a big one. Particularly because we don't want people to be able to define blocks that would be able to be dropped into there. I would lean more towards the second option.

rachel-fenichel:
I lean that way as well, but I don't have a good way of adding it to the JSON spec.  I'm going to chat with Neil about that today.  It's easy to add for inputs, but hard to add to previous and next connections.

rachel-fenichel:
Whether that second concern is a real problem depends in large part on how you're planning to have people define blocks for extensions, etc.

paulkaplan:
Closed by https://github.com/LLK/scratch-blocks/pull/1158

PR #1158, opened 2017-10-19
Hack to fix 1127 while we consider strict connection types

rachel-fenichel:
### Resolves

https://github.com/LLK/scratch-blocks/issues/1127

### Proposed Changes

Adds a check to `connection.canConnectWithReason_`, a corresponding error code, and a set of tests.

### Reason for Changes

#1127 is confusing to users, and it'll be a bit before we have a better solution in place upstream.

### Test Coverage

Added tests to connection_test.js


paulkaplan (approved):
Looks good.

RoboErikG (commented):
Looks reasonable as a temp workaround to me.

PR #1162, opened 2017-10-20
Implement isCurrentlyEditable to return false for FieldVariableGetter…

### LINKED ISSUE #1160: Clicking variable reporters doesn't emit 'stackclick' event
### Expected Behavior

Clicking on a regular reporter in the vertical playground with event logging on shows 3 events, including the `stackclick` event that is used by the VM to run a block. 
```
Blockly.Events.Ui {group: "", recordUndo: false, element: "selected",....
Blockly.Events.Ui {blockId: "#E+{@fR+6VwE.AD[Q6!1", ....
Blockly.Events.Ui {blockId: "#E+{@fR+6VwE.AD[Q6!1", ....
```

### Actual Behavior

Creating a new variable and clicking the reporter in the flyout (or the workspace) only emits a single event and no `stackclick`. 

```
Blockly.Events.Ui {group: "", recordUndo: false, element: "selected", ...
```

---

This is the cause of variable reporters not showing their values when they are clicked. It was tough to track down because click-to-report-value is broken due to other issues. But the variables were broken long before that, and this seems to be the cause.

paulkaplan:
… and FieldLabelEditable

### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/1160

### Proposed Changes

Implement `isCurrentlyEditable` to return false for FieldVariableGetter and FieldLabelEditable, to make it emit events like a block instead of an editable field. 

### Reason for Changes

Those two types of blocks are "editable" in that they are serialized, but don't have editors and shouldn't be treated like regular editable fields. 

### Test Coverage

Checked the events in the playground.

### DIFF
--- a/core/field_label_editable.js
+++ b/core/field_label_editable.js
@@ -96,3 +96,12 @@ Blockly.FieldLabelEditable.prototype.render_ = function() {
     this.textElement_.setAttribute('x', centerTextX);
   }
 };
+
+
+/**
+ * Click events should not treat this like an editable field.
+ * Suppress default editable field behaviour.
+ */
+Blockly.FieldLabelEditable.prototype.isCurrentlyEditable = function() {
+  return false;
+};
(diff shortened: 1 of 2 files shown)


rachel-fenichel (approved):

rachel-fenichel:
Actually, don't submit this.  I'll fix #1163 today, and we can avoid merge conflicts.

Issue #1163, opened 2017-10-20
Stop using "Editable" when we mean "Serializable"


paulkaplan:
Following up with https://github.com/LLK/scratch-blocks/pull/1162

We use "editable" to enable serialization, but it causes other problems like https://github.com/LLK/scratch-blocks/issues/1160. 

Implement a `serializable` flag and stop using `EDITABLE`.

PR #1165, opened 2017-10-20
Add a serializable attribute to fields.

rachel-fenichel:
### Resolves

https://github.com/LLK/scratch-blocks/issues/1163

### Proposed Changes

Adds a `SERIALIZABLE` attribute to fields, and uses it when a field should be saved to XML but should only be programmatically editable.

Renames `FieldLabelEditable` to `FieldLabelSerializable`.

### Reason for Changes

The way we were using `EDITABLE` was confusing and didn't distinguish between 'has an editor ui' and 'can be programmatically edited and should be saved, but has no editor ui'.

### Test Coverage

Updated tests in `field_variable_getter_test.js` and added tests in `xml_test.js`


picklesrus (commented):
Sorry, I don't have a lot of context here. I'm actually a little confused. Why do we need a new class? Wouldn't the member on the field be sufficient?

IIUC, this also means that every field of a type is either serializable or not, regardless of what block it was defined to be a part of. Is that right?   Is that what we want?

picklesrus:
Hah. I misread the file rename as a a copy and the SERIALIZABLE as not being on the prototype. So yeah, I was indeed confused.

LGTM.

PR #1166, opened 2017-10-21
Do not refresh the toolbox when a new variable is created on the work…

picklesrus:
…space but it already exists in the variable map.  This gets rid of some calls to Toolbox.refreshSelection. e.g. when you drag a variable block out of the toolbox, we were refreshing the toolbox.  It will also help with xml loading for workspaces using multiple of the same variable reporter.  
### Resolves

More needs to be done to solve #879, but this helps a little.

### Proposed Changes
Stop some of our overaggressive updating of the toolbox.

### Reason for Changes

Speed!

### Test Coverage

I wrote 2 tests. woo!

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1076,9 +1076,14 @@ Blockly.WorkspaceSvg.prototype.deleteVariableById = function(id) {
  * @package
  */
 Blockly.WorkspaceSvg.prototype.createVariable = function(name, opt_type, opt_id) {
+  var variableInMap = (this.getVariable(name) != null);
   var newVar = Blockly.WorkspaceSvg.superClass_.createVariable.call(this, name,
     opt_type, opt_id);
-  this.refreshToolboxSelection_();
+  // For performance reaons, only refresh the the toolbox for new variables.
+  // Variables that already exist should already be there.
+  if (!variableInMap) {
+    this.refreshToolboxSelection_();
+  }
   return newVar;
 };
 
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1079,7 +1079,7 @@ Blockly.WorkspaceSvg.prototype.createVariable = function(name, opt_type, opt_id)
   var variableInMap = (this.getVariable(name) != null);
   var newVar = Blockly.WorkspaceSvg.superClass_.createVariable.call(this, name,
     opt_type, opt_id);
-  // For performance reaons, only refresh the the toolbox for new variables.
+  // For performance reasons, only refresh the the toolbox for new variables.
   // Variables that already exist should already be there.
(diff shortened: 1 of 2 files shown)


core/workspace_svg.js
```diff
@@ -1076,9 +1076,14 @@ Blockly.WorkspaceSvg.prototype.deleteVariableById = function(id) {
  * @package
  */
 Blockly.WorkspaceSvg.prototype.createVariable = function(name, opt_type, opt_id) {
+  var variableInMap = (this.getVariable(name) != null);
   var newVar = Blockly.WorkspaceSvg.superClass_.createVariable.call(this, name,
```
rachel-fenichel, line 1080: Do you still need to call the superclass `createVariable` if you've already determined that the variable exists?

rachel-fenichel (approved):
One question.

core/workspace_svg.js
picklesrus: Ah, good question.  I don't quite have time to test at the moment, but I originally left it in because when you create a variable through the button this gets called and I wanted to make sure we don't suppress the duplicate message when you try and create a variable by the same name.  But maybe that error gets caught in the prompt. I'll look into it.

picklesrus: Well, at least the case where the user creates a variable through the create button is handled so I could skip this.

There is still at least 1 behavior change though.  Right now importing an existing variable from xml fails.  If I hide this behind the check it won't.  So we'd have to decide that we actually want that.

core/workspace_svg.js
```diff
@@ -1076,9 +1076,14 @@ Blockly.WorkspaceSvg.prototype.deleteVariableById = function(id) {
...
     opt_type, opt_id);
-  this.refreshToolboxSelection_();
+  // For performance reaons, only refresh the the toolbox for new variables.
```
paulkaplan: *reasons (typo)

paulkaplan:
This may be part of the next fix, but do we also want to add a flag to manually disable `refreshSelection`, or a new method that does the same, so that we can call `createVariable` from `XML#domToVariables` to take care of the importing from XML part?

picklesrus:
@rachel-fenichel and I chatted yesterday and came to the conclusion we should let this go as is since calling super.createVariable inside the conditional I added actually changes the behavior.  I'll file an issue to discuss what should happen in that case since we weren't sure whether what does happen is intentional or not.

@paulkaplan: Yep, fixing the xml loading is the next fix.  In my test project that has a decent number of variables, it should change the load time by an order of magnitude. e.g. ~5sec -> 500ms.  We want to clarify the strategy for the gui/vm/scratch-blocks interaction in our sync today, then I'll do the implementation.

PR #1167, opened 2017-10-21
Create a playground for working on custom procedure editing.

rachel-fenichel:
### Resolves

None.  This is to help debug.

### Proposed Changes

Add a two-workspace playground for working on custom procedure editing.  The workspace on the left will be the custom procedure editing workspace; the workspace on the right will hold the resulting procedure definition and call blocks.

### Reason for Changes

Dev tools.

### Test Coverage

Opened the playground to verify that it loads.

### DIFF
--- a/tests/custom_procedure_playground.html
+++ b/tests/custom_procedure_playground.html
@@ -1,0 +1,69 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):
Looks like a good start. Excited to have a place to see this develop.

Issue #1145, opened 2017-10-16
Single call to update workspace XML?


paulkaplan:
In order to change sprites in scratch-gui, we have to clear the workspace and load up new blocks from the new sprite. Right now that involves these three things.
```javascript
1:    workspace.clear()
2:    dom = XML.textToDom(xml)
3:    XML.domToWorkspace(dom, workspace)
```

Line 1 and 3 both cause `workspace.refreshSelection_` to be called, through the toggling of the bulkUpdate flag. That operation costs around 200ms right now, so it would really help to condense the process down to a single bulkUpdate toggle. 

@rachel-fenichel thoughts/concerns on how to do this? One option would be to change `setBulkUpdating` to push/pop the way the field cache does, instead of toggle on/off. Then we could call setBulkUpdating(false) before everything and leave the 3 calls the same. Another option would be to create a single method that wraps the clear -> load xml process. It seems like it would go in `XML`, but because of the `bulkUpdate` toggles living in `workspace`, it would need to change there as well.

rachel-fenichel:
Yes, I'm starting to think that setBulkUpdating should be pushed and popped.  Note that we need to catch any exceptions that might happen while pushing it, or else we can lock the system in a state where it never updates again.

picklesrus:
I think this is fixed now that scratch-gui/develop is using the function added in #1178. If there is more optimization to be done I think there will be more targeted so we can track those separately.

PR #1168, opened 2017-10-21
Revert #1122 and changes made on top of it.  

picklesrus:
See discusson on #748 in scratch-gui and #1145 .  We will fix this a different way next week.

### Resolves
Not a resolution, but see discussion on https://github.com/LLK/scratch-gui/issues/748

### Proposed Changes

Reverts #1122, though mostly by hand because of conflicts.

### Reason for Changes

The bulk update change had the right idea, but it causes some other problems.


paulkaplan:
Heads up, I added a commit to fix failing travis build due to linting on trailing spaces.

paulkaplan (approved):

PR #1172, opened 2017-10-23
Use "Color" instead of "Hue" as the slider color input.

### LINKED ISSUE #1171: Change 'hue' to 'color' in color picker
We've decided to use 'color' instead of 'hue in the 'pen' blocks because it may be easier to understand. 

It would be better if the picker was consistent with the blocks.

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/1171

### Proposed Changes

image

### Reason for Changes

The word "color" matches peoples expectations and is more widespread than the word "hue". We are changing the color components in the pen blocks to match. There is obviously a bit of "overlap" since we are referring to both the color as a whole and the hue component as "color". We'll see if that causes confusion.

### DIFF
--- a/core/field_colour_slider.js
+++ b/core/field_colour_slider.js
@@ -272,7 +272,7 @@ Blockly.FieldColourSlider.prototype.showEditor_ = function() {
   this.saturation_ = hsv[1];
   this.brightness_ = hsv[2];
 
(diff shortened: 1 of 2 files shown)


chrisgarrity (approved):
Looks good! [This is also good to know for localization]

PR #1177, opened 2017-10-24
Add buttons to custom procedure playground

rachel-fenichel:
### Resolves

Work for custom procedure editing.

### Proposed Changes

Add some buttons to the custom procedure playground, and some blocks:
image

### Reason for Changes

Work for custom procedure editing.

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -317,3 +317,16 @@ Blockly.Blocks['argument_reporter_string_number'] = {
     });
   }
 };
+
+Blockly.Blocks['procedures_mutator_root'] = {
+  /**
+   * Block for calling a procedure with no return value.
+   * @this Blockly.Block
+   */
+  init: function() {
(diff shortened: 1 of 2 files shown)


picklesrus (commented):
LGTM

blocks_vertical/procedures.js
```diff
@@ -317,3 +317,16 @@ Blockly.Blocks['argument_reporter_string_number'] = {
     });
   }
 };
+
+Blockly.Blocks['procedures_mutator_root'] = {
+  /**
+   * Block for calling a procedure with no return value.
```
paulkaplan, line 323: This comment doesn't seem right?

rachel-fenichel: Yeah.  In my next PR I'll update it to "The root block in the procedure editing workspace."

The extensions are also wrong--right now this is just a minimal placeholder block.

PR #1178, opened 2017-10-24
Add an xml utility to clear the workspace and then load xml into the …

picklesrus:
…dom.  In it, use a new toolbox refresh mechanism that allows for batch updates.  It mirrors the setResizesEnabled mechanism for managing batch updates to the workspace.

### Resolves

Adds scratch-blocks api as a toward resolving #1145
### Proposed Changes

Add a new xml utility that clears the workspace and loads from xml in one go.  This gives us the ability to batch up all the changes before doing things like updating the toolbox.

### Reason for Changes

Performance improvement to avoid refreshing the toolbox (which re-renders all the blocks) over and over again.
### Test Coverage

None for now.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -151,6 +151,14 @@ Blockly.WorkspaceSvg.prototype.isMutator = false;
 Blockly.WorkspaceSvg.prototype.resizesEnabled_ = true;
 
 /**
+ * Whether this workspace has toolbox/flyout refreshes enabled.
+ * Disable during batch operations for a performance improvement.
+ * @type {boolean}
+ * @private
+ */
+Blockly.WorkspaceSvg.prototype.toolboxRefreshEnabled_ = true;
+
+/**
  * Current horizontal scrolling offset in pixel units.
  * @type {number}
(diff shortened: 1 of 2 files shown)


picklesrus:
@paulkaplan @rachel-fenichel Let me know what you think. I don't like the name of the new function, so recommendations encouraged!

There are lots more places this could (and should) be applied (e.g. undo/redo, within domToWorkspace, etc.) but I think, once called from the gui, this is the biggest win for switching sprites so I figured I'd check it in so it could start being used.

picklesrus:
Oh, and to throw some numbers on this... On a macbook air, clearing the blocks from a "sprinkles" click and loading the blocks from the monster sprite of project 154860340 goes from ~5sec to ~500ms.

rachel-fenichel:
LGTM

paulkaplan (approved):
Looks good. And confirmed by integrating with the GUI that it calls refreshSelection exactly once. This PR brings it into the GUI to get that massive speedup. 

https://github.com/LLK/scratch-gui/pull/807/files

PR #1180, opened 2017-10-25
Fix comment and remove extension of the root block for custom procedure editing

rachel-fenichel:
The old version was inaccurate.

This block is mostly just a placeholder right now anyway.


paulkaplan (approved):

picklesrus (commented):
LGTM

PR #1184, opened 2017-10-25
Add the new clearWorkspaceAndLoadFromXml to the set of globals so the…

picklesrus:
… compiler does not rename it.  This should have been part of #1178

### Proposed Changes

Following up on #1178 to make sure the clearWorkspaceAndLoadFromXml doesn't get renamed by the compiler.

### DIFF
--- a/core/xml.js
+++ b/core/xml.js
@@ -317,6 +317,7 @@ Blockly.Xml.clearWorkspaceAndLoadFromXml = function(xml, workspace) {
   workspace.setToolboxRefreshEnabled(true);
   return blockIds;
 };
+
 /**
  * Decode an XML DOM and create blocks on the workspace.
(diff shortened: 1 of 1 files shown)


picklesrus:
@paulkaplan I expect you might need this before you can use the new Xml function from scratch-gui since it is likely to be renamed.

paulkaplan (approved):
Hmm maybe we are using some kind of different compilation settings because I already deployed that to https://llk.github.io/scratch-gui/develop/ and it seems to work?

rachel-fenichel:
It's possible to have it work by accident, but I think this change guarantees that it won't randomly break when the compiler changes something internally.

paulkaplan:
Ah ok sounds good!

Issue #1116, opened 2017-10-06
Auto-scroll category menu to prevent selected category from being hidden


ericrosenbaum:
### Expected Behavior

When the category menu is long enough to extend beyond the view, it should automatically scroll itself to make sure the currently selected category is visible. This should occur anytime the blocks flyout scrolls or auto-scrolls, including wheel/drag/scrollbar events, clicking a category menu button (such as one that is part-way off screen), or when an extension category is added.

### Actual Behavior

The category menu does not auto-scroll and categories extending below the view are hidden.

### Steps to Reproduce

Shorten the window height to hide the bottom-most category.

PR #1198, opened 2017-10-31
Tighten up the category menu

ericrosenbaum:
### Resolves

This is a workaround for https://github.com/LLK/scratch-blocks/issues/1116

### Proposed Changes

This change provides more blank space below the vertical category menu items, for adding extensions. It also makes the category menu a bit narrower. The goal is to allow room on a 768px high screen for at least two extensions (after we add the "my blocks" category). 

This change also requires a gui PR to shrink the "add extension" button to match: https://github.com/LLK/scratch-gui/pull/826

### Reason for Changes

We wanted to have enough vertical space in the category menu so that it is not necessary to scroll to see all its items, even after adding two extensions, on relatively small screens (such as 768px high).

### Test Coverage

No tests.

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -1111,10 +1111,10 @@ Blockly.Css.CONTENT = [
   '}',
 
   '.scratchCategoryMenu {',
-    'width: 60px;',
+    'width: 3.25rem;',
     'background: $colour_toolbox;',
     'color: $colour_toolboxText;',
-    'font-size: .7em;',
+    'font-size: .7rem;',
     'user-select: none;',
     '-webkit-user-select: none;',
     '-moz-user-select: none;',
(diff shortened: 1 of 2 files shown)


core/css.js
```diff
@@ -1111,10 +1111,10 @@ Blockly.Css.CONTENT = [
   '}',
 
   '.scratchCategoryMenu {',
-    'width: 60px;',
```
paulkaplan, line 1114: I believe there is some javascript that also uses this 60px number for calculations. I think I'd rather keep the number in px instead of rem to make that correspondence easier to see. If 3.25rem != 60px, can you also change the other place where `60` is used to in the JS?

paulkaplan (changes requested):
Small but important change requested.

paulkaplan (approved):

core/css.js
ericrosenbaum: Thanks- got it!

Issue #1200, opened 2017-11-01
Argument defaults on custom procedures


rachel-fenichel:
Custom procedure mutations have both names and default values for arguments.  There is no UI for setting default values.

In code, should I respect those default values or ignore them?
In UI, are we going to have a way to set default values?

rachel-fenichel:
Friendly ping--I'm starting to need this.

Should I care about argument defaults at all?  And should I expect that all mutations have an argumentDefaults array, or just be prepared to parse one if I find it?

Issue #1185, opened 2017-10-26
Procedure call blocks can't handle _updateDisplay being called multiple times


rachel-fenichel:
They never clear out the mutated fields before adding them again, so the number of fields will only increase.

Whoops.

rachel-fenichel:
Fixed by #1225

PR #1203, opened 2017-11-01
Handle mutating procedure calls multiple times

rachel-fenichel:
### Resolves

Part of #1185

procedure_callnoreturn blocks can now be updated multiple times, but procedure_callnoreturn_internal blocks cannot.

### Proposed Changes

- Moves procedure functions into a Blockly.ScratchBlocks.ProcedureUtils object that lives in blocks/procedures.js
- Decomposes procedure functions
- Adds comments to procedure mutation functions
- Make procedure_callnoreturn blocks handle mutations being called multiple times
- Adds argumentid to mutations (see https://github.com/LLK/scratch-blocks/issues/1200)
- Keeps shadow blocks and real blocks attached when applying the same mutation twice.

This doesn't yet handle the case where a mutation restructures the procedure call (e.g. moving parameters around.

### Reason for Changes

Make custom procedures better!

### Test Coverage

Tested in custom_procedure_playground.html.  Hitting "ok" applies a mutation to the blocks in the right-side workspace.

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -49,7 +49,10 @@ Blockly.ScratchBlocks.ProcedureUtils.getProcCode = function() {
 Blockly.ScratchBlocks.ProcedureUtils.callerMutationToDom = function() {
   var container = document.createElement('mutation');
   container.setAttribute('proccode', this.procCode_);
+  container.setAttribute('argumentnames', JSON.stringify(this.argumentNames_));
+  container.setAttribute('argumentdefaults', JSON.stringify(this.argumentDefaults_));
   container.setAttribute('argumentids', JSON.stringify(this.argumentIds_));
+  container.setAttribute('warp', this.warp_);
   return container;
 };
 
@@ -60,10 +63,18 @@ Blockly.ScratchBlocks.ProcedureUtils.callerMutationToDom = function() {
  */
 Blockly.ScratchBlocks.ProcedureUtils.callerDomToMutation = function(xmlElement) {
   this.procCode_ = xmlElement.getAttribute('proccode');
+  this.argumentNames_ =  JSON.parse(xmlElement.getAttribute('argumentnames'));
+  this.argumentDefaults_ =  JSON.parse(xmlElement.getAttribute('argumentdefaults'));
   this.argumentIds_ = JSON.parse(xmlElement.getAttribute('argumentids'));
+  this.warp_ = xmlElement.getAttribute('warp');
   this.updateDisplay_();
 };
-// TODO: Doc
+
+/**
+ * Remove all inputs on the block, including dummy inputs.
+ * @private
+ * @this Blockly.Block
+ */
 Blockly.ScratchBlocks.ProcedureUtils.removeAllInputs_ = function() {
   // remove all inputs, including dummy inputs.
   for (var i = 0, input; input = this.inputList[i]; i++) {
@@ -74,7 +85,16 @@ Blockly.ScratchBlocks.ProcedureUtils.removeAllInputs_ = function() {
   }
   this.inputList = [];
 };
-// TODO: Doc
+
+/**
+ * Disconnect old blocks from all value inputs on this block, but hold onto them
+ * in case they can be reattached later.
+ * @return {!Object.<string, Blockly.Block>} An object mapping parameter IDs to
+ *     the blocks that were connected to those IDs at the beginning of the
+ *     mutation.
+ * @private
+ * @this Blockly.Block
+ */
 Blockly.ScratchBlocks.ProcedureUtils.disconnectOldBlocks_ = function() {
   // Remove old stuff
   var connectionMap = {};
@@ -94,8 +114,16 @@ Blockly.ScratchBlocks.ProcedureUtils.disconnectOldBlocks_ = function() {
   }
   return connectionMap;
 };
-// TODO: Doc.
-Blockly.ScratchBlocks.ProcedureUtils.deleteOldShadows_ = function(connectionMap) {
+
+/**
+ * Delete all shadow blocks in the given map.
+ * @param {!Object.<string, Blockly.Block>} connectionMap An object mapping
+ *     parameter IDs to the blocks that were connected to those IDs at the
+ *     beginning of the mutation.
+ * @private
+ * @this Blockly.Block
+ */
+Blockly.ScratchBlocks.ProcedureUtils.deleteShadows_ = function(connectionMap) {
   // Get rid of all of the old shadow blocks if they aren't connected.
   if (connectionMap) {
     for (var id in connectionMap) {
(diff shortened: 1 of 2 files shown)


blocks_vertical/procedures.js
```diff
@@ -28,6 +28,289 @@ goog.provide('Blockly.Blocks.procedures');
...
+ *     mutation.
+ * @private
+ * @this Blockly.Block
+ */
+Blockly.ScratchBlocks.ProcedureUtils.disconnectOldBlocks_ = function() {
+  // Remove old stuff
+  var connectionMap = {};
+  for (var id in this.paramMap_) {
+    var input = this.paramMap_[id];
+    console.log(id + ' ' + input);
```
paulkaplan: Do we want to leave the logging in? (ditto for other places)

blocks_vertical/procedures.js
```diff
@@ -28,6 +28,289 @@ goog.provide('Blockly.Blocks.procedures');
...
 goog.require('Blockly.constants');
 
+// TODO: Create a namespace properly.
+Blockly.ScratchBlocks.ProcedureUtils = {};
+
+/**
+ * Returns the name of the procedure this block calls, or the empty string if
+ * it has not yet been set.
+ * @return {string} Procedure name.
+ * @this Blockly.Block
```
picklesrus, line 39: does the compiler mind this? IIUC (which is definitely questionable in this case) Blockly.Block isn't something it knows about in this context.

blocks_vertical/procedures.js
```diff
@@ -28,6 +28,289 @@ goog.provide('Blockly.Blocks.procedures');
...
+ */
+Blockly.ScratchBlocks.ProcedureUtils.getProcCode = function() {
+  return this.procCode_;
+};
+
+/**
+ * Create XML to represent the (non-editable) name and arguments.
```
picklesrus: Maybe add "for the function caller block" or some text in the comment to indicate this is for caller blocks, not the definition ones.

blocks_vertical/procedures.js
rachel-fenichel: Done.

blocks_vertical/procedures.js
```diff
@@ -28,6 +28,289 @@ goog.provide('Blockly.Blocks.procedures');
...
+  for (var id in this.paramMap_) {
+    var input = this.paramMap_[id];
+    console.log(id + ' ' + input);
+    if (input.connection) {
+      // Remove the shadow DOM.  Otherwise a shadow block will respawn
+      // instantly, and we'd have to remove it when we remove the input.
+      input.connection.setShadowDom(null);
+      var target = input.connection.targetBlock();
+      connectionMap[id] = target;
+      if (target) {
```
picklesrus, line 141: what does it mean when connectionMap[id] is null?  That there is a parameter, but there is no value provided via a connected, non-shadow, block?

blocks_vertical/procedures.js
```diff
@@ -28,6 +28,289 @@ goog.provide('Blockly.Blocks.procedures');
...
+
+/**
+ * Build a DOM node representing a shadow block of the given type.
+ * @param {string} type One of 'b' (boolean), 's' (string) or 'n' (number).
+ * @return {!Element} The DOM node representing the new shadow block.
+ * @private
+ * @this Blockly.Block
+ */
+Blockly.ScratchBlocks.ProcedureUtils.buildShadowDom_ = function(type) {
+  var xmlStart = '<xml xmlns="http://www.w3.org/1999/xhtml">';
```
picklesrus: I'm not a huge fan of writing xml as strings in javascript, especially since we turn around and call textToDom on it. Would it be better to just create the dom element directly?
ie.
goog.dom.createDom('shadow');
element.setAttribute('type', 'math_number');
 etc.

blocks_vertical/procedures.js
```diff
@@ -28,6 +28,289 @@ goog.provide('Blockly.Blocks.procedures');
...
+/**
+ * Remove all inputs on the block, including dummy inputs.
+ * @private
+ * @this Blockly.Block
+ */
+Blockly.ScratchBlocks.ProcedureUtils.removeAllInputs_ = function() {
+  // remove all inputs, including dummy inputs.
+  for (var i = 0, input; input = this.inputList[i]; i++) {
+    if (input.connection && input.connection.targetBlock()) {
+      console.warn("connection was still attached?!");
```
picklesrus: want to add some more info to the message that allows the developer to better narrow down when this happened? Perhaps the block id? If this happens in a large workspace, I could see this being tricky to reproduce.

blocks_vertical/procedures.js
```diff
@@ -29,6 +29,281 @@ goog.require('Blockly.Blocks');
...
+ * @this Blockly.Block
+ */
+Blockly.ScratchBlocks.ProcedureUtils.reattachBlock_ = function(input, inputType,
+    oldBlock, id, connectionMap) {
+  if (inputType == 'n' || inputType == 's') {
+    input.connection.setShadowDom(this.buildShadowDom_(inputType));
+  } else {
+    input.setCheck('Boolean');
+  }
+  connectionMap[id] = null;
```
picklesrus, line 267: hmm, I don't understand why the entry in the connection map gets set to null here. Which means I'm probably misunderstanding something higher level.  The connectionMap is mutable during the mutation? I thought it was supposed to be the state of the block before the mutation was applied?

picklesrus (commented):
I haven't fully read through createInputs_ yet. Which part of the blockly procedure code (if any) does this model?

blocks_vertical/procedures.js
rachel-fenichel: Not that I know of--it didn't mind it before, when we had @this on functions inside the block definition object.

Doc for context, which doesn't say either way: https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#this-type

blocks_vertical/procedures.js
rachel-fenichel: Done, and now there are two sets of functions (callerMutationToDom and definitionMutationToDom) because not all of the blocks know about the argument names and default values.

I'm open to renaming all of the blocks once I finish defining them all and we know what the full set will be.

blocks_vertical/procedures.js
rachel-fenichel: Correct.  Strings and numbers should always have something connected (a shadow or a real block) but booleans may just be empty.

blocks_vertical/procedures.js
rachel-fenichel: I may be overloading connectionMap.  After reattaching all blocks that I can, I want to delete all shadow blocks that were not reattached--and probably bump all non-shadow blocks that were not reattached.

The connection map has all of the connected blocks from right before the mutation was applied.  I'm using it to figure out if I have a block to reconnect to a given input.  Each block can only be reattached once so it's safe to remove it here.  After creating all of the inputs, any remaining blocks in the map must need to be deleted or bumped.

The alternative is to create a separate list of blocks (orphanBlocks?), add all blocks to it at the beginning, and then remove blocks from it as I go.  I'm happy to do that if it's clearer.

blocks_vertical/procedures.js
rachel-fenichel: Removed, because it was just for my debugging.

paulkaplan (approved):
VM has been updated accordingly!

PR #1212, opened 2017-11-06
Bring back the list data category

paulkaplan:
### Resolves

Brings back the "create list" option and in doing so brings back lots of the functionality for working with lists. This depends on https://github.com/LLK/scratch-blocks/pull/1211

List block creation uses the typed variables features of scratch-blocks. The vm will need to be changed quite a bit to bring it up-to-date with typed variables. So we should not expect lists to actually "work" correctly yet with the vm/gui. But they are created and used properly in scratch-blocks (that is, they are created with the list type, are kept separate from the scalar variables in dropdowns, etc.)

### DIFF
--- a/core/data_category.js
+++ b/core/data_category.js
@@ -62,8 +62,7 @@ Blockly.DataCategory = function(workspace) {
   }
 
   // Now add list variables to the flyout
-  // Disable new list button until it is supported in scratch-vm
-  // Blockly.DataCategory.addCreateButton(xmlList, workspace, 'LIST');
+  Blockly.DataCategory.addCreateButton(xmlList, workspace, 'LIST');
   variableModelList = workspace.getVariablesOfType('list');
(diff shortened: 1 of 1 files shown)


core/data_category.js
```diff
@@ -63,7 +63,7 @@ Blockly.DataCategory = function(workspace) {
 
   // Now add list variables to the flyout
   // Disable new list button until it is supported in scratch-vm
```
rachel-fenichel: Get rid of this comment (line 65), then good to go.

rachel-fenichel (approved):

Kenny2github:
Only commit came after reviews...?

paulkaplan:
@Kenny2github ah that is a quirk of the way I committed. First I committed with the comment rachel wanted to be removed, then amended the commit and force pushed. Probably a bad habit of mine, since it screws with history (makes it look like I committed after the review). I guess I did it so that there was no record of my mistake ;)

Issue #1213, opened 2017-11-06
Names for custom procedure block types


rachel-fenichel:
Proposed new names:
![image](https://user-images.githubusercontent.com/13686399/32468256-5a3d7756-c302-11e7-8543-18db82dd58c5.png)

Old names -> new names:
- `procedures_callnoreturn_internal` -> `procedures_defnoreturn`
- `procedures_defnoreturn` -> `procedures_defwrapper`
- `procedures_callnoreturn` -> `procedures_callnoreturn` (no change)

@paulkaplan we'd have to make this change simultaneously here and in scratch-vm.

rachel-fenichel:
Renames that we settled on:
- `procedures_callnoreturn_internal` -> `procedures_prototype`
- `procedures_defnoreturn` -> `procedures_definition`
- `procedures_callnoreturn` -> `procedures_call`

Paul already made the change in the VM.

PR #1214, opened 2017-11-06
Fix applying multiple mutations to a single procedure_callnoreturn_internal block

rachel-fenichel:
### Resolves

Part of https://github.com/LLK/scratch-blocks/issues/1185

### Proposed Changes

This PR is similar to #1203, but for the internal block on the procedure definition.

I'm using many of the same helper functions to apply the mutation and update the display.

I shuffled some code around to minimize the number of functions that had to be different between the procedure definition and call.  The main difference is that the definition has to know about the argument names, and the call doesn't have that information.

The naming isn't that clear, but I opened a separate issue for changing that since we'll need to do it on scratch-vm at the same time: #1213.

### Reason for Changes

To make custom procedures work.

### Test Coverage

Tested in the custom procedure playground.  I updated 'ok' to apply the mutation to the callnoreturn_internal block as well as the callnoreturn block.

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -367,7 +367,7 @@ Blockly.ScratchBlocks.ProcedureUtils.createInputCallerInternal_ = function(type,
     this.reattachBlock_(input, type, oldBlock, id, connectionMap);
   } else {
     var argumentText = this.argumentNames_[index];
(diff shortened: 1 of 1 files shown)


picklesrus (commented):
LGTM

Issue #1218, opened 2017-11-09
Input shapes aren't removed when an input is removed


rachel-fenichel:
### Expected Behavior

Deleting an input on a block should clean up everything associated with it.

### Actual Behavior

The input, connected block, and shadow block are deleted.  The input shape is not deleted.

In this gif there's only one input, so I know it'll be deleted when I click "remove random input"
![3c310616-0b8c-40f3-a569-a40a99883966](https://user-images.githubusercontent.com/13686399/32582850-af7ca080-c4a5-11e7-9c52-8930d1db3d00.gif)

### Steps to Reproduce

Create a block with input shapes
From the console call `block.removeInput(<inputName>)`

This is exposed by mutators in custom procedures.

### Operating System and Browser

Any

### One solution
  
I can fix this by adding the following code to `Blockly.Input.prototype.dispose`, but I don't know why the input shapes are associated with the block instead of the input.

```
  // If there's an input shape for this, get rid of it.
  var inputShape = this.sourceBlock_.inputShapes_[this.name];
  if (inputShape) {
    goog.dom.removeNode(inputShape);
    delete this.sourceBlock_.inputShapes_[this.name];
  }
```
![de9d4f76-eca3-4767-86da-06f1b48d272a](https://user-images.githubusercontent.com/13686399/32582925-081806e4-c4a6-11e7-9ff4-5f87dfe12dd8.gif)

tmickel:
> I don't know why the input shapes are associated with the block instead of the input

Hmmm, might have been me...perhaps an accident for the lack of knowing better ;p blockly renders its inputs (holes/notches in the blocks) in the block rendering right, so it seemed natural I guess...

rachel-fenichel:
Makes sense, and that was my guess.  It shouldn't be too hard to put in the right place.

Issue #1217, opened 2017-11-08
Malformed SVG: missing `d` attribute on a `path`


rachel-fenichel:
Bug found on pxt-blockly, which uses scratch-blocks rendering: https://github.com/Microsoft/pxt-blockly/pull/86 and https://github.com/Microsoft/pxt-blockly/issues/84

### Expected Behavior

Blocks load in IE.

### Actual Behavior

Some blocks can't load in IE 11.  In Chrome there is a `path` with no `d` attribute, but Chrome recovers cleanly.
![image](https://user-images.githubusercontent.com/13686399/32580973-d540a6c6-c49c-11e7-9e48-6199b9a2543c.png)

The problem occurs when loading from XML.  A block with a connected input with no shadow block will fail; a block with a connected input with a shadow block won't fail.  The problem has to do with the rendering of the `inputShape` on the input with no shadow block.

### Steps to Reproduce

In the vertical playground, import this code:
```
<xml xmlns="http://www.w3.org/1999/xhtml">
  <variables></variables>
  <block type="operator_and" id="{o/zPw8mbvA4/K#l:DY#" x="219" y="340">
    <value name="OPERAND2">
      <block type="operator_not" id=";;WvgQM]fo%[WIC=JHj$"></block>
    </value>
  </block>
  <block type="operator_multiply" id="qIedr}Oh7)bU:M)_1i?~" x="239" y="431">
    <value name="NUM1">
      <shadow type="math_number" id="5@*vu7.;lmFcBi7Skbnr">
        <field name="NUM"></field>
      </shadow>
    </value>
    <value name="NUM2">
      <shadow type="math_number" id=")TRj+SV/SRUKtZf,DWIO">
        <field name="NUM"></field>
      </shadow>
      <block type="operator_round" id="3k$HdNFyy9(QAA_ZfW*L">
        <value name="NUM">
          <shadow type="math_number" id="l;VaN.ZqP_bJzyh,?T4v">
            <field name="NUM"></field>
          </shadow>
        </value>
      </block>
    </value>
  </block>
</xml>
```

It should look like this: 
![image](https://user-images.githubusercontent.com/13686399/32580922-8905cc0a-c49c-11e7-9139-67604f3ae563.png)

Without dragging any blocks, open up developer tools->elements.

Inspect the paths on the blocks above.

### Operating System and Browser

Visible problem on IE 11 when loading from XML, but the problem is present on Chrome and other browsers--it just recovers nicely.

[...]

PR #1219, opened 2017-11-09
Make inputs own input shapes, and dispose of them correctly

rachel-fenichel:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1218 and https://github.com/LLK/scratch-blocks/issues/1217

### Proposed Changes

Move "input shapes" into the input class, and rename to "outlinePath".  Check for the existence of the path in `Input.dispose` and remove it from the DOM if necessary.  Update callers.

Always have a path on the input shape, even if it's empty.

### Reason for Changes

Input shapes were owned by the block and not being properly disposed of before.  They also caused problems in IE because the SVG was sometimes malformed.

### Test Coverage

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -531,8 +531,10 @@ Blockly.BlockSvg.prototype.updateColour = function() {
   this.svgPath_.setAttribute('fill-opacity', this.getOpacity());
 
   // Update colours of input shapes.
-  for (var shape in this.inputShapes_) {
-    this.inputShapes_[shape].setAttribute('fill', this.getColourTertiary());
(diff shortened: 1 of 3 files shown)


rachel-fenichel:
@microsoftsam FYI, I found another problem with input shapes yesterday (#1218) that you might run into with mutators on your if/else blocks.

paulkaplan (approved):
Looks good to me

picklesrus (commented):
LGTM

RoboErikG:
Btw, would it be worth adding a test to Travis that renders the problem block and verifies there aren't errors?

rachel-fenichel:
I think we have Travis set up to only use headless browsers.

PR #1224, opened 2017-11-10
Use data_listcontents instead of data_variable for list getter block

### LINKED ISSUE #1222: Data category uses the wrong block for list getter
### Expected Behavior

The block meant to be the list getter is `data_listcontents`, but the dynamic category builder for data uses the block `data_variable` instead. 

https://github.com/LLK/scratch-blocks/blob/develop/core/data_category.js#L69

The `data_listcontents` block is defined here: https://github.com/LLK/scratch-blocks/blob/develop/blocks_vertical/data.js#L152-L173

### Actual Behavior

The correct block is used. This is important because the field name is different (VARIABLE vs. LIST) which makes the VM not able to run this block. 

### Steps to Reproduce

In the playground, create a variable and list, and add the getter blocks for both into the workspace, then export XML and see how the blocks are both named `data_variable`. 

I assume this was a minor oversight in the original list block implementation, because both blocks use the `field_variable_getter` field.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1222

### Proposed Changes

Have the data category use the correct block in the flyout. 

screen shot 2017-11-10 at 10 15 21 am

### DIFF
--- a/core/data_category.js
+++ b/core/data_category.js
@@ -66,7 +66,7 @@ Blockly.DataCategory = function(workspace) {
   variableModelList = workspace.getVariablesOfType('list');
   variableModelList.sort(Blockly.VariableModel.compareByName);
   for (var i = 0; i < variableModelList.length; i++) {
-    Blockly.DataCategory.addDataVariable(xmlList, variableModelList[i]);
+    Blockly.DataCategory.addDataList(xmlList, variableModelList[i]);
   }
 
   if (variableModelList.length > 0) {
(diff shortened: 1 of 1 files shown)


core/data_category.js
```diff
@@ -172,6 +172,20 @@ Blockly.DataCategory.addHideVariable = function(xmlList, variable) {
 };
 
 /**
+ * Construct and add a data_listcontents block to xmlList.
+ * @param {!Array.<!Element>} xmlList Array of XML block elements.
+ * @param {?Blockly.VariableModel} variable Variable to select in the field.
+ */
+Blockly.DataCategory.addDataList = function(xmlList, variable) {
```
rachel-fenichel, line 179: Consider renaming some of the references to "variable" to make it clear it's a list.

rachel-fenichel (approved):
One naming comment, but I don't feel strongly about it.  LGTM

Issue #1195, opened 2017-10-30
Applying edits to an existing procedure


rachel-fenichel:
_"Editor" refers to the custom procedure editor in this bug._
---------------
### Questions
- Should labels be coalesced?  (See example 1).
- Should a label appear when the editor is opened?  (See example 2).
- When an input is deleted, should a connected block be deleted or bumped away? (See example 3).
- When an input is deleted, should all other connected blocks stick to their respective connections?  (See examples 4-6; I think the answer is yes.)

-----------------
### Example 1
I can add a new text label directly after an existing text label and save the edit.  When I reopen the procedure editor, the two labels have been coalesced into a single label:
![0b41bd3e-c51b-4e58-817c-956d12475b39](https://user-images.githubusercontent.com/13686399/32198666-3a0015ec-bd86-11e7-825c-df429575733b.gif)

-------------------

### Example 2

If my block ends with an input and I open the editor, it always has a new text label at the end.  Deleting the text label is a no-op, and it'll be back when I reopen the editor.  You'll see this repeatedly in the following examples.

If the block does not end with an input, no text label will be added.

![fe65a47a-9606-41f7-937f-091feedc93fc](https://user-images.githubusercontent.com/13686399/32198757-b3facaae-bd86-11e7-81dd-fad90c59f9ed.gif)

----------------------
### Example 3

This block has a text label and a single input named `animal`, to which I've connected an `aardvark`.
I open the custom procedure editor and delete `animal`. 
When the custom procedure call updates, the `aardvark` is deleted.
_In Blockly the aardvark would have been disconnected and bumped away, but would have remained on the workspace._

![da40185f-d31f-49c4-84be-ef8b32fc7695](https://user-images.githubusercontent.com/13686399/32198914-84e23ee0-bd87-11e7-9ef3-8f9bc6c21858.gif)

-------------

### Example 4

Here I have a custom procedure block with a text label and two string inputs.  The first is named `mammal` and the second is named `cephalopod`. 

I have a custom command block below the definition.  It has two blocks connected.  `aardvark` is connected to `mammal` and `cuttlefish` is connected to `cephalopod`.

I open the custom procedure editor and delete `mammal`, which is the first string input.

Expected behaviour: `mammal` input is deleted, so `aardvark` is deleted as well.
Actual behaviour: `mammal` input is deleted.  `aardvark` is moved into the `cephalopod` connection.  `cuttlefish` is deleted.

[...]

Issue #607, opened 2016-09-18
Implement custom procedures


MegaApuTurkUltra:
Lots of people's 2.0 custom blocks won't make any sense when converted into the new format.

For example this:

![](http://i.imgur.com/MLZY32w.png)

Would presumably become something like this in the conversion from 2.0 to 3 format

![](http://i.imgur.com/NiRO0vh.png)

Scratchers might get confused by things like this.

tmickel:
The current implementation is a direct copy-paste from Blockly - the actual version is only in the earliest design stages, and is likely to look significantly different. @carljbowman

tmickel:
From @TheBrokenRail:

> Here's Some Issues I Found With The UI in procedure blocks/custom blocks
> - [ ] Need Space Between Blocks In Block Palate
> - [ ] Input Block Is Offset From Curser
> - [ ] No Background For Warning About Standalone report block Needing To Be Used In a custom block definition
> - [ ] Needs Proper UI for input Selection
> - [ ] Gear And Warning Triangles should be maybe should be the same color as the blocks
> - [ ] Use Define Instead of to
> - [x] Bottom of define with report is longer than top

rachel-fenichel:
One more:
- Need to be able to add label text between inputs

tmickel:
Yep, and inputs need types (number, string, boolean).

towerofnix:
Don't forget default values for number/string inputs! That's implemented in the actual "blocks" part of Scratch 2.0 (though there's no UI for it atm).

CosmicWebServices:
Moved from #631:

![image](https://cloud.githubusercontent.com/assets/10731781/18887337/9df51f9c-84c1-11e6-9eb6-d884cac3ffbe.png)

The define part is hard to read because of the color of the text.

towerofnix:
Also the red outline is _really_ red. It's kind of hard to see. Later on the shape will probably be changed to look like a stack (or reporter) outline but for now the color should probably be fixed anyways.

Reasons for the same color as the outline of stack blocks:
- It _is_ the outline of a block.
- It represents a block more clearly.

Reasons for a lighter or different color:
- It's actually the outline of a preview of what the block will look like, generally, so it distinguishes between a real block and a preview.
- It looks similar to how custom block definitions currently look.

Also probably more but just some observations. :)

TheBrokenRail:
When selecting the name field it gets all big and weird and when closing even bigger fort a moment, can you change it to be like other text fields?

[...]

PR #1225, opened 2017-11-10
Custom procedure editing: renaming, adding, and deleting inputs and labels

rachel-fenichel:
### Resolves
Related to #607, https://github.com/LLK/scratch-blocks/issues/1195, https://github.com/LLK/scratch-blocks/issues/1185

### Proposed Changes

Each `argument` on a custom procedure now has an `id` and a `displayName` instead of an `id` and a `name`.
`argument.id` now corresponds to the `name` field on the `Blockly.Input` that stores and renders that argument.
`argument.displayName` is what is seen and edited by the user.

The `paramMap` is now gone, since the mapping is done by `argument.id`.

`procedures_mutator_root` is now called `procedures_declaration_root`.

### Reason for Changes

Continuing to implement custom procedures.

### Test Coverage

Tested in the custom procedure playground.  There are two useful buttons for testing: "add random input" and "remove random input".

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -107,6 +107,7 @@ Blockly.ScratchBlocks.ProcedureUtils.definitionDomToMutation = function(xmlEleme
 
 /**
  * Remove all inputs on the block, including dummy inputs.
+ * Assumes no input has shadow DOM set.
  * @private
  * @this Blockly.Block
  */
@@ -196,7 +197,16 @@ Blockly.ScratchBlocks.ProcedureUtils.createAllInputs_ = function(connectionMap) 
       }
       newLabel = component.substring(2).trim();
 
-      this.createInput_(inputType, inputCount, connectionMap);
+      var id = this.argumentIds_[inputCount];
+      var oldBlock = null;
+      if (connectionMap && (id in connectionMap)) {
+        oldBlock = connectionMap[id];
+      }
+      var input = this.appendValueInput(id);
+      if (inputType == 'b') {
+        input.setCheck('Boolean');
+      }
+      this.populateInput_(inputType, inputCount, connectionMap, id, oldBlock, input);
       inputCount++;
     } else {
       newLabel = component.trim();
@@ -263,8 +273,6 @@ Blockly.ScratchBlocks.ProcedureUtils.attachShadow_ = function(input, inputType) 
       newBlock.render(false);
     }
     newBlock.outputConnection.connect(input.connection);
-  } else {
-    input.setCheck('Boolean');
   }
 };
 
@@ -308,22 +316,8 @@ Blockly.ScratchBlocks.ProcedureUtils.attachArgumentReporter_ = function(
  * @private
  * @this Blockly.Block
  */
-Blockly.ScratchBlocks.ProcedureUtils.createInput_ = function(type, index,
-    connectionMap) {
-  // Shared
-  var id = this.argumentIds_[index];
-  var oldBlock = null;
-  if (connectionMap && (id in connectionMap)) {
-    oldBlock = connectionMap[id];
-  }
-  var input = this.appendValueInput(id);
-  if (type == 'b') {
-    input.setCheck('Boolean');
-  }
-  // End shared
-
-
-
+Blockly.ScratchBlocks.ProcedureUtils.populateInputCaller_ = function(type, index,
+    connectionMap, id, oldBlock, input) {
   if (connectionMap && oldBlock) {
     // Reattach the old block.
     connectionMap[id] = null;
@@ -351,21 +345,8 @@ Blockly.ScratchBlocks.ProcedureUtils.createInput_ = function(type, index,
  * @private
  * @this Blockly.Block
  */
-Blockly.ScratchBlocks.ProcedureUtils.createInputCallerInternal_ = function(type,
-    index, connectionMap) {
-  // Shared
-  var id = this.argumentIds_[index];
-  var oldBlock = null;
-  if (connectionMap && (id in connectionMap)) {
-    oldBlock = connectionMap[id];
-  }
-  var input = this.appendValueInput(id);
-  if (type == 'b') {
-    input.setCheck('Boolean');
-  }
-  // End shared
-
-
+Blockly.ScratchBlocks.ProcedureUtils.populateInputCallerInternal_ = function(type,
+    index, connectionMap, id, oldBlock, input) {
   var oldTypeMatches =
     Blockly.ScratchBlocks.ProcedureUtils.checkOldTypeMatches_(oldBlock, type);
   var argumentText = this.argumentNames_[index];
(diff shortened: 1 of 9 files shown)


blocks_vertical/procedures.js
```diff
@@ -90,32 +82,55 @@ Blockly.ScratchBlocks.ProcedureUtils.definitionMutationToDom = function() {
...
  * @this Blockly.Block
  */
-Blockly.ScratchBlocks.ProcedureUtils.removeAllInputs_ = function() {
-  // Delete inputs directly instead of with block.removeInput to avoid splicing
-  // out of the input list at every index.
-  for (var i = 0, input; input = this.inputList[i]; i++) {
-    input.dispose();
+Blockly.ScratchBlocks.ProcedureUtils.updateDisplay_ = function() {
+  var wasRendered = this.rendered;
+  this.rendered = false;
```
picklesrus, line 120: do you need this? Looks like it is always set below. But maybe some of these private helpers use it?

blocks_vertical/procedures.js
```diff
@@ -196,13 +202,67 @@ Blockly.ScratchBlocks.ProcedureUtils.createAllInputs_ = function(connectionMap)
       }
       newLabel = component.substring(2).trim();
 
-      var inputName = inputPrefix + inputCount;
-      this.createInput_(inputType, inputName, inputCount, connectionMap);
+      var id = this.argumentIds_[inputCount];
+      var oldBlock = null;
+      if (connectionMap && (id in connectionMap)) {
```
picklesrus: is there a difference between whether the id is in the connection map and 
connectionMap[id] returning null?  If not, would it be better to just get it and check for null rather than having to scan the whole map?

blocks_vertical/procedures.js
```diff
@@ -312,91 +345,208 @@ Blockly.ScratchBlocks.ProcedureUtils.attachArgumentReporter_ = function(
...
-  var id = this.argumentIds_[index];
-  var oldBlock = null;
-  if (connectionMap && (id in connectionMap)) {
-    oldBlock = connectionMap[id];
+Blockly.ScratchBlocks.ProcedureUtils.populateInputCallerInternal_ = function(type,
+    index, connectionMap, oldBlock, input) {
+  var oldTypeMatches =
+    Blockly.ScratchBlocks.ProcedureUtils.checkOldTypeMatches_(oldBlock, type);
+  var displayName = this.displayNames_[index];
+  if (connectionMap && oldBlock && oldTypeMatches) {
```
picklesrus: Not important but when oldBlock is null, oldTypeMatches returns false so you could get rid of it in the condition

blocks_vertical/procedures.js
```diff
@@ -312,91 +345,208 @@ Blockly.ScratchBlocks.ProcedureUtils.attachArgumentReporter_ = function(
...
+ * @private
+ * @this Blockly.Block
+ */
+Blockly.ScratchBlocks.ProcedureUtils.populateInputDeclarationRoot_ = function(type,
+    index, connectionMap, oldBlock, input) {
+  var oldTypeMatches =
+    Blockly.ScratchBlocks.ProcedureUtils.checkOldTypeMatches_(oldBlock, type);
+
+  var displayName = this.displayNames_[index];
+  if (connectionMap && oldBlock && oldTypeMatches) {
```
picklesrus: Is this code essentially the same as what's in populateInputCallerInternal_ except that setFieldValue gets called with different type?

blocks_vertical/procedures.js
```diff
@@ -312,91 +345,208 @@ Blockly.ScratchBlocks.ProcedureUtils.attachArgumentReporter_ = function(
...
+      throw new Error(
+          'Unexpected input type on a procedure mutator root: ' + input.type);
+    }
   }
+};
 
-  this.createAllInputs_(connectionMap);
-  this.deleteShadows_(connectionMap);
+/**
+ * Externally-visible function to add a label field to the mutator root block.
```
picklesrus: should this read declaration root now?

blocks_vertical/procedures.js
```diff
@@ -232,31 +292,6 @@ Blockly.ScratchBlocks.ProcedureUtils.buildShadowDom_ = function(type) {
 };
 
 /**
- * Reattach a block to the correct input after a mutation, and give the
```
picklesrus, line 235: this isn't new but just noticed it.  In the buildShadowDom function, does the 'hello world' text ever get shown to the user?

blocks_vertical/procedures.js
```diff
@@ -312,91 +345,208 @@ Blockly.ScratchBlocks.ProcedureUtils.attachArgumentReporter_ = function(
...
-    this.render();
-  }
+/**
+ * Externally-visible function to add a string/number field to the mutator root
+ * block.
+ * @public
+ */
+Blockly.ScratchBlocks.ProcedureUtils.addStringNumberExternal = function() {
+  this.procCode_ = this.procCode_ + ' %s';
+  this.displayNames_.push('string or number');
```
picklesrus, line 562: does this need to be i18n'd?

picklesrus (commented):
Generally okay with this, but let's chat about naming and documenting this. Same with tests.
The other thing that it feels like there are too many magic strings floating around in here.  That might be a function of the existing code (e.g. input type) but I wonder if there is a way around typing things like 'b',  'string or number', and block types all over.

blocks_vertical/procedures.js
rachel-fenichel: I'm using it to disable rendering while I create new child blocks, then do a single rendering pass at the end if necessary.  Adding new fields also means that I need to call initSvg, but only if it was previously rendered.

blocks_vertical/procedures.js
rachel-fenichel: Yes, the helloWorld text is shown to the user if they:
- had a procedure call with another block attached to the input.
- updated the procedure declaration, causing the mutation to be reapplied to the call
- removed the attached block, revealing the shadow block behind it

blocks_vertical/procedures.js
rachel-fenichel: If we keep a string here then yes, it needs i18n.  I don't know whether the plan is to have text here by default.

rachel-fenichel:
@paulkaplan @picklesrus ready for review again.  I'm going to look into testing.

core/procedures.js
```diff
@@ -342,7 +342,7 @@ Blockly.Procedures.getDefinition = function(name, workspace) {
  * @private
  */
 Blockly.Procedures.editProcedureCallback_ = function(block) {
-  if (block.type == 'procedures_defnoreturn') {
+  if (block.type == Blockly.BlockSvg.DEFINE_BLOCK_TYPE) {
```
paulkaplan: Seems slightly strange that the `PROCEDURES_CALL_BLOCK_TYPE` is defined on blockly, and `DEFINE_BLOCK_TYPE` is defined on BlockSvg?

paulkaplan:
Tests still seem to be failing, and had one tiny question about constant definitions.

Wanted to just make a checklist here of things that have to change in the VM (opcode mapping) to support this:
- [x] `procedures_defnoreturn` => `procedures_definition`
- [x] `procedures_callnoreturn` => `procedures_call`
- [x] `procedures_callnoreturn_internal` => `procedures_prototype`
- [x] `procedures_param` => `argument_reporter_string_number` or `argument_reporter_boolean`

Is that last one right?

rachel-fenichel:
Horizontal tests are failing, yeah.  I think it's a problem with mocks not being set up properly, but I'll take a look.

Opcode mapping: yes, those all look right.  Procedures_param is now two separate things.

core/procedures.js
rachel-fenichel: Fixed, which also fixed the tests.

PR #1228, opened 2017-11-14
Make events specify when a block is sprite-specific

fsih:
Towards https://github.com/LLK/scratch-vm/issues/564


rachel-fenichel:
Why is this in scratch-blocks and not the VM?

The blocks don't know which sprite they are attached to (that's the VM's job).  

The new "sprite-specific" tag is unused in scratch-blocks in this PR.  It's per block type, so you get the same information by checking the block type against a whitelist.

fsih:
Ok that's fair, I will find a better place for this

PR #1236, opened 2017-11-16
Return early from field init if the field has already been initialized.

rachel-fenichel:
### Resolves

None
### Proposed Changes

Most fields check whether they're already been initialized on a block before creating SVG elements.  These fields didn't, so if init was called twice there would be two instances of each SVG element.  One would be a memory leak, since we would lose the pointer to it.

### Reason for Changes

The previous behaviour was leaky and created duplicates of some elements.

### Test Coverage

Checked that I can open the playground and scroll through all blocks in the toolbox.


paulkaplan (approved):
LGTM

Issue #1231, opened 2017-11-16
Implement label editor


rachel-fenichel:
### Images 
From https://github.com/LLK/scratch-blocks/issues/607

Label editor with the text "block", not selected:
![](https://cloud.githubusercontent.com/assets/3409578/23229014/ad9d0dfa-f90c-11e6-8a20-f09ece78ec4f.png)

Label editor with the text "dance", not selected:
![](https://cloud.githubusercontent.com/assets/3409578/23230704/003b9c24-f913-11e6-8884-5f81e30e6ce5.png)

Label editor with the text "dance", selected:
![](https://cloud.githubusercontent.com/assets/3409578/23230624/b633d13c-f912-11e6-8812-8270545bf8c6.png)

Empty label editor, selected:
![](https://cloud.githubusercontent.com/assets/3409578/23229321/e4567952-f90d-11e6-81de-c1cd682a01a0.png)

### Rendering questions
- What is the minimum width of an empty label editor?
- What color is the label editor background when not selected?
- What color is the label editor background when selected?
- How large is the vertical padding around the text?
- How large is the horizontal padding around the text?

### Behaviour questions
- When are empty label editors combined?
  - Does emptying a label editor and hitting enter delete the label editor?
- Are there any constraints on what text is allowed in the label editor?
- What is the default text when a label editor is added to a block?
- Does adding a label editor automatically select the label editor for editing?
- Does a label editor get a delete icon above it when it is selected for editing?

### Notes
- The label editor is a text input on a block.
- The label editor's background is a different color from the rest of the block.
- The label editor's text is white (like most (all?) labels) instead of black (like all text inputs).

rachel-fenichel:
@carljbowman FYI these are the questions I have so far.  I'm not in a hurry to get specific colors or padding sizes, since I can just implement those with approximate values and we can tweak them later.

rachel-fenichel:
Delete icons and empty label editors:

Scratch 2.0 puts a delete icon above the selected label editor if it is not the only component of the procedure.

That is, there is a delete icon here:
![image](https://user-images.githubusercontent.com/13686399/32868291-586c0ed0-ca26-11e7-9995-987fdf78d81c.png)

But no delete icon here:
![image](https://user-images.githubusercontent.com/13686399/32868276-4154e500-ca26-11e7-9b81-9b7f3c03fff0.png)

In both cases I have selected the "bar" label for editing.

The same is true for arguments: you can't delete the only remaining one on the block.  

[...]

PR #1239, opened 2017-11-16
Tweak text input rendering when not on a shadow block

rachel-fenichel:
### Resolves

Part of https://github.com/LLK/scratch-blocks/issues/1231

### Proposed Changes

Get rid of rounded corners and white background when the text input field is not on a shadow block.

Doesn't do anything about the placement of the editor.

### Reason for Changes

Label editor rendering for custom procedures.

### Test Coverage

New: 
image

Old: 
image

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -96,14 +96,12 @@ Blockly.FieldTextInput.prototype.init = function() {
   // If not in a shadow block, draw a box.
   if (!this.sourceBlock_.isShadow()) {
     this.box_ = Blockly.utils.createSvgElement('rect', {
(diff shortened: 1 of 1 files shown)


paulkaplan:
So this changes all text input fields on non-shadow blocks right? Does just not happen anywhere else in scratch-blocks right now?

paulkaplan (approved):
Code LG, one small question about the impact of this.

rachel-fenichel:
It doesn't look like that code path is ever getting hit.  I put a breakpoint inside there and then opened up the playground, and I couldn't get it to ever hit that code.

There's actually a bunch of code in fields that isn't currently used--specifically for dropdowns with editable text.  I think some may start to be used in list blocks.

PR #1246, opened 2017-11-17
Fix expanding field_textinput width for label editors

paulkaplan:
### Resolves

Towards https://github.com/LLK/scratch-blocks/issues/1231

Fix the mega expansion of `field_textinput` when not in a shadow block (i.e. for label editors). 

original
vertical-inputs--original

Fixed
vertical-inputs--fixed

Also confirmed that the "expand past truncation" fields also still work, used in the horizontal playground (note there is a bug about which value is being displayed there, but that is not relevant)

develop
horizontal-inputs--original

this (should be identical) 
horizontal-inputs--changed


rachel-fenichel (approved):
LGTM!

Issue #1245, opened 2017-11-17
Allow for renaming a message or list to an existing variable name.


kchadha:
### Expected Behavior

Messages can share names with variables and lists.
Lists and variables cannot share names.

### Actual Behavior

Currently messages cannot share names with lists and variables.

### Steps to Reproduce

One example of reproducing this error is to create a variable named "x", and then to create a message named "x". A modal pops up that says: 'A variable named "x" already exists.'

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

### Additional comments

Currently variables often get looked up by their names, and since messages are now internally modeled as variables, this will be a big change. We would need to either change variables to get looked up by their unique IDs, or when looking up variables, also ensure that type information matches.

mrjacobbloom:
Note that in 2.0 variables and lists may not share names, though messages may have the same name as a list or variable

rachel-fenichel:
Is the plan to have them all have separate namespaces now?

kchadha:
Hmmm I may have been too eager to file an issue on Friday. For now, we will keep the functionality the same as in 2.0, which means that messages can share names with lists and variables, but lists and variables cannot share names with each other. I am editing the issue to reflect this.

rachel-fenichel:
I'd prefer separate namespaces for each type.  It wouldn't cause problems in loading old projects, and it would make implementation easier.

If you want to preserve the scratch 2.0 behaviour we would need to have a namespace tag on variable declarations and fields, as well as type tags.

thisandagain:
I'm ok with making this behavior more consistent (variables, lists, and messages all have separate namespaces) as long as this doesn't break loading / editing 2.0 projects.

rachel-fenichel:
Awesome.  It shouldn't break loading/editing 2.0 because it's becoming strictly more permissive.

rachel-fenichel:
Er, unless this breaks opening a project in 2.0 after it's been edited in 3.0.  Is that a concern?

thisandagain:
Nope. That isn't a concern. We'll support 2.0 -> 3.0, but not the other way around.

kchadha:
For the sake of complete clarity, are we saying that we want to be able to allow:

1. variables to share names with lists and messages but not themselves,
2. lists to share names with messages and variables but not themselves, and
3. messages to share names with lists and variables but not themselves?

There's overlap in those 3 statements, but I just want to make sure we're all on the same page.

[...]

Issue #1244, opened 2017-11-17
Modify messages for create variable modal (for variables, lists, and broadcast messages)


kchadha:
### Expected Behavior

The message in the modal that appears when creating a variable, list, or broadcast message should change depending on what is being created.

### Actual Behavior

Currently, the message for creating variables, messages, and lists all say: "New variable name:".

### Steps to Reproduce
Create a new messsage via the field dropdown in any of the broadcast message blocks.
Create a new list via the "Create list..." button.

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

kchadha:
Renaming requires further discussion as to whether this is a feature we want to keep. Scratch 2.0 only allows renaming variables, not lists or messages. Editing issue to reflect the scaled down scope.

Issue #749, opened 2016-12-18
Implement drop-down menus


TheBrokenRail:
- [x] All motion (and sensing of object (except for preset stage and sprite1)) menus that have sprites in them don't show sprites
(Was Using This iv VM with Sprites)
![first](https://cloud.githubusercontent.com/assets/17478432/21412972/265737e2-c7c2-11e6-8e7b-883d07330fc5.PNG)
- [ ] Broadcast Menu new broadcast doesn't do anything
![two](https://cloud.githubusercontent.com/assets/17478432/21412982/2da121f2-c7c2-11e6-8d5b-fec9ba583a65.PNG)
- [ ] Broadcast menu doesn't show broadcasts except the preset ones
![three](https://cloud.githubusercontent.com/assets/17478432/21412988/334c2dd6-c7c2-11e6-8537-b5316f645403.PNG)
- [ ] Sensing of block property menu does not show variables
![thing](https://cloud.githubusercontent.com/assets/17478432/21413067/c549ffec-c7c2-11e6-8214-14753afb5063.PNG)

TheBrokenRail:
Added screenshots to demonstrate problem.

TheBrokenRail:
I think this should be ready by the prototype cause menus need to work.

tmickel:
Thanks for the screenshots - that should be helpful when this is implemented!

Regarding your last point:
The "squared" menus with block-colored backgrounds are intentional; no reporters can be dropped into those menus.

TheBrokenRail:
What about circular menus with block colored backgrounds?

tmickel:
Those shouldn't exist- either it's a mistake, or the color palette is too subtle to see the difference. Which blocks appear that way?

TheBrokenRail:
some reporters, sensing of menus, variable getter

TheBrokenRail:
Why is variable getter shadow round and list getter shadow square?

TheBrokenRail:
I have a implementation for this that uses Blockly.Blocks.Sprite to hold sprite names and is set by vm on targets update, see closed PRs above.

TheBrokenRail:
The sensing variable menu is easily done by doubling the workspaces variable list and concaving it to the menu.

TheBrokenRail:
Should I make a PR to fix the sensing block?

TheBrokenRail:
Note, At the current way menus are implemented the menus won't be updated until they are re-rendered.

[...]

PR #1247, opened 2017-11-17
Broadcast message functionality

kchadha:
### Resolves

Resolves parts of #749 ("Broadcast Menu new broadcast doesn't do anything" and "Broadcast menu doesn't show broadcasts except the preset ones")

### Proposed Changes

This pull request adds the following pieces of functionality:
- All of the broadcast message related blocks are now populated with two options, "message1" which is a preset message the user can broadcast, and "New message..." which, like "Rename Variable"/"Delete Variable" for lists and variables will bring up a modal prompting the user to create a new message.
- Messages are treated as typed variables with the string "broadcast_msg" differentiating this type of variable from lists and scalar variables
- The default option of "message1" gets added to the variable model with the correct type (specified above) when the block gets added to the workspace.
- When the new message option is selected and the user is able to successfully create a new message, that message gets added to the variable model and the field on the block that was used to create the new message gets updated with the new message name.
- When a new message is created, all broadcast message drop-downs get updated with the new message name as a menu option.

### Reason for Changes

Broadcast messages were previously non-functional.

### Test Coverage

The changes have been tested with the existing unit tests.
@rachel-fenichel If you think there are any specific tests that need to be changed or added, please let me know.

### Additional Comments

There are a few outstanding tasks/issues remaining. These are outlined below:

- Modify the messages in the modals that pop up when creating or renaming lists and broadcast messages (issue #1244)
- Allow messages, lists, and variables to share common names (issue #1245). Resolving this issue might also require changing how variables are looked up, because I believe there are certain places where variables are looked up by name (type information would not get properly checked here).
- A next step (not covered in this PR) is to allow the drop-down field for the 'broadcast' and 'broadcast and wait' blocks to be input fields (allowing dropping other blocks into that slot).
- A question that came up is: we are currently are specifying broadcast messages as variables by specifying that their field dropdown is specifically a FieldVariable. We also specify that the drop downs should only contain variables of the types specified in the variableTypes list. E.g. broadcast message block drop downs should only contain names of broadcast messages (and not scalar variables or lists). Often, we are using this variableTypes list to discern what type the actual variable we're trying to create (by means of the 'New message...' option in the dropdown) is. See proposed changes to FieldVariable.initModel and FieldVariable.dropdownCreate, for examples. What should happen if variableTypes has more than one type in it? Perhaps we need a different way of specifying in the json representation of the block what the type of the variable is (similar to how we define a default message name for the broadcast message dropdowns, via 'variable: "message1"'

### DIFF
--- a/blocks_vertical/event.js
+++ b/blocks_vertical/event.js
@@ -80,7 +80,7 @@ Blockly.Blocks['event_whenbroadcastreceived'] = {
           "type": "field_variable",
           "name": "BROADCAST_OPTION",
           "variableTypes": ["broadcast_msg"],
-          "variable": "message1"
+          "variable": Blockly.Msg.DEFAULT_BROADCAST_MESSAGE_NAME
         }
       ],
       "category": Blockly.Categories.event,
@@ -153,7 +153,7 @@ Blockly.Blocks['event_broadcast_menu'] = {
             "type": "field_variable",
             "name": "BROADCAST_OPTION",
             "variableTypes":["broadcast_msg"],
-            "variable":"message1"
+            "variable": Blockly.Msg.DEFAULT_BROADCAST_MESSAGE_NAME
           }
         ],
         "colour": Blockly.Colours.event.secondary,
@@ -178,7 +178,7 @@ Blockly.Blocks['event_broadcast'] = {
           "type": "field_variable",
           "name": "BROADCAST_OPTION",
           "variableTypes": ["broadcast_msg"],
-          "variable": "message1"
+          "variable": Blockly.Msg.DEFAULT_BROADCAST_MESSAGE_NAME
         }
       ],
       "category": Blockly.Categories.event,
@@ -200,7 +200,7 @@ Blockly.Blocks['event_broadcastandwait'] = {
           "type": "field_variable",
           "name": "BROADCAST_OPTION",
           "variableTypes": ["broadcast_msg"],
-          "variable":"message1"
+          "variable": Blockly.Msg.DEFAULT_BROADCAST_MESSAGE_NAME
         }
       ],
       "category": Blockly.Categories.event,
--- a/blocks_vertical/event.js
+++ b/blocks_vertical/event.js
@@ -77,13 +77,10 @@ Blockly.Blocks['event_whenbroadcastreceived'] = {
       "message0": "when I receive %1",
       "args0": [
         {
-          "type": "field_dropdown",
+          "type": "field_variable",
           "name": "BROADCAST_OPTION",
-          "options": [
-            ['message1', 'message1'],
-            ['message2', 'message2'],
-            ['new message', 'new message']
-          ]
+          "variableTypes": ["broadcast_msg"],
+          "variable": "message1"
         }
       ],
       "category": Blockly.Categories.event,
@@ -153,13 +150,10 @@ Blockly.Blocks['event_broadcast_menu'] = {
         "message0": "%1",
         "args0": [
           {
-            "type": "field_dropdown",
+            "type": "field_variable",
             "name": "BROADCAST_OPTION",
-            "options": [
-              ['message1', 'message1'],
-              ['message2', 'message2'],
-              ['new message', 'new message']
-            ]
+            "variableTypes":["broadcast_msg"],
+            "variable":"message1"
           }
         ],
         "colour": Blockly.Colours.event.secondary,
@@ -181,8 +175,10 @@ Blockly.Blocks['event_broadcast'] = {
       "message0": "broadcast %1",
       "args0": [
         {
-          "type": "input_value",
-          "name": "BROADCAST_OPTION"
+          "type": "field_variable",
+          "name": "BROADCAST_OPTION",
+          "variableTypes": ["broadcast_msg"],
+          "variable": "message1"
         }
       ],
       "category": Blockly.Categories.event,
@@ -201,8 +197,10 @@ Blockly.Blocks['event_broadcastandwait'] = {
       "message0": "broadcast %1 and wait",
       "args0": [
         {
-          "type": "input_value",
-          "name": "BROADCAST_OPTION"
+          "type": "field_variable",
+          "name": "BROADCAST_OPTION",
+          "variableTypes": ["broadcast_msg"],
+          "variable":"message1"
         }
       ],
       "category": Blockly.Categories.event,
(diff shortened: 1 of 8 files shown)


rachel-fenichel:
https://i.imgur.com/jUTHMLn.gif

I'm excited to look through this.

For variable dropdowns with multiple types, a single "new variable" option would be insufficient.  In that case I would argue for having an explicit "Create variable" button in the flyout per type of variable, or a modal that lets you pick the type.  This is one of the very underspeced parts of typed variables.

core/field_variable.js
```diff
@@ -81,7 +81,18 @@ Blockly.FieldVariable.prototype.initModel = function() {
...
-    this.sourceBlock_.workspace.createVariable(this.getValue());
+    // Check if there was exactly one element specified in the
+    // variableTypes list. This is the list that specifies which types of
+    // variables to include in the dropdown.
+    // If there is exactly one element specified, make the variable
+    // being created this specified type. Else, default behavior is to create
+    // a scalar variable
+    if (this.getVariableTypes_().length === 1) {
+      this.sourceBlock_.workspace.createVariable(this.getValue(),
+        this.getVariableTypes_()[0]);
```
rachel-fenichel: I expected that you would add an explicit type to the scalar variables ('scalar', maybe).  What's the advantage of leaving it as the empty string?

For context, here's `workspace.createVariable`: 
```
/**
 * Create a variable with a given name, optional type, and optional id.
 * @param {!string} name The name of the variable. This must be unique across
 *     variables and procedures.
 * @param {?string} opt_type The type of the variable like 'int' or 'string'.
 *     Does not need to be unique. Field_variable can filter variables based on
 *     their type. This will default to '' which is a specific type.
 * @param {?string} opt_id The unique id of the variable. This will default to
 *     a UUID.
 * @return {?Blockly.VariableModel} The newly created variable.
 */
Blockly.Workspace.prototype.createVariable = function(name, opt_type, opt_id) {
  if (name.toLowerCase() == Blockly.Variables.noVariableText()) {
    return;
  }
  return this.variableMap_.createVariable(name, opt_type, opt_id);
};
```

msg/messages.js
```diff
@@ -148,6 +148,10 @@ Blockly.Msg.CANNOT_DELETE_VARIABLE_PROCEDURE = 'Can\'t delete the variable "%1"
 /// dropdown choice - Delete the currently selected variable.
 Blockly.Msg.DELETE_VARIABLE = 'Delete the "%1" variable';
 
+// Broadcast Message creation
+/// dropdown choice - Create a new message.
+Blockly.Msg.NEW_MESSAGE = 'New message...';
```
rachel-fenichel: Let's call this "NEW_BROADCAST_MESSAGE".  Msg.NEW_MESSAGE will be confusing down the line.

blocks_vertical/event.js
```diff
@@ -77,13 +77,10 @@ Blockly.Blocks['event_whenbroadcastreceived'] = {
...
-          "type": "field_dropdown",
+          "type": "field_variable",
           "name": "BROADCAST_OPTION",
-          "options": [
-            ['message1', 'message1'],
-            ['message2', 'message2'],
-            ['new message', 'new message']
-          ]
+          "variableTypes": ["broadcast_msg"],
+          "variable": "message1"
```
rachel-fenichel: Is "message1" going to be the same for all languages?  Blockly usually uses messages here so that they can be translated.

For instance, the default name [here](https://developers.google.com/blockly/guides/create-custom-blocks/variables#untyped_variable_blocks) is given by "BKY_VARIABLES_DEFAULT_NAME"

core/field_variable.js
kchadha: I agree, we should call out scalar variables specifically instead of leaving the original empty string to specify that type. The only reason that the change hasn't been made in this PR is that it seemed like it would be a very pervasive change and out of the scope of this PR, which is specifically for adding functionality behind the dropdown options for the broadcast messages. I will file an issue for this change.

kchadha: This is now referenced in issue #1251

kchadha:
@rachel-fenichel Thank you for looking over this PR and for your comments! I just pushed some changes addressing your comments about message names for translation. I also filed a new issue to address your comment regarding calling out the type for scalar variables more specifically (rather than keeping it as an empty string). Please let me know if there are any other changes to be made. Thanks again!

core/field_variable.js
```diff
@@ -171,13 +182,17 @@ Blockly.FieldVariable.dropdownCreate = function() {
...
   }
+  var isBroadcastType = false;
   if (workspace) {
     var variableTypes = this.getVariableTypes_();
     var variableModelList = [];
     // Get a copy of the list, so that adding rename and new variable options
     // doesn't modify the workspace's list.
     for (var i = 0; i < variableTypes.length; i++) {
       var variableType = variableTypes[i];
+      if (variableType === 'broadcast_msg'){
```
rachel-fenichel: Make 'broadcast_msg' a constant, probably in constants.js.

core/xml.js
```diff
@@ -116,6 +116,21 @@ Blockly.Xml.blockToDom = function(block, opt_noId) {
...
+          // set the type for blocks that exist by default in the toolbox
+          // (e.g. broadcast messages)
+          // TODO figure out if we need to do something different when there's
+          // more than one element in variableTypes field
+
+          // must check that field is an instance of FieldVariable because
+          // FieldVariableGetter doesn't have getVariableTypes_ function
+          if (field instanceof Blockly.FieldVariable) {
+            var variableTypes = field.getVariableTypes_();
+            if (variableTypes.length === 1) {
```
rachel-fenichel: use == instead of === here and above.

(We've had the long discussion about whether == or === is more correct and decided that both can lead to bugs, but we've settled on == unless === is specifically called for.)

rachel-fenichel (approved):
Looks good overall.  Two style changes, and then feel free to merge.

For looking up variables by id instead of name: yup, that's a real problem, and it's upstream: https://github.com/google/blockly/issues/1199.  It's also the cause of bad change events for variables.  I'm looking into it for now.  

You don't need to try to fix it, but leave a note anywhere that it's problematic so that I can go through and do a cleanup once I get it right in Blockly.  You can refer to https://github.com/LLK/scratch-blocks/issues/1253

The last attempt to fix it failed and had to be rolled back: https://github.com/google/blockly/pull/1254

core/field_variable.js
kchadha: Should I also use == here instead of ===?

rachel-fenichel: Yes.

kchadha:
@rachel-fenichel just letting you know that I added some TODO comments referencing #1245 and #1253 as part of this PR in my most recent commit.

PR #1254, opened 2017-11-21
Sort the More category by procedure procCodes

### LINKED ISSUE #1250: Automatically sort custom procedure category alphabetically
### Expected/Requested Behavior

The custom procedure category is sorted alphabetically, by custom block label text. This way, it's easy to navigate: just scroll to the right letter, as you would flip through a dictionary to find a word.

### Actual Behavior

The custom procedure category is [not sorted](https://github.com/LLK/scratch-blocks/blob/f9cd4fa83344341db15c536b0b73971f75ad6ee1/core/procedures.js#L76-L93).

For comparison, here is the behavior in 2.0, as I posted [a while ago](https://github.com/LLK/scratch-vm/issues/717#issuecomment-336951813):

Scratch 2.0 doesn't really do any sorting on the list of custom blocks in this category - it tends to be finicky, sometimes putting the most recently dragged block at the top of the list and sometimes seeming to order the list based on the positions of the procedure definitions in the workspace.

### Steps to Reproduce

View the custom procedure category when multiple custom blocks are defined in it.

(Admittedly, I wasn't certain how to do this myself - I looked at the code, and saw that the procedure list isn't sorted anywhere.)

Debian (Testing), Firefox 58.0.

towerofnix:
### Resolves

Resolves #1250.

### Proposed Changes

Sorts procedure definitions in the More category by their name, so that the category acts like a dictionary.

### Reason for Changes

Convenience - finding a particular custom block is much easier when they are predictably, intuitively sorted. Also, to resolve 'help wanted' issue #1250.

### Test Coverage

No new tests added. Could be worth adding tests to see if this function sorts correctly - but I was having trouble getting the unit test setup to work, so I couldn't/can't do that myself.

### DIFF
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -93,6 +93,31 @@ Blockly.Procedures.allProcedureMutations = function(root) {
 };
 
 /**
+ * Sorts an array of procedure definition mutations alphabetically.
+ * (Does not mutate the given array.)
+ * @param {!Array.<Element>} mutations Array of mutation xml elements.
+ * @return {!Array.<Element>} Sorted array of mutation xml elements.
+ */
+Blockly.Procedures.sortProcedureMutations = function(mutations) {
+  var newMutations = mutations.slice();
+
+  newMutations.sort(function(a, b) {
+    var procCodeA = a.getAttribute('proccode');
+    var procCodeB = b.getAttribute('proccode');
+
+    if (procCodeA < procCodeB) {
+      return -1;
+    } else if (procCodeA > procCodeB) {
+      return 1;
+    } else {
+      return 0;
+    }
+  });
+
+  return newMutations;
+};
+
+/**
  * Comparison function for case-insensitive sorting of the first element of
  * a tuple.
  * @param {!Array} ta First tuple.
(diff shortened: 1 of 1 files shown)


core/procedures.js
```diff
@@ -92,6 +92,31 @@ Blockly.Procedures.allProcedureMutations = function(root) {
   return mutations;
 };
 
+/**
+ * Sorts an array of procedure definition mutations alphabetically.
+ * (Does not mutate the given array.)
+ * @param {!Array.<Element>} mutations Array of mutation xml elements.
+ * @return {!Array.<Element>} Sorted array of mutation xml elements.
+ */
+Blockly.Procedures.sortProcedureMutations = function(mutations) {
```
rachel-fenichel: This only needs to be used inside scratch blocks, so let's make it private: add an underscore at the end of the name, and add @private below the @return annotation in the comment.

towerofnix: Fixed in https://github.com/LLK/scratch-blocks/pull/1254/commits/944ac2ce3e3b63d54b3cbd33cc1704b60b77a114 / https://github.com/LLK/scratch-blocks/pull/1254/commits/abcb8388a550ce61749fadff4a744b4a40e383fc.

Issue #1194, opened 2017-10-30
Undoing procedure edits


rachel-fenichel:
Scratch 2.0 can't undo/redo procedure edits.

Should Scratch 3.0 be able to undo/redo procedure edits?  If not, what's the specific reason to avoid it?

towerofnix:
First thoughts are that it definitely should support undoing a procedure edit; after all, it's possible one might accidentally make an edit which *removes* an input, thus messing up old scripts. In that case, being able to undo would be very helpful.

PR #1259, opened 2017-11-22
Add logic for edit/create custom procedures, wire up the playground

paulkaplan:
[work in progress, wanted to get it up so people could start taking a look]

This PR implements the `editProcedureCallback_` and `createProcedureCallback_` by wiring them to a common `externalProcedureDefCallback_`. This new function is just a stub, intended to be overwritten externally. An example implementation is then added to the custom procedure playground. 

new-custom-procedure-playground

I'd like to use this as a starting point for talking about how communication should work. Right now the external function gets the mutator from the procedure block of a given procedure, and a callback to call with an updated version of that mutator.

### DIFF
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -360,7 +360,7 @@ Blockly.Procedures.newProcedureMutation = function() {
  * @private
  */
 Blockly.Procedures.createProcedureDefCallback_ = function(workspace) {
-  Blockly.Procedures.externalProcedureDefCallback_(
+  Blockly.Procedures.externalProcedureDefCallback(
     Blockly.Procedures.newProcedureMutation(),
     Blockly.Procedures.createProcedureCallbackFactory_(workspace)
   );
@@ -426,7 +426,7 @@ Blockly.Procedures.editProcedureCallback_ = function(block) {
       block.getProcCode(), block.workspace);
   }
   // Block now refers to the procedure prototype block, it is safe to proceed.
-  Blockly.Procedures.externalProcedureDefCallback_(
+  Blockly.Procedures.externalProcedureDefCallback(
     block.mutationToDom(),
     Blockly.Procedures.editProcedureCallbackFactory_(block)
   );
@@ -449,10 +449,10 @@ Blockly.Procedures.editProcedureCallbackFactory_ = function(block) {
 
 /**
  * Callback to create a new procedure custom command block.
- * @private
+ * @public
  */
-Blockly.Procedures.externalProcedureDefCallback_ = function(/** mutator, callback */) {
-  alert('External procedure editor must be override Blockly.Procedures.externalProcedureDefCallback_');
+Blockly.Procedures.externalProcedureDefCallback = function(/** mutator, callback */) {
+  alert('External procedure editor must be override Blockly.Procedures.externalProcedureDefCallback');
 };
 
 /**
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -276,31 +276,33 @@ Blockly.Procedures.getCallers = function(name, ws, definitionRoot,
 };
 
 /**
- * When a procedure definition changes its parameters, find and edit all its
- * callers.
- * @param {!Blockly.Block} defBlock Procedure definition block.
+ * Find and edit all callers with a procCode using a new mutation.
+ * @param {string} name Name of procedure (procCode in scratch-blocks).
+ * @param {!Blockly.Workspace} ws The workspace to find callers in.
+ * @param {!Element} mutation New mutation for the callers.
  */
-Blockly.Procedures.mutateCallers = function(defBlock) {
-  // TODO(#1143) Update this for scratch procedures.
-  var oldRecordUndo = Blockly.Events.recordUndo;
-  var name = defBlock.getProcedureDef()[0];
-  var xmlElement = defBlock.mutationToDom(true);
-  var callers = Blockly.Procedures.getCallers(name, defBlock.workspace);
-  for (var i = 0, caller; caller = callers[i]; i++) {
-    var oldMutationDom = caller.mutationToDom();
-    var oldMutation = oldMutationDom && Blockly.Xml.domToText(oldMutationDom);
-    caller.domToMutation(xmlElement);
-    var newMutationDom = caller.mutationToDom();
-    var newMutation = newMutationDom && Blockly.Xml.domToText(newMutationDom);
-    if (oldMutation != newMutation) {
-      // Fire a mutation on every caller block.  But don't record this as an
-      // undo action since it is deterministically tied to the procedure's
-      // definition mutation.
-      Blockly.Events.recordUndo = false;
-      Blockly.Events.fire(new Blockly.Events.BlockChange(
-          caller, 'mutation', null, oldMutation, newMutation));
-      Blockly.Events.recordUndo = oldRecordUndo;
+Blockly.Procedures.mutateCallersAndPrototype = function(name, ws, mutation) {
+  var defineBlock = Blockly.Procedures.getDefineBlock(name, ws);
+  var prototypeBlock = Blockly.Procedures.getPrototypeBlock(name, ws);
+  if (defineBlock && prototypeBlock) {
+    var callers = Blockly.Procedures.getCallers(name,
+        defineBlock.workspace, defineBlock, true /* allowRecursive */);
+    callers.push(prototypeBlock);
+    Blockly.Events.setGroup(true);
+    for (var i = 0, caller; caller = callers[i]; i++) {
+      var oldMutationDom = caller.mutationToDom();
+      var oldMutation = oldMutationDom && Blockly.Xml.domToText(oldMutationDom);
+      caller.domToMutation(mutation);
+      var newMutationDom = caller.mutationToDom();
+      var newMutation = newMutationDom && Blockly.Xml.domToText(newMutationDom);
+      if (oldMutation != newMutation) {
+        Blockly.Events.fire(new Blockly.Events.BlockChange(
+            caller, 'mutation', null, oldMutation, newMutation));
+      }
     }
+    Blockly.Events.setGroup(false);
+  } else {
+    alert('No define block on workspace'); // TODO decide what to do about this.
   }
 };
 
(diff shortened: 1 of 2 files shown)


core/procedures.js
```diff
@@ -358,9 +424,58 @@ Blockly.Procedures.editProcedureCallback_ = function(block) {
...
+      }
+      // And update the prototype block
+      block.domToMutation(mutation);
+    }
+  };
+};
+
+/**
+ * Callback to create a new procedure custom command block.
+ * @private
```
rachel-fenichel: Shouldn't be private, since you're explicitly telling people to override it.

core/procedures.js
```diff
@@ -306,40 +306,106 @@ Blockly.Procedures.mutateCallers = function(defBlock) {
 
 /**
  * Find the definition block for the named procedure.
- * @param {string} name Name of procedure.
+ * @param {string} procCode The identifier of the procedure to delete.
```
rachel-fenichel: Nothing about this forces it to be a deletion.

core/procedures.js
```diff
@@ -306,40 +306,106 @@ Blockly.Procedures.mutateCallers = function(defBlock) {
...
  * @param {!Blockly.Workspace} workspace The workspace to search.
  * @return {Blockly.Block} The procedure definition block, or null not found.
  */
-Blockly.Procedures.getDefinition = function(name, workspace) {
+Blockly.Procedures.getDefineBlock = function(procCode, workspace) {
```
rachel-fenichel: This whole function can be replaced by finding the prototype block and then getting its parent.  Or vice versa.

core/procedures.js
```diff
@@ -358,9 +424,58 @@ Blockly.Procedures.editProcedureCallback_ = function(block) {
       return;
     }
     block = innerBlock;
+  } else if (!block.type == Blockly.PROCEDURES_PROTOTYPE_BLOCK_TYPE){
```
rachel-fenichel: The above logic checks whether the right inputs, connections, etc. exists.  It can go into `getPrototypeBlock`.  Then this becomes

```
if (block.type == Blockly.PROCEDURES_DEFINITION_BLOCK_TYPE ||
    block.type == Blockly.PROCEDURES_CALL_BLOCK_TYPE) {
  block = Blockly.Procedures.getPrototypeBlock(block.getProcCode(), block.workspace)
}
```

core/procedures.js
```diff
@@ -358,9 +424,58 @@ Blockly.Procedures.editProcedureCallback_ = function(block) {
...
+
+/**
+ * Callback factory for editing an existing custom procedure.
+ * @param {!Blockly.Block} block The procedure prototype block being edited.
+ * @return {function(?Element)} Callback for editing the custom procedure.
+ * @private
+ */
+Blockly.Procedures.editProcedureCallbackFactory_ = function(block) {
+  return function(mutation) {
+    if (mutation) {
```
rachel-fenichel, line 451: This belongs in Blockly.Procedures.mutateCallers: https://github.com/LLK/scratch-blocks/blob/develop/core/procedures.js#L311

core/procedures.js
```diff
@@ -358,9 +417,58 @@ Blockly.Procedures.editProcedureCallback_ = function(block) {
...
+      // Update all the callers
+      var defineBlock = Blockly.Procedures.getDefineBlock(block.getProcCode(),
+        block.workspace);
+      if (defineBlock) {
+        var callers = Blockly.Procedures.getCallers(block.getProcCode(),
```
rachel-fenichel: This use of getCallers is kind of strained, since you had to go find the defineBlock even though you want to search in the entire workspace.  I added https://github.com/LLK/scratch-blocks/issues/1264 so we can circle back and clean it up later.

core/procedures.js
```diff
@@ -358,9 +417,58 @@ Blockly.Procedures.editProcedureCallback_ = function(block) {
...
+          if (oldMutation != mutation) {
+            Blockly.Events.fire(new Blockly.Events.BlockChange(
+                caller, 'mutation', null, oldMutation, mutation));
+          }
+        }
+      } else {
+        alert('No define block on workspace'); // TODO decide what to do about this.
+      }
+      // And update the prototype block
+      block.domToMutation(mutation);
```
rachel-fenichel: Does this also need an event?

tests/custom_procedure_playground.html
```diff
@@ -193,8 +204,14 @@
...
           break;
       }
+    }
 
-      applyMutation();
+    function cancel() {
+      callback = null;
+      mutationRoot = null;
+      declarationWorkspace.clear();
+      definitionWorkspace.refreshToolboxSelection_()
```
rachel-fenichel, line 213: No need to refresh the toolbox selection here.

core/procedures.js
```diff
@@ -306,40 +306,99 @@ Blockly.Procedures.mutateCallers = function(defBlock) {
...
+  }
+  return null;
+};
+
+/**
+ * Create a mutation for a brand new custom procedure.
+ * @return {Element} The mutation for a new custom procedure
+ */
+Blockly.Procedures.newProcedureMutation = function() {
+  var mutation = goog.dom.createDom('mutation');
```
rachel-fenichel: This can also be done with textToDom, especially since it's always the same.

rachel-fenichel (changes requested):

core/procedures.js
paulkaplan: Unfortunately the define block can't get the procCode, so we can't use getPrototypeBlock when all you have is the definition block.

tests/custom_procedure_playground.html
paulkaplan: That is done because the procedures category needs to be updated because of the change in blocks, and mutateCallers doesn't reach into the toolbox workspace to change those. Should it?

paulkaplan:
@rachel-fenichel I updated this PR with more commits splitting out the logic into mutateCallers and whatnot. I tried to integrate some of the undo logic, but I think a more complete inspection of undoing should be undertaken separately.

tests/custom_procedure_playground.html
rachel-fenichel: Good point.  This is fine.

core/procedures.js
```diff
@@ -358,9 +423,39 @@ Blockly.Procedures.editProcedureCallback_ = function(block) {
...
+ * Callback factory for editing an existing custom procedure.
+ * @param {!Blockly.Block} block The procedure prototype block being edited.
+ * @return {function(?Element)} Callback for editing the custom procedure.
+ * @private
+ */
+Blockly.Procedures.editProcedureCallbackFactory_ = function(block) {
+  return function(mutation) {
+    if (mutation) {
+      Blockly.Procedures.mutateCallersAndPrototype(block.getProcCode(),
+        block.workspace, mutation);
```
rachel-fenichel: Nit: indent +2

core/procedures.js
```diff
@@ -82,7 +82,7 @@ Blockly.Procedures.allProcedureMutations = function(root) {
   var blocks = root.getAllBlocks();
   var mutations = [];
   for (var i = 0; i < blocks.length; i++) {
-    if (blocks[i].type === 'procedures_prototype') {
+    if (blocks[i].type == 'procedures_prototype') {
```
rachel-fenichel: Use a constant for `procedures_prototype`.

core/procedures.js
```diff
@@ -276,70 +276,135 @@ Blockly.Procedures.getCallers = function(name, ws, definitionRoot,
...
     }
   }
   return null;
 };
 
+/**
+ * Find the prototype block for the named procedure.
+ * @param {string} procCode The identifier of the procedure.
+ * @param {!Blockly.Workspace} workspace The workspace to search.
+ * @return {Blockly.Block} The procedure prototype block, or null not found.
```
rachel-fenichel, line 336: Annotate with `@private, @package, or @public`.  For this and all other functions you created.  Bonus points if you add annotations to functions that you touched.

core/procedures.js
```diff
@@ -276,70 +276,135 @@ Blockly.Procedures.getCallers = function(name, ws, definitionRoot,
...
+
 /**
  * Callback to create a new procedure custom command block.
- * TODO(#1299): Implement.
+ * @param {!Blockly.Workspace} workspace The workspace to create the new procedure on.
+ * @private
+ */
+Blockly.Procedures.createProcedureDefCallback_ = function(workspace) {
+  Blockly.Procedures.externalProcedureDefCallback(
+    Blockly.Procedures.newProcedureMutation(),
```
rachel-fenichel: Nit: indent +2, here and on the next lines.

rachel-fenichel: Matching pattern three here: https://google.github.io/styleguide/jsguide.html#formatting-function-arguments

core/procedures.js
```diff
@@ -358,9 +423,39 @@ Blockly.Procedures.editProcedureCallback_ = function(block) {
       return;
     }
     block = innerBlock;
+  } else if (block.type == Blockly.PROCEDURES_CALL_BLOCK_TYPE) {
+    // This is a call block, find the prototype corresponding to the procCode
+    block = Blockly.Procedures.getPrototypeBlock(
+      block.getProcCode(), block.workspace);
```
rachel-fenichel: Nit: indentation.

core/procedures.js
```diff
@@ -358,9 +423,39 @@ Blockly.Procedures.editProcedureCallback_ = function(block) {
...
   }
-  alert('TODO(#603): implement function editing (procCode was "' +
-      block.procCode_ + '")');
+  // Block now refers to the procedure prototype block, it is safe to proceed.
+  Blockly.Procedures.externalProcedureDefCallback(
+    block.mutationToDom(),
```
rachel-fenichel: Nit: indentation.

rachel-fenichel (commented):
Overall looks good--I'm down to the nits on this review.

core/procedures.js
rachel-fenichel: Good point.

rachel-fenichel:
I agree that we should look into undo separately.  Tracked in https://github.com/LLK/scratch-blocks/issues/1194

The best way to approach events will be to come up with a set of test cases (sequences of actions) and then check that they can be undone and redone in a meaningful way.

paulkaplan:
Updated those nits, thanks @rachel-fenichel

rachel-fenichel (approved):
Ship it

Issue #600, opened 2016-09-15
Lists color


tmickel:
In Scratch 1.4 and 2.0, list blocks are a different color from variables. In the current spec they are all the same color. Is that wanted?

towerofnix:
IMO they're totally different data types and should have different colors to reflect this.

jeandrek:
@liam4 I agree that they should have different colours, but I don't think variables are data types. Text strings and numbers are data types.

How about having "variables" and "lists" as categories in the palette?

towerofnix:
Yeah, they're rather different data _containers_, not types, oops!

> How about having "variables" and "lists" as categories in the palette?

But then we'd need another category, because nine categories is.. a sad number. :P

edit: actually, _eleven_ categories! Wow.

carljbowman:
Updated the blocks color spec to include a Lists color among the Data Category

![r3_vertical-colors](https://cloud.githubusercontent.com/assets/3409578/21144386/70d41c78-c118-11e6-9cd4-2ffff8e58e58.png)

<img width="413" alt="data colors" src="https://cloud.githubusercontent.com/assets/3409578/21144392/734bc906-c118-11e6-93a3-4c46b721f3a3.png">

### Data - Vars
Primary: `#FF8C1A`
Secondary: `#FF8000 `
Tertiary: `#DB6E00 `

### Data - Lists
Primary: `#FF661A `
Secondary: `#FF5500 `
Tertiary: `#E64D00 `

thisandagain:
Looks great. Thanks @carljbowman

mrjacobbloom:
Whenever this is implemented, remember to update `Blockly.Categories` in [core/constants.js](https://github.com/LLK/scratch-blocks/blob/d7745b12fdb5ee9a4761b7e99fd201673e0e7a8e/core/constants.js#L252) so that lists can be styled separately from variable blocks with CSS

thisandagain:
Tracking implementation in https://github.com/LLK/scratch-blocks/issues/1257

PR #1267, opened 2017-11-29
Changing color of list blocks 

### LINKED ISSUE #1257: List blocks should be a different color from variable blocks.
### Expected Behavior

List blocks and variable blocks should be different colors because lists and variables are able to share names. See discussion in #1245 for more info.

### Actual Behavior

Currently, list blocks and variable blocks are the same color.

### Steps to Reproduce

Create a new variable.
Create a new list.

kchadha:
Changing color of list blocks to differentiate from variable blocks.

### Resolves

Resolves #1257 

### Proposed Changes

Change the color of list blocks to differentiate from scalar variables according to color chart in #600 and discussion in #1257.

### Reason for Changes

Lists and variables will be able to share names, so color is a visible way of differentiating between the blocks that would otherwise look the same (e.g. variable getter).

### Test Coverage

Existing tests pass.

### DIFF
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -166,7 +166,7 @@ Blockly.Blocks['data_listcontents'] = {
         }
       ],
       "category": Blockly.Categories.data,
-      "extensions": ["colours_data", "output_string"],
+      "extensions": ["colours_data_lists", "output_string"],
       "checkboxInFlyout": true
     });
   }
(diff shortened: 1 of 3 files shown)


rachel-fenichel (approved):

mrjacobbloom:
Note that the data-category attribute for list blocks is the same despite a color difference, that should probably change or else list blocks will only be theme-able in the same color as scalar blocks

kchadha:
@mrjacobbloom I'm not sure what you mean by "theme-able"...

mrjacobbloom:
The SVG <g> element for each block has an attribute data-category that reflects the category property in the object passed to jsonInit so that blocks can be styled by category with plain old CSS. Right now both scalar and list blocks are using the category `Blockly.Categories.data` but we should probably add a lists category to differentiate now that they're different colors. I can do that in a separate PR though.

kchadha:
Ah, okay. If you're okay with taking that on, I will merge this PR.

PR #1269, opened 2017-11-29
Custom procedure remove button

### LINKED ISSUE #1241: Procedures: delete icon
The delete icon is a singleton--it's always associated with the selected argument or label.

image

Who owns it: the field being edited?  The procedure declaration editor?

When is it visible?  When a component is selected *and* there are multiple components on the block.

How does it know which component it's associated with?  It needs that to delete the correct component.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1241
### Proposed Changes

One commit on top of the other custom procedures work that hasn't landed yet, so only look at the last commit https://github.com/LLK/scratch-blocks/commit/49ad421605001258e499dfd2dcbad00b5652c803

I wasn't sure how to chain together the click action for the remove button with the actual declaration removal code, so that is definitely something I want feedback on @rachel-fenichel.

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -599,7 +599,7 @@ Blockly.ScratchBlocks.ProcedureUtils.removeFieldCallback = function(field) {
     return;
   }
   var inputNameToRemove = null;
-  for(var n = 0; n < this.inputList.length; n++) {
+  for (var n = 0; n < this.inputList.length; n++) {
     var input = this.inputList[n];
     if (input.connection) {
       var target = input.connection.targetBlock();
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -263,7 +263,7 @@ Blockly.ScratchBlocks.ProcedureUtils.addLabelField_ = function(text) {
 Blockly.ScratchBlocks.ProcedureUtils.addLabelEditor_ = function(text) {
   if (text) {
     this.appendDummyInput(Blockly.utils.genUid()).
-        appendField(new Blockly.FieldTextInput(text));
+        appendField(new Blockly.FieldTextInputRemovable(text));
   }
 };
 
@@ -588,6 +588,53 @@ Blockly.ScratchBlocks.ProcedureUtils.addStringNumberExternal = function() {
   this.focusLastEditor_();
 };
 
+/**
+ * Callback to remove a field, only for the declaration block.
+ * @param {Blockly.Field} field The field being removed.
+ * @public
+ */
+Blockly.ScratchBlocks.ProcedureUtils.removeFieldCallback = function(field) {
+  // Do not delete if there is only one input
+  if (this.inputList.length === 1) {
+    return;
+  }
+
+  for(var n = 0; n < this.inputList.length; n++) {
+    var input = this.inputList[n];
+    var name = input.name;
+    if (input.connection) {
+      var target = input.connection.targetBlock();
+      if (target.getField(field.name) == field) {
+        Blockly.WidgetDiv.hide(true);
+        this.removeInput(name);
+        this.onChangeFn();
+        this.updateDisplay_();
+      }
+    } else {
+      for (var j = 0; j < input.fieldRow.length; j++) {
+        if (input.fieldRow[j] == field) {
+          Blockly.WidgetDiv.hide(true);
+          this.removeInput(name);
+          this.onChangeFn();
+          this.updateDisplay_();
+        }
+      }
+    }
+  }
+};
+
+/**
+ * Callback to pass removeField up to the declaration block from arguments.
+ * @param {Blockly.Field} field The field being removed.
+ * @public
+ */
+Blockly.ScratchBlocks.ProcedureUtils.removeArgumentCallback_ = function(
+    field) {
+  if (this.parentBlock_ && this.parentBlock_.removeFieldCallback) {
+    this.parentBlock_.removeFieldCallback(field);
+  }
+};
+
 Blockly.Blocks['procedures_definition'] = {
   /**
    * Block for defining a procedure with no return value.
@@ -705,6 +752,9 @@ Blockly.Blocks['procedures_declaration'] = {
   populateArgument_: Blockly.ScratchBlocks.ProcedureUtils.populateArgumentOnDeclaration_,
   addProcedureLabel_: Blockly.ScratchBlocks.ProcedureUtils.addLabelEditor_,
 
+  // Exist on declaration and arguments editors, with different implementations.
+  removeFieldCallback: Blockly.ScratchBlocks.ProcedureUtils.removeFieldCallback,
+
   // Only exist on procedures_declaration.
   createArgumentEditor_: Blockly.ScratchBlocks.ProcedureUtils.createArgumentEditor_,
   focusLastEditor_: Blockly.ScratchBlocks.ProcedureUtils.focusLastEditor_,
@@ -749,7 +799,7 @@ Blockly.Blocks['argument_editor_boolean'] = {
     this.jsonInit({ "message0": " %1",
       "args0": [
         {
-          "type": "field_input",
+          "type": "field_input_removable",
           "name": "TEXT",
           "text": "foo"
         }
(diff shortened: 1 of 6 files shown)


blocks_vertical/procedures.js
```diff
@@ -565,6 +565,53 @@ Blockly.ScratchBlocks.ProcedureUtils.addStringNumberExternal = function() {
...
+      if (target.getField(field.name) == field) {
+        Blockly.WidgetDiv.hide(true);
+        this.removeInput(name);
+        this.onChangeFn();
+        this.updateDisplay_();
+      }
+    } else {
+      for (var j = 0; j < input.fieldRow.length; j++) {
+        if (input.fieldRow[j] == field) {
+          Blockly.WidgetDiv.hide(true);
```
rachel-fenichel: You're doing the same things here and in lines 585-588.  Instead have a variable `nameToRemove` declared outside of the loop and set to null.  In the loop, if the checks pass, set `nameToRemove` to `name` and break.  Then do all of these steps based on `nameToRemove`, if it's not null.

core/block.js
```diff
@@ -1477,6 +1480,23 @@ Blockly.Block.newFieldTextInputFromJson_ = function(options) {
   return field;
 };
 
+/**
```
rachel-fenichel: Move this into your new field file.  I just filed https://github.com/google/blockly/issues/1486 with an explanation + plans to do the same for the default field types.

core/css.js
```diff
@@ -163,7 +163,7 @@ Blockly.Css.CONTENT = [
   '}',
 
   '.blocklyWidgetDiv.fieldTextInput {',
-    'overflow: hidden;',
+    // 'overflow: hidden;',
```
rachel-fenichel: ?

core/field_textinput_removable.js
```diff
@@ -0,0 +1,88 @@
+/**
+ * @license
+ * Visual Blocks Editor
+ *
+ * Copyright 2012 Google Inc.
```
rachel-fenichel: Update license year, fileoverview, and author.

core/field_textinput_removable.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * @fileoverview Text input field.
+ * @author fraser@google.com (Neil Fraser)
+ */
+'use strict';
+
+goog.provide('Blockly.FieldTextInputRemovable');
+
+goog.require('Blockly.FieldTextInput');
```
rachel-fenichel: Alphabetize the Blockly.* requires, per https://google.github.io/styleguide/jsguide.html#file-goog-require

core/field_textinput_removable.js
```diff
@@ -0,0 +1,88 @@
...
+goog.require('Blockly.Msg');
+goog.require('Blockly.utils');
+goog.require('goog.asserts');
+goog.require('goog.dom');
+goog.require('goog.dom.TagName');
+goog.require('goog.userAgent');
+
+
+/**
+ * Class for an editable text field.
```
rachel-fenichel: Class for an editable text field which displays a deletion icon when selected.

core/field_textinput_removable.js
```diff
@@ -0,0 +1,88 @@
...
+ * @param {string} text The initial content of the field.
+ * @param {Function=} opt_validator An optional function that is called
+ *     to validate any constraints on what the user entered.  Takes the new
+ *     text as an argument and returns either the accepted text, a replacement
+ *     text, or null to abort the change.
+ * @param {RegExp=} opt_restrictor An optional regular expression to restrict
+ *     typed text to. Text that doesn't match the restrictor will never show
+ *     in the text field.
+ * @extends {Blockly.Field}
```
rachel-fenichel: @extends {Blockly.FieldTextInput}

core/field_textinput_removable.js
```diff
@@ -0,0 +1,88 @@
...
+ * @private
+ */
+Blockly.FieldTextInputRemovable.prototype.showEditor_ = function() {
+  Blockly.FieldTextInputRemovable.superClass_.showEditor_.call(this);
+
+  var div = Blockly.WidgetDiv.DIV;
+  var removeButton =
+      goog.dom.createDom(goog.dom.TagName.IMG, 'blocklyTextDropDownArrow');
+  removeButton.setAttribute('src',
+    Blockly.mainWorkspace.options.pathToMedia + 'icons/remove.svg');
```
rachel-fenichel: Indent +2 spaces

core/field_textinput_removable.js
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Show the inline free-text editor on top of the text with the remove button.
+ * @private
+ */
+Blockly.FieldTextInputRemovable.prototype.showEditor_ = function() {
+  Blockly.FieldTextInputRemovable.superClass_.showEditor_.call(this);
+
+  var div = Blockly.WidgetDiv.DIV;
+  var removeButton =
+      goog.dom.createDom(goog.dom.TagName.IMG, 'blocklyTextDropDownArrow');
```
rachel-fenichel: `blocklyTextDropdownArrow`?

core/field_textinput_removable.js
```diff
@@ -0,0 +1,88 @@
...
+  removeButton.setAttribute('src',
+    Blockly.mainWorkspace.options.pathToMedia + 'icons/remove.svg');
+  removeButton.style.width = 25 + 'px';
```
rachel-fenichel: These are all constants.  Can you move them into CSS?

core/field_textinput_removable.js
```diff
@@ -0,0 +1,88 @@
...
+  removeButton.style['margin-left'] = -15 + 'px';
+
+  this.removeButtonMouseWrapper_ = Blockly.bindEvent_(removeButton,
+    'mousedown', this, this.removeCallback_);
+  div.appendChild(removeButton);
+};
+
+Blockly.FieldTextInput.prototype.removeCallback_ = function() {
+  if (this.sourceBlock_ && this.sourceBlock_.removeFieldCallback) {
+    this.sourceBlock_.removeFieldCallback(this);
```
rachel-fenichel: This looks reasonable.  Add an else that warns the developer that they should have implemented removeFieldCallback on the source block, and of course add documentation.

core/block.js
paulkaplan: If I do that, does this file need to be updated to actually use that function? Or is that done automatically somehow?

core/css.js
paulkaplan: Woops fixed with custom class

core/block.js
rachel-fenichel: You would change line 1346 (which you just added) to call the new function.

blocks_vertical/procedures.js
```diff
@@ -575,29 +575,28 @@ Blockly.ScratchBlocks.ProcedureUtils.removeFieldCallback = function(field) {
   if (this.inputList.length === 1) {
     return;
   }
-
+  var inputNameToRemove = null;
   for(var n = 0; n < this.inputList.length; n++) {
```
rachel-fenichel: nit: space between `for` and `(`

rachel-fenichel (commented):
This differs from the scratch 2.0 behaviour in that the remove icon is always shown, even if you can't actually remove the input.  Scratch 2.0 hides the icon if it's not usable.  Are you planning to implement that?

paulkaplan:
Thanks for all the feedback. I updated with more commits fixing the comments. 

WRT the remove icon always showing, I will return to that in another pass. I'll file an issue.

core/field_textinput_removable.js
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Show the inline free-text editor on top of the text with the remove button.
+ * @private
+ */
+Blockly.FieldTextInputRemovable.prototype.showEditor_ = function() {
+  Blockly.FieldTextInputRemovable.superClass_.showEditor_.call(this);
+
+  var div = Blockly.WidgetDiv.DIV;
+  div.className += ' removableTextInput';
```
rachel-fenichel, line 64: Does this get removed at the correct time?  Or does the widget div continue to have this forever?

rachel-fenichel (approved):
One question, and LGTM

core/field_textinput_removable.js
paulkaplan: It gets removed correctly! Followed the pattern from the fieldtextinput which does a similar thing. It gets removed when the widget is hidden.

Issue #1207, opened 2017-11-03
Add Extension Icon support to block rendering


carljbowman:
Splitting from https://github.com/LLK/scratch-gui/issues/815

### Icon on blocks spec
This is a hi-res (300%) PNG for zooming on measurements. Extension Icons on blocks will need some special treatment when it comes to alignment and spacing.

In all cases the Vertical height of extension blocks will be taller than their "standard" block counter-parters. E.g. Command blocks with extension icons are 8px (2units) taller than a Command block without an extension icon.

**Something to Consider**
We will have similar needs to dynamically grow blocks vertically with "[input first](https://github.com/LLK/scratch-blocks/issues/1206)" blocks. Some languages reorder block parameters (e.g. Korean) and place inputs as the first element on a block. In these cases we will need to add addtional vertical padding so that the top notch does not overlap the input.

![r1_extension-icon-spec](https://user-images.githubusercontent.com/3409578/32376958-a8b1d13a-c07c-11e7-9833-e13701a0baf8.png)

rachel-fenichel:
Input first blocks will still have the icon before the first input, right?  If so, the first input on an extension block will be past the notch (based on position of icon + bar).

rachel-fenichel:
Does the vertical bar get taller as the block gets taller?  For instance, in this block would the vertical bar's height be based on a distance from top and bottom, or is it a fixed height?
![image](https://user-images.githubusercontent.com/13686399/32514062-be7e0c90-c3b0-11e7-8796-cafff45e85a7.png)

carljbowman:
@rachel-fenichel - Thanks for the initial thoughts and questions.

1. Extension icons should always be the first element on extension blocks. So the "input first" stuff is not a real issue for anything that is an Extension.

2. Thanks for clarifying. No, the vertical bar does not need to get taller as blocks are nested. That can be a fixed height.

rachel-fenichel:
For a stack block with an extension icon, the icon is moved down by one grid unit from where it would normally be.  Does that one-grid-unit offset remain as the block gets taller?  E.g.
![image](https://user-images.githubusercontent.com/13686399/34015746-4c19ee26-e0d4-11e7-87f9-2810642d2643.png)

Spec:
![image](https://user-images.githubusercontent.com/13686399/34015774-626b9760-e0d4-11e7-80f4-94d54f299428.png)

PR #1284, opened 2017-12-05
Extension block rendering

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/1207

Note that this PR does not include the vertical separator line after the icon shown in the spec- I'll make a separate issue to add that part.

### Proposed Changes

Update the blocks rendering to bring blocks with extension icons on them up to spec. This includes:

- Increasing the height of command and reporter blocks
- Adding horizontal offset to the icon in some cases
- Adding vertical offset to the icon in some cases
 
### Reason for Changes

To accommodate the icon on extension blocks at the size we want, we needed to make some layout changes to the blocks.

### Test Coverage

I've added an extension blocks category to the vertical playground via the default toolbox, to visually test out the rendering.


rachel-fenichel (commented):
Before I dive too much into the code:
- This change effectively makes any block with a leading image render as an extension block.  Is that a design that you're comfortable with?  If so, it needs to be communicated to anyone developing with the library.
- Please handle RTL, and test in RTL.

ericrosenbaum:
@rachel-fenichel re: communication, I guess it's more like: any block with a leading image will be rendered in such a way that it is laid out properly according to our spec (there's nothing extension-specific in particular about the rendering- it's just making the blocks sized to fit the images properly). The code I wrote assumes that these leading images are 40x40px. In our spec for designers, we will of course explain this requirement as well.   

I'll have a look at RTL- I forgot to do that, and it's definitely not quite right yet. It's more noticeable once you get to my next PR with the vertical separator lines.

rachel-fenichel:
Right, I just want to be clear that the spec is effectively for all blocks with leading images, rather than being a spec for extension blocks specifically.  Similarly, the next PR makes all blocks that lead with an image have a bar after the image, regardless of whether it's intended to be an extension block.

Thanks for fixing RTL.  I'll take a closer look at the code now.

blocks_vertical/extensions.js
```diff
@@ -0,0 +1,211 @@
...
+  /**
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": "%1 play a drum",
+      "args0": [
+        {
+          "type": "field_image",
+          "src": Blockly.mainWorkspace.options.pathToMedia + "extensions/music-block-icon.svg",
```
rachel-fenichel, line 62: The Blockly documentation [says](https://developers.google.com/blockly/guides/create-custom-blocks/define-blocks#image) that you can use a relative path here, so you shouldn't need to include the `Blockly.mainWorkspace.options.pathToMedia` part.

blocks_vertical/extensions.js
```diff
@@ -0,0 +1,211 @@
...
+goog.require('Blockly.Colours');
+goog.require('Blockly.constants');
+goog.require('Blockly.ScratchBlocks.VerticalExtensions');
+
+
+Blockly.Blocks['extension_pen_down'] = {
+  /**
+   * @this Blockly.Block
+   */
+  init: function() {
```
rachel-fenichel, line 35: You can use Blockly.defineBlocksWithJsonArray for this: https://github.com/LLK/scratch-blocks/blob/develop/core/blockly.js#L358

core/block_render_svg_vertical.js
```diff
@@ -631,6 +631,14 @@ Blockly.BlockSvg.prototype.render = function(opt_bubble) {
   // If there are no icons, cursorX will be 0, otherwise it will be the
   // width that the first label needs to move over by.
 
+  // If the first field is an image, as in extension blocks, and this is a
+  // reporter block, add a horizontal offset.
+  if ((this.inputList[0].fieldRow[0] instanceof Blockly.FieldImage) &&
+    this.outputConnection) {
```
rachel-fenichel: Indent continuation lines +4 spaces, here and everywhere: https://google.github.io/styleguide/jsguide.html#formatting-indent

core/block_render_svg_vertical.js
```diff
@@ -681,6 +689,14 @@ Blockly.BlockSvg.prototype.renderFields_ =
     // Offset the field upward by half its height.
     // This vertically centers the fields around cursorY.
     var yOffset = -field.getSize().height / 2;
+
+    // If the first field is an image, as in extension blocks, and this field is the first field,
```
rachel-fenichel: Nit: line wrap at 80

rachel-fenichel (commented):
I have a few style nits, but LGTM after that.

blocks_vertical/extensions.js
ericrosenbaum: We're using this pattern across 13 files- that path variable has been set to "../media/", so it's already relative. Is it okay to leave it as is?

blocks_vertical/extensions.js
ericrosenbaum: I just followed the pattern of the rest of the blocks definition files in that directory. It looks like defineBlocksWithJsonArray is used mainly in the unit tests. Is it okay if I leave it as is? Or is there some reason that defineBlocksWithJsonArray is better?

core/block_render_svg_vertical.js
ericrosenbaum: all set

core/block_render_svg_vertical.js
ericrosenbaum: all set

blocks_vertical/extensions.js
rachel-fenichel: Discussed offline; doesn't need to happen in this PR, but should happen eventually.

blocks_vertical/extensions.js
rachel-fenichel: There's a code smell here, but not one that we need to handle now.  Leave as is.

rachel-fenichel:
After offline discussion: this is fine for now, and as a followup Eric will use an `isScratchExtension` flag to check if it's an extension block, instead of checking whether it starts with an image.

rachel-fenichel (approved):

PR #1293, opened 2017-12-07
Case sensitive variables

### LINKED ISSUE #1008: Cannot open project with both "X" and "x" variables
### Expected Behavior

Should be able to load a project like this one: https://llk.github.io/scratch-gui/develop/#189524895

### Actual Behavior

It fails to load with this error: 

```
Uncaught (in promise) Error: Variable "X" is already in use and its id is "Q:(NMDNi+=uGHv?K1nO^-x" which conflicts with the passed in id, "Q:(NMDNi+=uGHv?K1nO^-X"
```

### Steps to Reproduce

I made another repro with a simplified case: try to import a project with an uppercase and lowercase version of the same variable, like this one: 

image

https://llk.github.io/scratch-gui/develop/#191677809

---

Tracking down the root cause, it looks like scratch-blocks is lowercasing variable names to check for uniqueness. Scratch 2 has case-sensitive variable names, so I assume scratch 3 wants to also. We can create an issue on scratch-blocks once we confirm that we want that behavior and it'll be ok to fix it that way in scratch-blocks.

### LINKED ISSUE #1292: "Used name" checking for variables should be case-insensitive
As noted in https://developers.google.com/blockly/guides/app-integration/best-practices#4_variable_and_function_names (h/t @mrjacobbloom), scratch differs from blockly in allowing variable names to be case sensitive.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1292
Fixes https://github.com/LLK/scratch-gui/issues/1008

### Proposed Changes

Makes variable naming case sensitive. 

### Reason for Changes

### Test Coverage

Added a test in variable creation for names that differ only in case.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -840,7 +840,7 @@ Blockly.Block.prototype.renameVar = function(oldName, newName) {
     for (var j = 0, field; field = input.fieldRow[j]; j++) {
       if ((field instanceof Blockly.FieldVariable ||
           field instanceof Blockly.FieldVariableGetter) &&
-          Blockly.Names.equals(oldName, field.getValue())) {
+          oldName == field.getValue()) {
         field.setValue(newName);
       }
     }
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -197,8 +197,7 @@ Blockly.FieldVariable.dropdownCreate = function() {
       variableModelList = variableModelList.concat(variables);
     }
     for (var i = 0; i < variableModelList.length; i++){
-      if (createSelectedVariable &&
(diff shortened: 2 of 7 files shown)


rachel-fenichel:
Is this blocking something today?  Every line of code that it touches is changed by https://github.com/google/blockly/pull/1495

paulkaplan:
@rachel-fenichel nope, definitely can wait until after.

paulkaplan:
Welp, this is going to be complicated by variables and broadcast messages using different comparators. Will resubmit after that is figured.

PR #1294, opened 2017-12-07
Reduce connecting snap radius

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/854

### Proposed Changes

Reduce the distance threshold for blocks to snap together or disconnect. 

### Reason for Changes

Previously, the threshold was higher, so you had to drag a block down off the bottom of a stack for a distance almost twice the height of the block in order to disconnect it (otherwise it would re-attach when you release). Now, you have to drag less far to disconnect a block from a stack. This parameter can't be too low, or else the blocks will jump around while you drag to insert in the middle of a stack (as noted by @rachel-fenichel), so @carljbowman and I played around with it until we found an acceptable compromise.

### DIFF
--- a/core/constants.js
+++ b/core/constants.js
@@ -48,7 +48,7 @@ Blockly.SNAP_RADIUS = 48;
  * Maximum misalignment between connections for them to snap together,
  * when a connection is already highlighted.
  */
-Blockly.CONNECTING_SNAP_RADIUS = 96;
+Blockly.CONNECTING_SNAP_RADIUS = 68;
 
 /**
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):

tmickel:
pretty nice!!! actually makes the blocks "feel" a bit faster (guess 'cause the insertion marker disappears faster when you want it to). weird how little tweaks like this can affect perceived performance.

note also this was set really high for horizontal blocks (since they are giant). you might want different values for horizontal/vertical (but feels pretty good right now)

PR #1297, opened 2017-12-12
Micromanage events for new procedure blocks because they are shadows

### LINKED ISSUE #1291: Adding procedure argument emits create event with block instead of shadow
To repro:
1. Enable logging in the custom procedure playground for the main workspace (run `definitionWorkspace.addChangeListener(function(e){console.log(e);});`)
2. Right-click -> edit the existing procedure definition in the definition workspace.
3. Click "Add text/number input" button.
4. Click "ok"
5. Check out the events in the console. There should be two "Create" events. The first is the argument on the call block, the second is the new argument on the prototype block. Both of these blocks are shadows, but the events list them as `<block` instead of `<shadow`. The call block event is most important: it leads to https://github.com/LLK/scratch-gui/issues/1005

Because scratch-vm listens to these events to keep in sync, not knowing that these blocks (particularly the first one, in the caller) are actually shadows is a problem. 

@rachel-fenichel Is this a bug with the way events are emitted? Or is this the expected behavior?

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1291 using the suggestion from @rachel-fenichel 

### Proposed Changes

Wrap all uses of `workspace.newBlock` in a "try disable events => (create block) =>  finally enable events => emit new block" sequence. This allows us to emit correct events for procedure operations. 

### Reason for Changes

As rachel mentioned, 
> The create event that's fired in new block is really optimistic. Anywhere that we go through domToBlock events are disabled until all of the blocks are created, to sidestep problems like this. We can do the same here:

I made a few minor tweaks to your recommended code @rachel-fenichel, adding try/finally (you mentioned that as a necessity when disabling/enabling events in the past) and also implemented the fix for all three places in the code we were using newBlock then setting it to a shadow. 

### Test Coverage

Went through the repro in the linked issues, the events are emitted with the xml listing the new blocks as shadows. I also built into the GUI and went through the repro steps from the original bug report (https://github.com/LLK/scratch-gui/issues/1005) and confirmed that it fixes that issue as well

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -304,16 +304,24 @@ Blockly.ScratchBlocks.ProcedureUtils.attachShadow_ = function(input,
     argumentType) {
   if (argumentType == 'n' || argumentType == 's') {
     var blockType = argumentType == 'n' ? 'math_number' : 'text';
-    var newBlock = this.workspace.newBlock(blockType);
-    if (argumentType == 'n') {
-      newBlock.setFieldValue('1', 'NUM');
-    } else {
-      newBlock.setFieldValue('', 'TEXT');
+    Blockly.Events.disable();
+    try {
+      var newBlock = this.workspace.newBlock(blockType);
+      if (argumentType == 'n') {
+        newBlock.setFieldValue('1', 'NUM');
+      } else {
+        newBlock.setFieldValue('', 'TEXT');
+      }
+      newBlock.setShadow(true);
+      if (!this.isInsertionMarker()) {
+        newBlock.initSvg();
+        newBlock.render(false);
+      }
+    } finally {
+      Blockly.Events.enable();
     }
-    newBlock.setShadow(true);
-    if (!this.isInsertionMarker()) {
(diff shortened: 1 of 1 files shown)


blocks_vertical/procedures.js
```diff
@@ -484,16 +500,24 @@ Blockly.ScratchBlocks.ProcedureUtils.checkOldTypeMatches_ = function(oldBlock,
  */
 Blockly.ScratchBlocks.ProcedureUtils.createArgumentEditor_ = function(
     argumentType, displayName) {
-  if (argumentType == 'n' || argumentType == 's') {
-    var newBlock = this.workspace.newBlock('argument_editor_string_number');
-  } else {
-    var newBlock = this.workspace.newBlock('argument_editor_boolean');
+  Blockly.Events.disable();
```
rachel-fenichel, line 503: There's nothing wrong with adding this code, but I think events should always be off for the procedure declaration editor, so this should not do anything.

rachel-fenichel:
LG

PR #1302, opened 2017-12-13
Broadcast Inputs and Implicit Message Deletion

### LINKED ISSUE #1268: Broadcast blocks should have input fields
### Expected Behavior

'broadcast' and 'broadcast and wait' blocks should have input fields in addition to the drop downs (allowing plugging in blocks as input). 

E.g. 

image

### Actual Behavior

The field on these blocks is currently only a field variable dropdown and not a pluggable input.

### LINKED ISSUE #1270: Broadcast message block drop down triggers error when used from block in flyout
### Expected Behavior

Should be able to create a new broadcast message variable using field dropdown of any broadcast block in flyout.

### Actual Behavior

Trying to create a new message from drop down of broadcast block in flyout causes an unchecked type error:

image

### Steps to Reproduce

Load a new Scratch 3.0 project.
From flyout, without dragging block into the workspace, click on the drop down field for the 'broadcast' block.
Click 'New message...'
In the modal, create a new variable and click ok (or press enter).
Modal does not close and a type error is logged in the console.

### LINKED ISSUE #1290: Broadcast message deletion behaviour
In Scratch 2.0 broadcast messages are never explicitly deleted.  Removing the last reference to a message on any workspace in the project removes the message from dropdown menus.

In Scratch 3.0
- Will messages have explicit deletion?  _From an implementation perspective this is much better, since it can match variable and list deletion behaviour._
- Can messages be renamed?
- Can the default message ('message1') ever be deleted?
- Are there backwards compatibility concerns?

References:
- Variable deletion discussion: https://github.com/LLK/scratch-blocks/issues/975
- Variable deletion discussion: https://github.com/LLK/scratch-blocks/issues/372#issuecomment-227532505

### LINKED ISSUE #1304: Importing empty-string as message name from SB2
Opening this up for discussion so that others can weigh in and I don't go down any rabbit holes unnecessarily...

In Scratch 2, users can create an empty string message by plugging an input into a broadcast block and taking it back out. The empty string message option becomes available in block dropdowns and can then be used as a valid message. Note: it is not possible to explicitly create an empty string message (trying to do so results in a noop).

What should the behavior be for importing empty string messages into scratch 3? Below are some possibilities:

- Should empty string message names be translated into some unique (user-friendly) identifier such as 'messageN' where N is some number such that 'messageN' is a fresh name?
- Should we attempt to allow the empty string message name (only for sb2 imports)?

I have tried the second option above, but it poses some block rendering problems where the size of the option to be selected (the option we are hovering over) is very tiny and the check mark appears to be in between options (see below):

image

image

kchadha:
### Resolves

Resolves #1268 
Resolves #1270 
Resolves #1290 
Resolves #1304 

### Proposed Changes

Broadcast blocks (namely `broadcast` and `broadcast and wait`) now have pluggable inputs.
New messages can once again be created from the flyout (resolving the issue in #1270). 
New messages created from a block in the flyout are created as variables in the flyout's target workspace, thus appearing in all broadcast block dropdown menus.

A message is created when either: 
1) the broadcast block dropdown menu is clicked for the first time OR
2) any broadcast block with the message selected in its dropdown field is dragged out of the flyout and created on the workspace **(this is not new functionality)** OR
3) the message is explicitly created using the `New message` option in any of the broadcast block dropdowns.

A message is deleted using the implicit deletion behavior from SB2. Most of this functionality is implemented in scratch-vm. A message is deleted (disappears from dropdowns) if all blocks referencing that message are removed from the VM. This deletion behavior updates every time the workspace is updated (e.g. switching sprites or switching between the blocks tab and the costumes/sounds tabs). When the workspace is updated, the broadcast blocks in the flyout update their fields according to SB2 behavior -- the broadcast message that comes first in sorted order will appear selected in the field of each of the broadcast blocks in the flyout. The sorting behavior is the same as that used to sort the options in the dropdown menu (alphanumeric sorting by message name).

The assertion that shadows cannot have variables is no longer true, so the assertion statement has been removed.

#### Diversions from SB2
SB2 projects that have empty string message names will be converted to a fresh name `message#`.
In SB2, plugging in an input, e.g. `((1) + (1))`, into a `broadcast [foo]` or `broadcast [foo] and wait` block, and then taking that input back out will result in removing the old message name `foo` and replacing it with the empty string. In the proposed changes, `foo` will remain in the message dropdowns, as well as in the shadow behind the plugged in input. 

The reason for these divergences are to avoid the potential for buggy behavior when introducing this round-about way of creating empty string message names when they are otherwise disallowed from explicit creation.

If we decide that we would like to maintain this SB2 behavior, that change can be made.

### Reason for Changes

To maintain compatibility with Scratch 2.0 which allows blocks to be plugged into the broadcast block inputs, and has the implicit deletion behavior.

### Test Coverage

Existing tests pass.

### Additional Notes

This PR depends on corresponding PRs in scratch-vm and scratch-gui:
[LLK/scratch-gui#1033](https://github.com/LLK/scratch-gui/pull/1033)
[LLK/scratch-vm#860](https://github.com/LLK/scratch-vm/pull/860)

These three PRs can be tested together [here](https://llk.github.io/scratch-gui/broadcast-input/).

### DIFF
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -51,6 +51,8 @@ Blockly.FieldVariable = function(varname, opt_validator, opt_variableTypes) {
   this.setValue(varname || '');
   this.addArgType('variable');
   this.variableTypes = opt_variableTypes;
+  this.variableType = (opt_variableTypes && opt_variableTypes.length == 1) ?
+    opt_variableTypes[0] : '';
 };
 goog.inherits(Blockly.FieldVariable, Blockly.FieldDropdown);
 
@@ -84,14 +86,14 @@ Blockly.FieldVariable.prototype.initModel = function() {
     // Check if there was exactly one element specified in the
     // variableTypes list. This is the list that specifies which types of
     // variables to include in the dropdown.
-    // If there is exactly one element specified, make the variable
-    // being created this specified type. Else, default behavior is to create
-    // a scalar variable
-    if (this.getVariableTypes_().length == 1) {
-      this.sourceBlock_.workspace.createVariable(this.getValue(),
-        this.getVariableTypes_()[0]);
-    } else {
-      this.sourceBlock_.workspace.createVariable(this.getValue());
+    this.sourceBlock_.workspace.createVariable(this.getValue(), this.variableType);
+  } else {
+    // Using shorter name for this constant
+    var broadcastMsgType = Blockly.BROADCAST_MESSAGE_VARIABLE_TYPE;
+    var broadcastVars = this.sourceBlock_.workspace.getVariablesOfType(broadcastMsgType);
+    if (this.variableType == broadcastMsgType && broadcastVars.length != 0) {
+      broadcastVars.sort(Blockly.VariableModel.compareByName);
+      this.setValue(broadcastVars[0].name);
     }
   }
 };
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -205,9 +205,13 @@ Blockly.FieldVariable.dropdownCreate = function() {
     }
   }
   // Ensure that the currently selected variable is an option.
-  // TODO (#1270): Remove isBroadcastType check here when flyout variables fixed.
-  if (createSelectedVariable && workspace && !isBroadcastType) {
-    var newVar = workspace.createVariable(name);
+  if (createSelectedVariable && workspace) {
+    var newVar = null;
+    if (isBroadcastType && workspace.isFlyout) {
+      newVar = workspace.targetWorkspace.createVariable(name, Blockly.BROADCAST_MESSAGE_VARIABLE_TYPE);
+    } else {
+      newVar = workspace.createVariable(name);
+    }
     variableModelList.push(newVar);
   }
   variableModelList.sort(Blockly.VariableModel.compareByName);
@@ -217,10 +221,7 @@ Blockly.FieldVariable.dropdownCreate = function() {
     options[i] = [variableModelList[i].name, variableModelList[i].getId()];
   }
   if (isBroadcastType) {
-    // TODO (#1270): Re-enable create broadcast message dropdown in flyout when fixed.
-    if (!workspace.isFlyout) {
-      options.push([Blockly.Msg.NEW_BROADCAST_MESSAGE, Blockly.NEW_BROADCAST_MESSAGE_ID]);
-    }
+    options.push([Blockly.Msg.NEW_BROADCAST_MESSAGE, Blockly.NEW_BROADCAST_MESSAGE_ID]);
   } else {
     options.push([Blockly.Msg.RENAME_VARIABLE, Blockly.RENAME_VARIABLE_ID]);
     if (Blockly.Msg.DELETE_VARIABLE) {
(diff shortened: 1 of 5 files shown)


rachel-fenichel:
No variables in shadows: read [this thread](https://groups.google.com/forum/#!searchin/blockly/shadow$20variable%7Csort:date/blockly/4g-jvqrydLg/lcpIcC5bBgAJ).

kchadha:
@rachel-fenichel, thanks for the pointer to the thread. That makes a lot of sense. I guess this should also factor into the discussion for #1290. If we explicitly disallow deletion/renaming for broadcast messages (not saying this is what will necessarily happen), would this then be a non-issue? e.g. Is deletion/renaming the only consideration we should make when deciding to have a field_variable in a shadow?

rachel-fenichel:
I think deletion/renaming is the only concern, but it includes implicit deletion/renaming.

Variable fields may now be better at creating variables at the right time, so maybe it's okay.  I would experiment with some of the workflows from that thread to see if anything ends up feeling too weird.

kchadha:
Right now I have no deletion of broadcast messages implemented at all (explicit or implicit), so testing that out will probably have to wait until we make a decision about what deletion behavior should be.

kchadha:
@paulkaplan Does it make sense to change this and corresponding PR's mentioned above to be against develop instead of these new separate branches ('broadcast-messages' branch)?

My original intuition for the separate branches was so that they would be a staging area for all of the new broadcast block changes (including merging in Rachel's changes from upstream blockly). I think this might not make sense anymore, since merging in Rachel's work will be a pretty substantial change/refactor anyways, and it's not planned for our current milestone.

core/field_variable.js
```diff
@@ -84,14 +86,14 @@ Blockly.FieldVariable.prototype.initModel = function() {
...
-        this.getVariableTypes_()[0]);
-    } else {
-      this.sourceBlock_.workspace.createVariable(this.getValue());
+    this.sourceBlock_.workspace.createVariable(this.getValue(), this.variableType);
+  } else {
+    // Using shorter name for this constant
+    var broadcastMsgType = Blockly.BROADCAST_MESSAGE_VARIABLE_TYPE;
+    var broadcastVars = this.sourceBlock_.workspace.getVariablesOfType(broadcastMsgType);
+    if (this.variableType == broadcastMsgType && broadcastVars.length != 0) {
+      broadcastVars.sort(Blockly.VariableModel.compareByName);
```
paulkaplan, line 95: Can you check that this isn't destroying anything (sort is in-place)

paulkaplan (approved):

core/field_variable.js
kchadha: Confirmed, it's not destroying anything. The result of `this.sourceBlock_.workspace.getVariablesOfType(broadcastMsgType)` does not change, but broadcastVars (which is specifically being used for the sort) does.

image

paulkaplan (approved):
LG

rachel-fenichel:
Did you check for problems related to variables in shadows?

kchadha:
@rachel-fenichel I believe we don't have any issues with variables in shadows right now. Variables in shadows aren't implicitly deleted unless all blocks (including obscured shadows) with a particular variable are deleted. 

This is not the same behavior as in scratch 2 where a block covering a shadow deletes the underlying variable in the shadow field. 

We decided to avoid this in Scratch 3 (unless an issue with it comes up) because in Scratch 2, this is how the empty string names are created, and we wanted to avoid those in Scratch 3, if possible.

PR #1305, opened 2017-12-17
Add data-lists to Blockly.Categories

mrjacobbloom:
### Resolves

There's not a standalone issue for this but the discussion is in https://github.com/LLK/scratch-blocks/pull/1267#issuecomment-347996948

### Proposed Changes

Now that list blocks are a different color from scalar variable blocks, add an entry for them in `Blockly.Categories` and update the vertical blocks accordingly. This makes it so that the svg `<g>` tag that represents the block will have the attribute `data-category="data-lists"`

Note that I named the property `Blockly.Categories.dataLists` -- if you guys have anything against camelcase I have no problem changing that.

### Reason for Changes

To make scratch-blocks fully styleable with CSS! Before this change, all blocks in the data category could be styled but there was no way to focus in on list blocks.

### Test Coverage

Right now there are no tests for all the stylability data-attributes, but maybe that's something to consider! I tried my best to get the test runner working but I'm still having some difficulties -- anyway, `vertical_playground.html` seems to work fine and all blocks in the data category are showing the expected `data-category`

### DIFF
--- a/core/constants.js
+++ b/core/constants.js
@@ -268,6 +268,7 @@ Blockly.Categories = {
   "sound": "sounds",
   "pen": "pen",
   "data": "data",
+  "dataLists": "data-lists",
   "event": "events",
   "control": "control",
   "sensing": "sensing",
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -164,7 +164,7 @@ Blockly.Blocks['data_listcontents'] = {
           "variableTypes": ["list"]
         }
       ],
-      "category": Blockly.Categories.data,
+      "category": Blockly.Categories.dataLists,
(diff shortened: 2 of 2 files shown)


kchadha:
LGTM.

kchadha (approved):

core/constants.js
```diff
@@ -268,6 +268,7 @@ Blockly.Categories = {
   "sound": "sounds",
   "pen": "pen",
   "data": "data",
+  "dataLists": "data-lists",
```
kchadha, line 271: Something I didn't catch before I merged this PR, is the hyphenated constant a problem? E.g. should "data-lists" be "data_lists"? @paulkaplan

PR #1307, opened 2017-12-18
Empty/whitespace message names no longer cause type error.

### LINKED ISSUE #1298: Creating new message with <space> as name causes type error.
### Expected Behavior

Should be able to create new message with name <space>

### Actual Behavior

Attempting to do this causes error:
`Uncaught TypeError: Cannot read property 'toLowerCase' of null`

### Steps to Reproduce

Drag any of the broadcast blocks out of the flyout.
Use dropdown menu option to create a new message.
Attempt to name the new message the space character.

kchadha:
### Resolves

Resolves #1298.

### Proposed Changes

Fixes the bug that was causing a type error when trying to create a new message without a name or with whitespace as a name.

This change only resolves the bug, but does not yet allow creating a message with whitespace as a a name (as can be done in SB2). This functionality will require fixing the block dropdown menu rendering issues discussed in #1304.

### Reason for Changes

Bug #1298 

### Test Coverage

Existing tests pass.

### DIFF
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -263,7 +263,9 @@ Blockly.FieldVariable.prototype.onItemSelected = function(menu, menuItem) {
     } else if (id == Blockly.NEW_BROADCAST_MESSAGE_ID) {
       var thisField = this;
       var setName = function(newName) {
-        thisField.setValue(newName);
+        if (newName) {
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):

paulkaplan:
LG

PR #1308, opened 2017-12-19
Allow Whitespace Messages

kchadha:
This is another example of a change where there is a separate validator for the names of broadcast messages vs. other variables...

### Resolves

Completes #1298, by allowing whitespace message names.

### Proposed Changes

This pull request allows users to create new broadcast messages with arbitrary white space in the name (either solely whitespace as the message name, or arbitrary whitespace interspersed between other text). Broadcast messages no longer follow the same whitespace stripping rules as other variables.

Importing projects with whitespace messages also no longer have dropdown rendering issues as discussed in #1304.

### Reason for Changes

Compatibility with Scratch 2.0.

### Test Coverage

Existing tests pass.

### DIFF
--- a/core/variables.js
+++ b/core/variables.js
@@ -307,7 +307,7 @@ Blockly.Variables.promptName = function(promptText, defaultText, callback, opt_t
  * @return {string} The validated and possibly transformed name of the variable.
  */
 Blockly.Variables.validateName_ = function(name, opt_type) {
-  if (!opt_type || !opt_type == Blockly.BROADCAST_MESSAGE_VARIABLE_TYPE) {
+  if (!opt_type || opt_type != Blockly.BROADCAST_MESSAGE_VARIABLE_TYPE) {
     name = name.replace(/[\s\xa0]+/g, ' ').replace(/^ | $/g, '');
     if (name == Blockly.Msg.RENAME_VARIABLE ||
         name == Blockly.Msg.NEW_VARIABLE) {
--- a/core/variables.js
+++ b/core/variables.js
@@ -185,7 +185,7 @@ Blockly.Variables.createVariable = function(workspace, opt_callback, opt_type) {
   }
   // This function needs to be named so it can be called recursively.
   var promptAndCheckWithAlert = function(defaultName) {
-    Blockly.Variables.promptName(newMsg, defaultName, opt_type,
+    Blockly.Variables.promptName(newMsg, defaultName,
       function(text) {
         if (text) {
           // TODO (#1245) use separate namespaces for lists, variables, and
@@ -223,7 +223,7 @@ Blockly.Variables.createVariable = function(workspace, opt_callback, opt_type) {
             opt_callback(null);
           }
         }
-      });
+      }, opt_type);
   };
   promptAndCheckWithAlert('');
 };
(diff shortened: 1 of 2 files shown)


core/variables.js
```diff
@@ -285,25 +285,33 @@ Blockly.Variables.renameVariable = function(workspace, variable,
  * Prompt the user for a new variable name.
  * @param {string} promptText The string of the prompt.
  * @param {string} defaultText The default value to show in the prompt's field.
+ * @param {string} opt_type Optional type of variable, like 'string' or 'list'.
  * @param {function(?string)} callback A callback. It will be passed the new
  *     variable name, or null if the user picked something illegal.
  */
-Blockly.Variables.promptName = function(promptText, defaultText, callback) {
+Blockly.Variables.promptName = function(promptText, defaultText, opt_type, callback) {
```
rachel-fenichel: If the new parameter is optional it has to be the last parameter--otherwise if it's left out it gets replaced by the next param, which is (almost) never what you want.

core/variables.js
```diff
@@ -285,25 +285,33 @@ Blockly.Variables.renameVariable = function(workspace, variable,
...
-        // Ok, not ALL names are legal...
-        newVar = null;
-      }
+      newVar = Blockly.Variables.validateName_(newVar, opt_type);
     }
     callback(newVar);
   });
 };
 
+Blockly.Variables.validateName_ = function(name, opt_type) {
```
rachel-fenichel: Needs JSDoc.

core/variables.js
```diff
@@ -285,25 +285,33 @@ Blockly.Variables.renameVariable = function(workspace, variable,
...
+  if (!opt_type || !opt_type == Blockly.BROADCAST_MESSAGE_VARIABLE_TYPE) {
+    name = name.replace(/[\s\xa0]+/g, ' ').replace(/^ | $/g, '');
+    if (name == Blockly.Msg.RENAME_VARIABLE ||
```
rachel-fenichel: This means that a broadcast message can be named "Rename variable" or "new variable", even though no other variable type allows that.  Is that intentional?

core/variables.js
kchadha: done.

core/variables.js
kchadha: done.

core/variables.js
kchadha: This is intentional. These names were allowed in 2.0, and setting these as broadcast message names doesn't cause conflicts. 
I'm also not entirely convinced that we need to disallow these cases from other types of variables, but I didn't want to mess with variables in this PR.

core/variables.js
```diff
@@ -287,23 +287,37 @@ Blockly.Variables.renameVariable = function(workspace, variable,
...
 };
 
+/**
+ * Validate the variable name provided by the user.
+ * @param {string} name The user-provided name of the variable.
+ * @param {string} opt_type Optional type of variable, like 'string' or 'list'.
+ * @return {string} The validated and possibly transformed name of the variable.
+ */
+Blockly.Variables.validateName_ = function(name, opt_type) {
+  if (!opt_type || !opt_type == Blockly.BROADCAST_MESSAGE_VARIABLE_TYPE) {
```
paulkaplan: hmmmm can you use `x != y` instead of `!x == y`. I'm surprised by stickiness every time.

paulkaplan (changes requested):
Tiny operator change requested

core/variables.js
kchadha: done

core/variables.js
rachel-fenichel: sg

paulkaplan (approved):

core/variables.js
mrjacobbloom: Not stickiness, the code was doing what you you would think and it just happened to work out:
screen shot 2017-12-21 at 3 34 59 pm

PR #1310, opened 2017-12-19
Bail on dropdown editor if no options instead of exploding

paulkaplan:
Previously, trying to open a menu with no options does this https://github.com/LLK/scratch-gui/issues/1047

Now it just doesn't open anything. Not the best behavior, but it doesn't explode 💁‍♂️

This is not the long-term intended behavior anyway, since a variable will be available at some point. So this is just a backstop to prevent errors.


kchadha (approved):
LGTM.

rachel-fenichel (approved):

Issue #1299, opened 2017-12-13
Remove "Go to definition" option until implemented


ericrosenbaum:
### Expected Behavior

Selecting the "go to definition" item in the context menu for a custom block should scroll the workspace to the definition for that block.

For now, we should probably remove this item from the context menu.

### Actual Behavior

This feature is not implemented.

PR #1311, opened 2017-12-20
Remove 'Go To Definition' option from context menu

thisandagain:
### Resolves

GH-1299

### Proposed Changes

- Removes "Go to Definition" item from context menu

### Reason for Changes

- Removes option for feature that is not yet implemented


paulkaplan (approved):
LGTM

PR #1313, opened 2017-12-20
Add warp getters/setters for custom procedures

### LINKED ISSUE #1277: Implement procedure declaration method for toggling "warp"
In order to finish the implementation of the custom procedure editor, we need methods along the lines of the `addBooleanExternal` etc. methods to toggle the "warp" (aka "run without screen refresh").

paulkaplan:
I know that warp is not actually used as a boolean from within
scratch-blocks, but parsing/stringifying just allows us to guarantee it
is a boolean even after being modified by domToMutation.

### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1277

See PR linked below to see it used in the GUI.

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -44,7 +44,7 @@ Blockly.ScratchBlocks.ProcedureUtils.callerMutationToDom = function() {
   var container = document.createElement('mutation');
   container.setAttribute('proccode', this.procCode_);
   container.setAttribute('argumentids', JSON.stringify(this.argumentIds_));
-  container.setAttribute('warp', this.warp_);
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
How does parsing and stringifying it add guarantees about the type?

paulkaplan:
Previously, if you read the `warp_` property initially, it would be a boolean. Later, after updating with `domToMutation`, it would become a string (because it was read from `getAttribute`. So this just makes it consistent.

rachel-fenichel:
Gotcha.  This still doesn't save us from someone arbitrarily setting the attribute to a value other than `true` or `false` (what I thought you were saying it would do) but I don't think that's a big concern.

LGTM

PR #1320, opened 2017-12-21
Fix custom block argument remove icon

### LINKED ISSUE #1289: Why is the custom procedure editor's remove-input/label button a "no" sign?
I'm just curious - I would expect to see an X mark.

"No" sign

My opinion: X marks are associated with closing things, but also with crossing things out. "No" signs (like the one used in the screenshot above) are more related to halting, preventing, disabling; but not really removing. Maybe scissors are the most intuitive? (I suppose you could argue that scissors are related to cutting things *out*, to be used later (think: cut, paste), but I'm thinking of cutting the input *off* (think: remove), which is also the way Scratch 2.0 uses scissors (i.e., the Delete tool).)

towerofnix:
### Resolves

Resolves #1289.

### Proposed Changes

Changes the custom block's "remove argument" icon, per @carljbowman's asset given in #1289.

### Reasons for Changes

To be updated with the official design change, and to make the icon less confusing:

> My opinion: X marks are associated with closing things, but also with crossing things out. "No" signs (like the one used in the screenshot above) are more related to halting, preventing, disabling; but not really removing. Maybe scissors are the most intuitive? (I suppose you could argue that scissors are related to cutting things out, to be used later (think: cut, paste), but I'm thinking of cutting the input off (think: remove), which is also the way Scratch 2.0 uses scissors (i.e., the Delete tool).)

### Test Coverage

Not applicable.

### DIFF
--- a/media/icons/remove.svg
+++ b/media/icons/remove.svg
@@ -1,13 +1,19 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg width="20px" height="20px" viewBox="0 0 20 20" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-    <!-- Generator: Sketch 47.1 (45422) - http://www.bohemiancoding.com/sketch -->
-    <title>delete-argument</title>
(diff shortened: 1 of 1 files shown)


thisandagain:
Awesome. Thanks @towerofnix

paulkaplan (approved):
LG, thanks!

Issue #1019, opened 2017-07-21
Hidden workspace still responds to keyboard events


paulkaplan:
### Expected Behavior

_Please describe what should happen_

A hidden workspace should not respond to keyboard undo/redo. 

### Actual Behavior

_Describe what actually happens_

After a call to `workspace.setVisible(false)`, you can still undo/redo using the keyboard. 

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

1. Open the [vertical playground](http://llk.github.io/scratch-blocks/playgrounds/tests/vertical_playground.html)
2. Add some blocks, move some things around.
3. Press "Hide"
4. Undo using the keyboard (<kbd>cmd + z</kbd>)
5. Press "Show"
6. The last operation has been undone while the workspace was hidden.

paulkaplan:
@rachel-fenichel  is there any legit use case for keeping that behavior? If not, should this be filed as a bug upstream on blockly?

paulkaplan:
_update: submitted the following as https://github.com/LLK/scratch-blocks/pull/1020_

First impression fix would be to add a third short-circuit condition to `Blockly.onKeyDown_`, changing 
```javascript
Blockly.onKeyDown_ = function(e) {
  if (Blockly.mainWorkspace.options.readOnly || Blockly.utils.isTargetInput(e)) { ...
```
to 
```javascript
Blockly.onKeyDown_ = function(e) {
  if (Blockly.mainWorkspace.options.readOnly || Blockly.utils.isTargetInput(e) ||
    !Blockly.mainWorkspace.rendered) { ...
```

But it doesn't seem that `workspace.setVisible` sets the `rendered` property to false. Is that by design? Is there a different visibility check I could do? Or should I added `this.rendered = isVisible` to the end of the `setVisible` function?

rachel-fenichel:
I need to check whether that's a valid use of `rendered`.  The comment on it says 
```
/**
 * The render status of an SVG workspace.
 * Returns `true` for visible workspaces and `false` for non-visible,
 * or headless, workspaces.
 * @type {boolean}
 */
Blockly.WorkspaceSvg.prototype.rendered = true;
```
but I'm not convinced that hidden rendered workspaces are the same as non-rendered workspaces at this point.  I'd rather have a `visible` attribute and have `rendered` only cover the difference between a `workspace` and a `workspaceSvg`.

fsih:
[...]

PR #1322, opened 2017-12-25
Change wording in create new block dialog

sjhuang26:
### Resolves

[GUI issue #1019](https://github.com/LLK/scratch-gui/issues/1019) 

### Proposed Changes

In the "Create new block" dialog, when a "number or text" input is added to the block, have its default content be "number or text" instead of "string or number."

### Reason for Changes

The terminology "string" may be confusing for beginners.

### Test Coverage

None.


paulkaplan (approved):
LG thanks!

rachel-fenichel:
By the way, these should actually be messages (Blockly.Msg.SOMETHING) instead of just strings in the code.  I'll make an issue for that if we don't have one.

PR #1327, opened 2017-12-28
Bugfix: LLK/scratch-gui/issues/919

apple502j:
### Resolves
<https://github.com/LLK/scratch-gui/issues/919>
### Proposed Changes
Deleted  "options" array and changed line 102 from "field_dropdown" to "input_value" in scratch-blocks/blocks_vertical/event.js ,"event_whenbackdropswitchesto"
### Reason for Changes
This code makes a bug.

### DIFF
--- a/blocks_vertical/event.js
+++ b/blocks_vertical/event.js
@@ -99,11 +99,8 @@ Blockly.Blocks['event_whenbackdropswitchesto'] = {
       "message0": "when backdrop switches to %1",
       "args0": [
         {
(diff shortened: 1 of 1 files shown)


AmazingMech2418:
I don't think this will fix the problem. The code just deletes the menu options. It should just add to the array.

apple502j:
Looks block, switch backdrop  to () uses this code.

apple502j:
Oh, I found my code is a bit wrong. Close and make a new pull request.

PR #1330, opened 2018-01-01
Bugfix: LLK/scratch-gui/issues/919 (CHANGED)

apple502j:
### Resolves
LLK/scratch-gui#919
### Proposed Changes
Deleted "options" array and changed line 102 from "field_dropdown" to "input_value" in scratch-blocks/blocks_vertical/event.js ,"event_whenbackdropswitchesto".

Add```

       '<value name="BACKDROP">'+
         '<shadow type="looks_backdrops"></shadow>'+
       '</value>'+

```

to blocks_vertical/default_toolbox.js
### Reason for Changes
Bug request.
Old request was wrong because it won't change default_toolbox.js

### DIFF
--- a/blocks_vertical/default_toolbox.js
+++ b/blocks_vertical/default_toolbox.js
@@ -304,6 +304,9 @@ Blockly.Blocks.defaultToolbox = '<xml id="toolbox-categories" style="display: no
     '</block>'+
     '<block type="event_whenthisspriteclicked" id="event_whenthisspriteclicked"></block>'+
     '<block type="event_whenbackdropswitchesto" id="event_whenbackdropswitchesto">'+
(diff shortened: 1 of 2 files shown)


paulkaplan (changes requested):
There are a few problems with this approach.
1. The `looks_backdrop` menu block has the looks color, so it ends up with a yellow block with a purple input: image
2. In scratch 2 that block does not have an input (as in, you cannot drop something into it). This change allows something to be dropped in, which we do not want. It should remain a field instead of an input.

chrisgarrity:
@apple502j I'm closing this PR as we'll soon be implementing all the blocks as core-extensions for localization, and the dynamic menu will get handled at that point.

Issue #1318, opened 2017-12-20
Modify costume and backdrop reporters to be more consistent / flexible


thisandagain:
Our current costume and backdrop reporters have a sort of asymmetry that can be frustrating (even for beginners). We currently only allow for reporting a costume *number* and a backdrop *name* rather than allow for getting either value for both. Proposed change:

### Current
```
(costume #)
(backdrop name)
```

### Proposed
```
(costume [number / name])
(backdrop [number / name])
```

paulkaplan:
One issue: we still haven't dealt with the "menus in monitors" and this would make that problem even worse.

thisandagain:
@paulkaplan Thanks for pointing that out. Let's do this and then I'll move https://github.com/LLK/scratch-gui/issues/785 into January for resolution. Does that work for you?

thisandagain:
Minor change to this after looking at the data. Usage statistics from shared projects in the past 30 days:

| Block Name | Opcode | Projects Shared |
| ----------- | ---------- | -------------------- |
| `costume #` | `costumeIndex` | 54,253 |
| `backdrop name` | `sceneName` | 9,167 |

Because of this I recommend we flip the drop-down defaults to be *number* first and *name* second.

thisandagain:
(Original comment updated)

PR #1332, opened 2018-01-04
Modify costume and backdrop reporters to be more consistent / flexible

paulkaplan:
### Resolves

Blocks part of https://github.com/LLK/scratch-blocks/issues/1318. Will be closable after VM and GUI parts are also merged, those depend on this.

### Proposed Changes

Make the reporters for costume and backdrop number / name consistent by using a menu to switch between number and name

```
(costume [number / name])
(backdrop [number / name])
```

Here is what the new blocks look like (ordering and appearance based on active sprite/backdrop is handled by the GUI, this is just what the blocks look like). 

image

### DIFF
--- a/blocks_vertical/default_toolbox.js
+++ b/blocks_vertical/default_toolbox.js
@@ -201,9 +201,8 @@ Blockly.Blocks.defaultToolbox = '<xml id="toolbox-categories" style="display: no
         '</shadow>'+
       '</value>'+
     '</block>'+
-    '<block type="looks_costumeorder" id="looks_costumeorder"></block>'+
-    '<block type="looks_backdroporder" id="looks_backdroporder"></block>'+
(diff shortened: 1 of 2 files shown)


ericrosenbaum:
I noticed an inconsistency in our code: it looks like Blockly generally [suggests](https://developers.google.com/blockly/guides/create-custom-blocks/dropdown-menus) using all caps for the internal arg value for menu items.  We do this in some places, but in other places we use `_underscores_` and in yet others we just use plain strings. It's a developer ergonomics thing, not something to fix right now, but ideally we would be consistent.

ericrosenbaum (approved):
LGTM

paulkaplan:
@ericrosenbaum yeah I agree, I noticed that on another PR. We can change them all at the same time.

Issue #1199, opened 2017-11-01
Audit currently implemented blocks for Droppable / Non-droppable inputs


carljbowman:
We need to update some of the 3.0 blocks to not allow reporters to be dropped into their inputs, e.g. "touching [mouse pointer]"

I will follow up with a complete audit.

towerofnix:
IMO there should definitely be some discussion on this. I don't want to get too specific before actually seeing the audit, but - in general there's two important things I think need to be focused on:

* It shouldn't be *discouraging* to get stalled by putting a block in an input..
* *Or* by being stalled by *not* being able to put a block in an input.

You know, low floor, high ceiling :)

(Again, IMO -) I think the best way to see what's discouraging is to, well, do playtests. *Let people* try sticking reporters in weird places. See if being able to is stalling, or discouraging. Try another playtest in which these are much more rigid; one where most (if not all) dropdown inputs *aren't* droppable.

What makes *sense* to limit at first may well not be the best choice. I imagine it's most important to see what's confusing *in practice*. For example, when teaching kids how to use variables, would they be stalled by dropping a variable into the first input of "set (..) to"? Certainly it seems like they *could* be, but it still feels like it warrants a try.

It's worth taking a look at LLK/scratch-vm#709, btw; two relevant things:

> We had a long conversation about this today after a play test. We believe it is important for some blocks to have non-droppable menus to avoid confusion and error states. (from @thisandagain)

> Can this be a feature that advanced users are allowed to disable? It's a pain to manually edit JSON (from @MegaApuTurkUltra)

What I'm most hoping for (I hope I'm not being rude?) is that, alongside each recommended change in the audit, an explanation is given (e.g. "touching (..) should be non-droppable because..."), and that these are also *practiced* - I don't know the procedure for playtests, but I think it would make sense to have a time or two where people are given a sheet of the new droppable and non-droppable inputs, for them to explore and discuss?

carljbowman:
@towerofnix - I appreciate response. I agree, we do not want to discourage people from trying new things and tinkering, but we also want to set people up for a good experience. 

There are some inputs that if they accept blocks it's not conceptually clear to the creator what will happen. In some cases that can be desired, but for people are new and who are learning how blocks work that can result in unexpected behavior or doing nothing. That confusion can create false expectations about how the majority of blocks inputs work or worse discourage the creator as they think they have made an error.

We have been trying everything that we are developing in workshops with people on a regular basis and we have seen this confusion.

[...]

PR #1339, opened 2018-01-10
Make some blocks non-droppable by removing menu shadows

ericrosenbaum:
### Resolves

Part of https://github.com/LLK/scratch-blocks/issues/1199

This requires a GUI PR as well.

### Proposed Changes

Remove the menu shadows for some blocks to make the input non-droppable (i.e. so that you cannot drag and drop a reporter block onto the input).

### Reason for Changes

See discussion in the issue above. Our focus is preventing confusion for beginners.


paulkaplan (approved):
LG

PR #1353, opened 2018-01-19
Update the color sliders when eyedropper changes the value.

### LINKED ISSUE #823: Eyedropper does not update sliders

paulkaplan:
The hsv slider values are intentionally kept separate from the field
value so they are more stable when changed. They are initialized when
the editor is opened and only updated by the sliders. This was causing
the sliders not to update when a new color was chosen via eyedropper. To
get around this, manually update the internal hsv values when coming
back from the eyedropper callback so that the sliders get updated (which
happens in setValue).

Current behavior
slider-update--develop

Fixed 
slider-update--fixed

Fixes https://github.com/LLK/scratch-gui/issues/823

### DIFF
--- a/core/field_colour_slider.js
+++ b/core/field_colour_slider.js
@@ -256,6 +256,11 @@ Blockly.FieldColourSlider.prototype.sliderCallbackFactory_ = function(channel) {
 Blockly.FieldColourSlider.prototype.activateEyedropperInternal_ = function() {
   var thisField = this;
   Blockly.FieldColourSlider.activateEyedropper_(function(value) {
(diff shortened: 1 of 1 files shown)


kchadha (approved):
LGTM.

PR #1360, opened 2018-01-22
Handling variable naming/renaming -- case sensitivity, and used name checking

### LINKED ISSUE #1008: Cannot open project with both "X" and "x" variables

### LINKED ISSUE #1183: Incorrect "change" event when renaming variables
In the vertical playground, create a variable and drag the new "set <x> to <y>" block into the workspace. Open the dropdown and rename the variable to something else. 

There are two events, a `VarRename` and a `Change` for the variable field. The `VarRename` seems ok, it has the varId, oldName and newName. The `Change` event, though, doesn't look it has the right fields: the `oldValue` is the variable ID and the `newValue` is the new variable name. It looks like a confusion between the field value and the field name w/r/t variables, which is something that was changed to separate variable Ids and names. 

I believe this is the root cause of https://github.com/LLK/scratch-vm/issues/723

The change event is a problem because it is identical to the change event you get when changing to a different variable in the dropdown. In that case, `oldValue` and `newValue` are both variable IDs, so everything works.

### LINKED ISSUE #1292: "Used name" checking for variables should be case-insensitive

### LINKED ISSUE #1343: Disallow renaming variables/lists to existing variable/list (resp.) names
### Expected Behavior

Should follow Scratch 2.0 behavior of disallowing variables to be renamed to existing variable names. Extend this behavior to lists as well.

### Actual Behavior

Currently, variables can be renamed to an existing variable name, conflating the two variables into one, and replacing all uses of the renamed variable to the existing name.

### Steps to Reproduce

Create a variable 'bar'.
Use it on the workspace.
Create a variable 'foo'.
Use it on the workspace.
Rename 'foo' to 'bar'.
Observe that all instances of 'foo' on the workspace now refer to 'bar', and 'foo' has been removed from the tool palette.

kchadha:
### Resolves

Resolves #1292 
Resolves [scratch-gui#1008](https://github.com/LLK/scratch-gui/issues/1008)
Resolves #1343 
Resolves #1183

### Proposed Changes

Allows variables, lists, and broadcast messages to have case sensitive names. In the case of variables and lists, X and x are different names. In the case of broadcast messages, as in Scratch 2.0, users can create new messages with different capitalizations of the same message name  (e.g. 'message', 'mEssaGe', and 'Message' can all co-exist in the same broadcast message dropdown menu, but will all trigger the same 'when I receive' hats. This means that even though the dropdown has the three different capitalizations listed above, all of these name options refer to the same underlying message being broadcasted. This is consistent with the behavior of Scratch 2.0.

This PR also disallows renaming variables to the name of an existing variable. The previous behavior was to replace the variable being renamed with a reference to the existing variable that shares a name and type with the new name.

### Reason for Changes

The issues listed above, and backwards compatibility with scratch 2.0.

#1183 is closed by a combination of this PR and the variables_by_id work. This is explained in [this comment](https://github.com/LLK/scratch-blocks/issues/1183#issuecomment-359558583).

### Test Coverage

Tests have been modified to reflect new variable case sensitivity and the new rules about disallowing a variable to be renamed to an existing variable name.

#### Additional Notes

This work should be updated when the upstream [blockly issue#1537](https://github.com/google/blockly/issues/1537) is resolved. Validators for creating and renaming variables will move into their own separate functions to handle type-specific variable name validators.

### DIFF
--- a/core/variable_map.js
+++ b/core/variable_map.js
@@ -77,11 +77,11 @@ Blockly.VariableMap.prototype.renameVariable = function(variable, newName) {
     if (!conflictVar) {
       this.renameVariableAndUses_(variable, newName, blocks);
     } else {
-      // We don't want to rename if the variable if one with the exact new name
+      // We don't want to rename the variable if one with the exact new name
       // already exists.
       console.warn('Unexpected conflict when attempting to rename ' +
         'variable with name: ' + variable.name + ' and id: ' + variable.getId() +
-        ' to new name: ' + newName + '. Variable with the new name already exist' +
+        ' to new name: ' + newName + '. A variable with the new name already exists' +
         ' and has id: ' + conflictVar.getId());
 
     }
--- a/core/variable_map.js
+++ b/core/variable_map.js
@@ -74,10 +74,16 @@ Blockly.VariableMap.prototype.renameVariable = function(variable, newName) {
   Blockly.Events.setGroup(true);
   try {
     // The IDs may match if the rename is a simple case change (name1 -> Name1).
-    if (!conflictVar || conflictVar.getId() == variable.getId()) {
+    if (!conflictVar) {
       this.renameVariableAndUses_(variable, newName, blocks);
     } else {
-      this.renameVariableWithConflict_(variable, newName, conflictVar, blocks);
+      // We don't want to rename if the variable if one with the exact new name
+      // already exists.
+      console.warn('Unexpected conflict when attempting to rename ' +
+        'variable with name: ' + variable.name + ' and id: ' + variable.getId() +
+        ' to new name: ' + newName + '. Variable with the new name already exist' +
+        ' and has id: ' + conflictVar.getId());
+
     }
   } finally {
     Blockly.Events.setGroup(false);
@@ -299,7 +305,7 @@ Blockly.VariableMap.prototype.getVariable = function(name, opt_type) {
   var list = this.variableMap_[type];
   if (list) {
     for (var j = 0, variable; variable = list[j]; j++) {
-      if (Blockly.Names.equals(variable.name, name)) {
+      if (variable.name == name) {
         return variable;
       }
     }
--- a/core/variable_map.js
+++ b/core/variable_map.js
@@ -73,7 +73,6 @@ Blockly.VariableMap.prototype.renameVariable = function(variable, newName) {
   var blocks = this.workspace.getAllBlocks();
   Blockly.Events.setGroup(true);
   try {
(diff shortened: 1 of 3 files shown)


core/variable_map.js
```diff
@@ -74,10 +74,16 @@ Blockly.VariableMap.prototype.renameVariable = function(variable, newName) {
   Blockly.Events.setGroup(true);
   try {
     // The IDs may match if the rename is a simple case change (name1 -> Name1).
```
rachel-fenichel: Delete this comment, since it's no longer relevant.

tests/jsunit/workspace_test.js
```diff
@@ -235,8 +235,8 @@ function test_renameVariable_ReferenceExists() {
 }
 
 function test_renameVariable_TwoVariablesSameType() {
-  // Expect 'renameVariable' to change oldName variable name to newName.
-  // Expect oldName block name to change to newName
+  // Cannot rename variable to a name that already exists
```
rachel-fenichel, line 238: For these tests, add a note to each one to say that it's different from the behaviour in blockly.  It'll help us in the future.

rachel-fenichel (commented):
Two changes then LGTM.

core/variable_map.js
kchadha: Done.

tests/jsunit/workspace_test.js
kchadha: Done.

rachel-fenichel (approved):

paulkaplan (approved):
LGTM

PR #1361, opened 2018-01-25
Var Type Name Validators

### LINKED ISSUE #1351: Creating duplicate messages
#1350 introduces a UX issue where users would see a discrepancy between the first time and additional times they try to create a new message `message1` using any of the broadcast blocks in the flyout.

E.g.
1. Open a new project
2. Use one of the broadcast blocks in the flyout to create a new message `message1`
3. Repeat step 2.

Performing step 2 looks like a no-op to the user, while performing step 3 results in the following pop-up:

image

The discrepancy here is confusing to the user. What's happening behind the scenes is that performing step 2 the first time transforms the potential variable (a concept introduced by #1350) into a real variable of the same name (this same thing would happen if any of the broadcast blocks are dragged from the flyout into the workspace). Repeating step 2 yields an error because now, a variable with the same name and type already exists.

----------------------------------------------------------------------------------------------------

### Proposed Solution:
Change validator for broadcast messages to not complain when a user tries to create a new message with the same name as an existing broadcast message. In Scratch 2.0, users can create broadcast messages with the same name as many times as they want. The result of performing this action appears to be a no-op to the user.

kchadha:
### Resolves

Resolves #1351 
And some residual todos from #1244

### Proposed Changes

The proposed changes are a major refactor of the code that handles creation and renaming of variables. The changes include:

- The validation of variable names according to type is more explicit now. There are two functions, one that handles name validation for broadcast messages and one that handles name validation for both scalar variables and lists.

- The user can now 'create' (results in a no-op) new broadcast messages that have the same name as one that already exists as a broadcast message. This issue is discussed in #1351, and this new behavior is consistent with Scratch 2.0.

- The prompt for creating or renaming variables no longer recurses when the supplied name already exists. This behavior was already not present in the scratch-gui due to a bug, but this new behavior is consistent with Scratch 2.0.

- The validators for scalar variables and lists no longer merge runs of whitespace in the middle of a variable name (e.g. 'my        variable' and 'my  variable' can co-exist as two separate scalar variable names). White space at the beginning or end of a scalar variable or list name is still trimmed. This new behavior is consistent with Scratch 2.0.

### Reason for Changes

The issues listed above, code cleanup, and consistency with Scratch 2.0 behavior.

### Test Coverage

Existing tests pass.

### DIFF
--- a/core/variables.js
+++ b/core/variables.js
@@ -269,19 +269,21 @@ Blockly.Variables.createVariable = function(workspace, opt_callback, opt_type) {
   // Decide on a modal message based on the opt_type. If opt_type was not
   // provided, default to the original message for scalar variables.
   var newMsg, modalTitle;
-  if (opt_type === Blockly.BROADCAST_MESSAGE_VARIABLE_TYPE) {
+  if (opt_type == Blockly.BROADCAST_MESSAGE_VARIABLE_TYPE) {
     newMsg = Blockly.Msg.NEW_BROADCAST_MESSAGE_TITLE;
     modalTitle = Blockly.Msg.BROADCAST_MODAL_TITLE;
-  } else if (opt_type === Blockly.LIST_VARIABLE_TYPE) {
+  } else if (opt_type == Blockly.LIST_VARIABLE_TYPE) {
     newMsg = Blockly.Msg.NEW_LIST_TITLE;
     modalTitle = Blockly.Msg.LIST_MODAL_TITLE;
-  } else if (opt_type === Blockly.SCALAR_VARIABLE_TYPE){
+  } else {
+    // Note: this case covers 1) scalar variables, 2) any new type of
+    // variable not explicitly checked for above, and 3) a null or undefined
+    // opt_type -- turns a falsey opt_type into ''
+    // TODO (#1251) Warn developers that they didn't provide an opt_type/provided
+    // a falsey opt_type
+    opt_type = opt_type ? opt_type : '';
     newMsg = Blockly.Msg.NEW_VARIABLE_TITLE;
     modalTitle = Blockly.Msg.VARIABLE_MODAL_TITLE;
-  } else {
-    console.warn('Encountered unexpected type "' + opt_type + '" in call to ' +
-        'Blockly.Variables.createVariable');
-    return;
   }
   var validate = Blockly.Variables.nameValidator_.bind(null, opt_type);
 
@@ -359,12 +361,9 @@ Blockly.Variables.nameValidator_ = function(type, text, workspace, opt_callback)
   } else if (type == Blockly.LIST_VARIABLE_TYPE) {
     return Blockly.Variables.validateScalarVarOrListName_(text, workspace, type,
         Blockly.Msg.LIST_ALREADY_EXISTS);
-  } else if (type == Blockly.SCALAR_VARIABLE_TYPE) {
+  } else {
     return Blockly.Variables.validateScalarVarOrListName_(text, workspace, type,
         Blockly.Msg.VARIABLE_ALREADY_EXISTS);
-  } else {
-    console.warn('Encountered unexpected type "' + type + '" in call to ' +
-      'Blockly.Variables.nameValidator_');
   }
 };
 
@@ -417,12 +416,6 @@ Blockly.Variables.validateBroadcastMessageName_ = function(name, workspace, opt_
  */
 Blockly.Variables.validateScalarVarOrListName_ = function(name, workspace,
     type, errorMsg) {
-  if (type != Blockly.SCALAR_VARIABLE_TYPE &&
-      type != Blockly.LIST_VARIABLE_TYPE) {
-    console.warn ('Encountered unexpected type "' + type + '" when trying to validate' +
-        'name as a scalar variable or a list.');
-    return null;
-  }
   // For scalar variables, we don't want leading or trailing white space
   name = Blockly.Variables.trimName_(name);
   if (!name) {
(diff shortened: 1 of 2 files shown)


core/variables.js
```diff
@@ -259,72 +259,182 @@ Blockly.Variables.realizePotentialVar = function(varName, varType, potentialVarW
...
+  } else {
+    console.warn('Encountered unexpected type "' + opt_type + '" in call to ' +
+        'Blockly.Variables.createVariable');
+    return;
   }
+  var validate = Blockly.Variables.nameValidator_.bind(null, opt_type);
   // This function needs to be named so it can be called recursively.
-  var promptAndCheckWithAlert = function(defaultName) {
-    Blockly.Variables.promptName(newMsg, defaultName,
+  //var promptAndCheckWithAlert = function(defaultName) {
```
rachel-fenichel: Remove commented-out code.

core/variables.js
```diff
@@ -259,72 +259,182 @@ Blockly.Variables.realizePotentialVar = function(varName, varType, potentialVarW
...
-      }, opt_type, modalTitle);
-  };
-  promptAndCheckWithAlert('');
+      }, modalTitle);
+};
+
+/**
+ * This function provides a common interface for variable name validation agnostic
+ * of type. This is so that functions like Blockly.Variables.createVariable and
+ * Blockly.Variables.renameVariable can call a single function (with a single type
```
rachel-fenichel: "type" is duplicated: on the end of this line and the start of the next line.

core/variables.js
```diff
@@ -259,72 +259,182 @@ Blockly.Variables.realizePotentialVar = function(varName, varType, potentialVarW
...
+ *     is determined to be invalid/in-use, and the calling function should not
+ *     proceed with creating or renaming the variable.
+ * @private
+ */
+Blockly.Variables.nameValidator_ = function(type, text, workspace, opt_callback) {
+  // The validators for the different variable types require slightly different arguments.
+  // For broadcast messages, if a broadcast message of the provided name already exists,
+  // the validator needs to call a function that updates the selected
+  // field option of the dropdown menu of the block that was used to create the new message.
+  // For scalar variables and lists, the validator needs a retry function that is responsible
```
rachel-fenichel: I don't see a retry function.  Is this comment out of date?

core/variables.js
```diff
@@ -337,77 +447,55 @@ Blockly.Variables.createVariable = function(workspace, opt_callback, opt_type) {
...
+  } else if (varType == Blockly.SCALAR_VARIABLE_TYPE) {
+    promptMsg = Blockly.Msg.RENAME_VARIABLE_TITLE;
+    modalTitle = Blockly.Msg.RENAME_VARIABLE_MODAL_TITLE;
+  } else {
+    // Only lists and variables can be renamed... this is an error
+    console.warn('Encountered unexpected variable type ' + varType +
+        'when attempting to rename a variable.');
+    return;
+  }
+  var validate = Blockly.Variables.nameValidator_(varType);
```
rachel-fenichel: Line 286 is 
`  var validate = Blockly.Variables.nameValidator_.bind(null, opt_type);`

Why is this one handled differently?

rachel-fenichel (commented):
Structurally this looks fine.

You're making it very hard to create variables of any types other than the 3 in scratch.  This is a problem as a library, though people will be able to easily see where they need to add code to resolve it.

core/variables.js
kchadha: done.

core/variables.js
kchadha: done.

core/variables.js
kchadha: fixed.

core/variables.js
kchadha: Oops! that was unintentional. Fixed.

kchadha:
I agree that the proposed changes make it harder to create variables of any other type, and that this is not necessarily a good thing.

One of the things I was debating about was whether or not to keep the functionality of defaulting to the scalar variable type if type information isn't provided or isn't one of the three we already know about (instead of having a warning), but I think the warning will help make it clear to a developer that we expect some sort of name validation, and that it can be type specific or it can follow along with the scalar variable/list naming conventions.

I think in general, this refactor makes it clearer what the rules of each type of variable are, and gives examples of the different things you can do with the name validators. If there are other things I can do to make this more library/developer friendly, within the scope of these changes, definitely let me know.

paulkaplan:
@rachel-fenichel @kchadha and I discussed this a little bit. @kchadha correct me if any of this doesn't seem right:

I think it makes sense to have a "default" validation behavior (probably the case-insensitive scalar one) to enable people to experiment with adding new variable types when using scratch-blocks as a library without having to make their own fork. I still don't have a great vision of how people might use custom variable types without having to fork to make other edits to support it, but it does seem like it if it is just an issue of validation behavior, having a default, particularly since you can't provide your own, is a good idea.

One caveat is we should make sure we are falling back to the _behavior_ of scalar variable types, never actually _using_ the scalar variable type when we don't know what else to use. Doing that would cause very hard to debug issues later down the line. That used to be an issue in the field_dropdown before all the recent work you all did, and it doesn't look like anything like that was introduced here, but just wanted to call it out.

Does that seem right to you @kchadha and @rachel-fenichel?

TL;DR I think it is ok to use the scalar name validation behavior for new variable types we don't know have. That means
- remove [this warning in create](https://github.com/kchadha/scratch-blocks/blob/7a63cb1237f7df655b8ecc2c6e431b6f9659c579/core/variables.js#L282-L283) and using the scalar variable messages instead
- remove [this warning in rename](https://github.com/kchadha/scratch-blocks/blob/7a63cb1237f7df655b8ecc2c6e431b6f9659c579/core/variables.js#L462-L463) and using the scalar variable messages instead
- remove [this warning in the validator](https://github.com/kchadha/scratch-blocks/blob/7a63cb1237f7df655b8ecc2c6e431b6f9659c579/core/variables.js#L366-L367) and use the scalar variable validator
- remove [this early return in the scalar validator](https://github.com/kchadha/scratch-blocks/blob/7a63cb1237f7df655b8ecc2c6e431b6f9659c579/core/variables.js#L420-L424)

rachel-fenichel:
@paulkaplan @kchadha I like that solution.

kchadha:
@paulkaplan's suggestion sounds good to me.
One remaining question I have is: what should be the behavior if the type is not provided?

rachel-fenichel:
I suggest defaulting to the empty string; once you switch scalar variables to 'scalar' you could add a warning if the type is empty: "did you mean to do that?"

kchadha:
@rachel-fenichel @paulkaplan, I made the requested changes. The only thing I left in is a warning/early return if the developer has called Blockly.renameVariable on a broadcast message, since we explicitly disallow renaming broadcast messages. Let me know if anything needs to change.
Thanks for reviewing!

rachel-fenichel (approved):
LG

paulkaplan (approved):
LG

PR #1365, opened 2018-01-29
Constants for variable/list renaming prompts.

kchadha:
In the PR https://github.com/LLK/scratch-blocks/pull/1361, I accidentally forgot to commit the messages file. Without this, the variable/list specific prompt messages don't work, and the lack of this file introduces a regression where attempting to rename a list causes a type error.

Existing tests pass.


paulkaplan (approved):
LG

PR #1377, opened 2018-02-08
Keep stop block mutation in sync with dropdown value

### LINKED ISSUE #1357: Crash on switching sprites after connecting "stop other scripts in this sprite"
### Expected Behavior

No crash!

### Actual Behavior

If you use a "stop other scripts in this sprite" block, with another block connected after it, and switch away from and back to that sprite, it crashes.

This seems related to https://github.com/LLK/scratch-blocks/issues/1356

### Steps to Reproduce

- Build this stack:

screen shot 2018-01-22 at 1 28 05 pm

- Switch to another sprite or the stage, and then back

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1357
Fixes https://github.com/LLK/scratch-blocks/issues/1357

### Proposed Changes

Fire mutator event change to keep the stop dropdown field value in sync with the mutator.

I had to include the call to `setValue` in the validator in order to group the events so that the value change and the mutator change cannot get out of sync. That feels wrong, is there a better way @rachel-fenichel?

### Reason for Changes

The two bugs listed above were the result of a mutation change not being emitted. Now a mutation change is emitted manually.

This bug is not reproducible in the playground, although you could see that there was no mutation event being emitted when the value was changed. A full test of the bugs reported above requires VM integration.

### DIFF
--- a/blocks_vertical/control.js
+++ b/blocks_vertical/control.js
@@ -191,7 +191,17 @@ Blockly.Blocks['control_stop'] = {
         ['other scripts in sprite', OTHER_SCRIPTS]
       ];
     }, function(option) {
+      // Create an event group to keep field value and mutator in sync
+      // Return null at the end because setValue is called here already.
+      Blockly.Events.setGroup(true);
+      var oldMutation = Blockly.Xml.domToText(this.sourceBlock_.mutationToDom());
       this.sourceBlock_.setNextStatement(option == OTHER_SCRIPTS);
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
@paulkaplan this seems reasonable.  I agree that the way setValue is happening in the validator feels wrong, but can't see the right way right now. If I suddenly figure it out I'll let you know.

LGTM.

Issue #1381, opened 2018-02-11
Behavior of color picker on sensor blocks broken 


ivanixgames:
### Expected Behavior
This issue originally opened in scratch-gui, #766.
Please see detail there of symptoms.
https://github.com/LLK/scratch-gui/issues/766

Users expect behavior to mimic that of color picker found in paint editor.
Upon tapping or clicking in the color range of the slider, the thumb should move to that position

### Actual Behavior
For hue and brightness the thumb moves incrementally while the mouse button is held down.
The saturation does not work at all and can only be adjusted by dragging the thumb.
 
### Analysis
I traced the issue down to field_colour_slider.js in this repo.  will be issuing a PR shortly.

PR #1382, opened 2018-02-11
Fix behavior of color picker on sensor blocks

ivanixgames:
### Resolves
https://github.com/LLK/scratch-blocks/issues/1381

Expectation:
Users expect behavior to mimic that of color picker found in paint editor.
Upon tapping or clicking in the color range of the slider, the thumb should move to that position

Actual Behavior, 2 issues:
1) For hue and brightness the thumb moves incrementally while the mouse button is held down.
2) The saturation does not work at all and can only be adjusted by dragging the thumb.

### Proposed Changes
For issue 1:
a)  All 3 sliders ( Color, Saturation, Brightness ) should call setMoveToPointEnabled(true) .

For issue 2:
a)  orginal code was using range of  0 to 1 using step of 0.001.  However the sliderbase class rounds up steps so can only use steps of 1 or greater, this requires a max range to be scaled up as well, which in turn requires an external conversion (SATURATION_CONV) to be used to convert back to range of 0 to 1 outside the slider.  Note that this issue and solution is not enabled when setMoveToPointEnable(true) is called. 
However this fix should remain in case in future it is decided that incrementally thumb movement is desired again.
 
### Reason for Changes
Provide consistent color pick behavior with that on paint editor and as reported in
original issue on scratch-gui.

### Test Coverage
Test manually using Chromebrowser on test file
file://.../scratch-blocks/tests/vertical_playground.html

1) drag 'touching color' block from sensors category onto workspace.
2) click on color,  color picker dialog pop ups.
3) move mouse pointer to some part of colored range bar for each of Color, Saturation, Brightness
4) Click or press down on left mouse button.

Expectation:
-  thumb to move to position
- value to also be adjusted base on new thumb position
Results 
- passed with minor issue

Minior issue:
Repeated clicking in color range will cause thumb to move up and off of slider!
I traced this down to an issue with google ui sliderbase and its animation logic.
Issue will be submitted to github on that repo shortly later.
The issue is revealed because scratch uses thumb style property of margin-top: -1.
However, I believe this is a bug in google ui sliderbase.js, not in scratch code.

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -830,7 +830,7 @@ Blockly.Css.CONTENT = [
   '.blocklyDropDownDiv .goog-slider-horizontal .goog-slider-thumb {',
     'width: 26px;',
     'height: 26px;',
-    'margin-top: -1px;',
+    'top: -1px;',
     'position: absolute;',
     'background-color: white;',
     'border-radius: 100%;',
--- a/core/field_colour_slider.js
+++ b/core/field_colour_slider.js
@@ -67,6 +67,13 @@ Blockly.FieldColourSlider.activateEyedropper_ = null;
 Blockly.FieldColourSlider.EYEDROPPER_PATH = 'eyedropper.svg';
 
 /**
+ * Factor to convert between slider integer range and saturation decimal range 0 - 1.
+ * google closure sliderbase only allows whole step integers,
+ * so we cannot step between 0 and 1 fractionally.
+ */
+Blockly.FieldColourSlider.prototype.SATURATION_CONV = 100;
+
+/**
  * Install this field on a block.
  * @param {!Blockly.Block} block The block containing this field.
  */
@@ -178,7 +185,7 @@ Blockly.FieldColourSlider.prototype.updateDom_ = function() {
 Blockly.FieldColourSlider.prototype.updateSliderHandles_ = function() {
   if (this.hueSlider_) {
     this.hueSlider_.setValue(this.hue_);
-    this.saturationSlider_.setValue(this.saturation_);
+    this.saturationSlider_.setValue(this.saturation_ * this.SATURATION_CONV);
     this.brightnessSlider_.setValue(this.brightness_);
   }
 };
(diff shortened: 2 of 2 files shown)


ivanixgames:
Issue and PR posted regarding minor issue mentioned above.
https://github.com/google/closure-library/issues/876

thisandagain:
Thanks @ivanixgames! We'll take a look at this soon. Sorry for the slow response.

paulkaplan (changes requested):
Awesome find with the `setMoveToPointEnabled`, I'm glad that is already built in. 

But can you explain why the saturation conversion is needed? Is it needed for the `setMoveToPointEnabled` change?

I tried just making the `setMoveToPointEnabled` change locally without the saturation fix and it seems to work fine?

click-move
 
That is with this change
image

Also for the drifting problem, is there a way in our CSS that we can fix that? I wouldn't want to merge this without that being addressed.

ivanixgames:
@paulkaplan 
The saturation conversion is not needed if setMoveToPointEnabled is used.  
That patch is only if the original intended behavior was desired.

For the drifting issue. if you want to fix in css,  make sure margin-top is not set or set to 0px  for horizontal slider, and margin-left is left unset or set to 0px for vertical slider. 
See https://user-images.githubusercontent.com/858678/36078378-4a4e5a0e-0fb0-11e8-8abb-ae6ef686261d.png

paulkaplan:
@ivanixgames cool! Would you mind removing the saturation thing then, because we do want to use setMoveToPointEnabled?

paulkaplan:
@ivanixgames cool! Would you mind removing the saturation thing then, because we do want to use setMoveToPointEnabled?

paulkaplan:
Also @ivanixgames I found that if you just change `margin-top: -1px` to `top: -1px` (since it is position absolute) that the sliders don't drift but are still positioned correctly. So if you could make those two changes (removing the saturation change and changing margin-top to top in css.js line 837) we can merge this PR. Thanks for tracking down this issue!

ivanixgames:
@paulkaplan   Not mind at all, just  give me about a day to make and retest the changes.  Cheers

paulkaplan (approved):

PR #1384, opened 2018-02-14
Block drag

fsih:
### Resolves

In preparation for fixing https://github.com/LLK/scratch-gui/issues/1234

### Proposed Changes

Send events when blocks are dragged out to the right of the workspace. Reset drags when they end to the right of the workspace.

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -188,7 +188,17 @@ Blockly.BlockDragger.prototype.dragBlock = function(e, currentDragDeltaXY) {
   this.dragIcons_(delta);
 
   this.deleteArea_ = this.workspace_.isDeleteArea(e);
-  this.draggedConnectionManager_.update(delta, this.deleteArea_);
+  var isOutside = this.workspace_.isOutside(e);
+  if (isOutside) {
+    // Let mouse events through to GUI
+    this.draggingBlock_.svgGroup_.setAttribute("pointer-events", "none");
+  } else {
+    this.draggingBlock_.svgGroup_.setAttribute("pointer-events", "auto");
+  }
+  this.draggedConnectionManager_.update(delta, this.deleteArea_, isOutside);
+  if (!this.draggedConnectionManager_.wouldDeleteBlock() && !this.draggedConnectionManager_.closestConnection_) {
+    this.fireDragEvent_(isOutside);
+  }
 
   this.updateCursorDuringBlockDrag_();
 };
@@ -204,6 +214,8 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
   // Make sure internal state is fresh.
   this.dragBlock(e, currentDragDeltaXY);
   this.dragIconData_ = [];
+  this.fireEndDragEvent_();
+  this.draggingBlock_.svgGroup_.setAttribute("pointer-events", "auto");
 
   Blockly.BlockSvg.disconnectUiStop_();
 
@@ -256,6 +268,31 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
       ws.refreshToolboxSelection_();
     });
   }
+};
+
+/**
+ * Fire a move event at the end of a block drag.
+ * @param {?boolean} isOutside True if the drag is going outside the visible area.
+ * @private
+ */
+Blockly.BlockDragger.prototype.fireDragEvent_ = function(isOutside) {
+  var event = new Blockly.Events.BlockDrag(this.draggingBlock_);
+  event.oldCoordinate = this.startXY_;
+  event.isOutside = isOutside;
+  event.recordNew();
+  Blockly.Events.fire(event);
+};
+
+/**
+ * Fire a move event at the end of a block drag.
+ * @param {?boolean} isOutside True if the drag is going outside the visible area.
+ * @private
+ */
+Blockly.BlockDragger.prototype.fireEndDragEvent_ = function() {
+  var event = new Blockly.Events.BlockEndDrag(this.draggingBlock_);
+  event.oldCoordinate = this.startXY_;
+  event.recordNew();
+  Blockly.Events.fire(event);
 };
 
 /**
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -78,6 +78,12 @@ Blockly.BlockDragger = function(block, workspace) {
    */
   this.wouldDeleteBlock_ = false;
 
+  /**
+   * Whether the currently dragged block is to the right of the workspace. Keep
+   * track so that we can fire events only when this changes.
+   * @type {boolean}
+   * @private
+   */
   this.wasOutside_ = false;
 
   /**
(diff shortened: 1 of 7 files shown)


core/block_dragger.js
```diff
@@ -146,6 +154,7 @@ Blockly.BlockDragger.initIconData_ = function(block) {
  * @package
  */
 Blockly.BlockDragger.prototype.startBlockDrag = function(currentDragDeltaXY) {
+  this.wasOutside_ = false;
```
rachel-fenichel: Don't need to set this (set in the constructor).

core/block_dragger.js
```diff
@@ -204,6 +218,9 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
   // Make sure internal state is fresh.
   this.dragBlock(e, currentDragDeltaXY);
   this.dragIconData_ = [];
+  var isOutside = this.workspace_.isOutside(e);
```
rachel-fenichel: Don't need to update this--it's updated in the call to `this.dragBlock` two lines above.

core/block_dragger.js
```diff
@@ -258,6 +283,28 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
   }
 };
 
+/**
+ * Fire an event when the dragged blocks move to the right of the workspace, or back into
+ * the workspace.
+ * @param {?boolean} isOutside True if the drag is going outside the visible area.
```
rachel-fenichel: No question mark on the annotation--it's not nullable.

core/workspace_svg.js
```diff
@@ -1110,6 +1110,17 @@ Blockly.WorkspaceSvg.prototype.isDeleteArea = function(e) {
...
 
+/**
+ * Is the mouse event outside the blocks UI, to the right of the workspace?
+ * @param {!Event} e Mouse move event.
+ * @return {boolean} True if event is outside the blocks UI.
+ */
+Blockly.WorkspaceSvg.prototype.isOutside = function(e) {
+  var mousePoint = Blockly.utils.mouseToSvg(e, this.getParentSvg(),
+      this.getInverseScreenCTM());
+  return this.getParentSvg().width.baseVal.value < mousePoint.x;
```
rachel-fenichel: I'm a bit leery of this implementation, because it assumes that the only "outside" is to the right side.

Note that Blockly assumes that there's nothing useful to the "outside" of the toolbox, based on the configuration of the workspace, so it's not like we're doing the right thing either.

I think this will break in RTL.  If you keep the editor in the same place but inject with RTL, everything you're looking for will be in the delete area.  If the editor and stage swap places, this will just be looking on the wrong side.

core/workspace_svg.js
```diff
@@ -1110,6 +1110,17 @@ Blockly.WorkspaceSvg.prototype.isDeleteArea = function(e) {
...
 };
 
+/**
+ * Is the mouse event outside the blocks UI, to the right of the workspace?
+ * @param {!Event} e Mouse move event.
+ * @return {boolean} True if event is outside the blocks UI.
+ */
+Blockly.WorkspaceSvg.prototype.isOutside = function(e) {
+  var mousePoint = Blockly.utils.mouseToSvg(e, this.getParentSvg(),
+      this.getInverseScreenCTM());
```
rachel-fenichel: Why are you converting to SVG here?  You started with a mouse point, and your concern is about window coordinates, not workspace coordinates.  This may be unnecessarily expensive for what you're actually trying to do.

core/events.js
```diff
@@ -95,6 +95,8 @@ Blockly.Events.BLOCK_CHANGE = Blockly.Events.CHANGE;
  * @const
  */
 Blockly.Events.MOVE = 'move';
+Blockly.Events.DRAG_OUTSIDE = 'dragOutside';
```
rachel-fenichel: Nit: comments for these.

core/events.js
```diff
@@ -866,6 +868,146 @@ Blockly.Events.Move.prototype.run = function(forward) {
   }
 };
 
+/**
```
rachel-fenichel, line 881: FYI I'm eventually going to split this file up: https://github.com/google/blockly/issues/1618

No change needed in this PR though.

rachel-fenichel (commented):
Overall LG, except for some concerns about the meaning and implementation of `isOutside`.

Everything else is a nit.

core/events.js
```diff
@@ -866,6 +868,146 @@ Blockly.Events.Move.prototype.run = function(forward) {
...
+ */
+Blockly.Events.DragOutside.prototype.isNull = function() {
+  return false;
+};
+
+/**
+ * Run a drag outside event.
+ */
+Blockly.Events.DragOutside.prototype.run = function() {
+  console.error('Not implemented');
```
fsih: @rachel-fenichel when is this run function used? Should I implement it?

core/workspace_svg.js
fsih: I copied this from somewhere since it seemed to give me the right number. What is the difference between the SVG coordinates and mouse coordinates?

core/workspace_svg.js
fsih: I think eventually I can have isOutside take an RTL argument and return the right thing for RTL. Is there a way to set to RTL now?

core/events.js
rachel-fenichel: Run is called to undo and redo events.  You don't need to implement it at all, because the version implemented in the base class is a no-op and that's what you want.

core/workspace_svg.js
rachel-fenichel: Use [this.RTL](https://github.com/LLK/scratch-blocks/blob/develop/core/workspace.js#L47)

core/block_dragger.js
fsih: I think it makes sense to leave this here, because it's possible to use block dragger for multiple drags

rachel-fenichel: We never reuse a block dragger.

fsih: I don't think anything is logically preventing it though? If someone were using scratch-blocks as a library, they might think it's reasonable to do

fsih:
I think I have a better implementation for isOutside now and am ready for re-review

core/block_dragger.js
rachel-fenichel: There's a bunch of stuff that's set in the constructor that you would need to reset between drags, so reusing a dragger probably breaks pretty fast.  It certainly only lets you drag the same block repeatedly, unless you reach in to change draggingBlock directly.  On top of that, everything but the constructor is marked package or private.

I'd much rather make it clearer that the dragger is not to be reused, if anything.

core/workspace_svg.js
```diff
@@ -1075,10 +1076,18 @@ Blockly.WorkspaceSvg.prototype.createVariable = function(name, opt_type, opt_id)
...
+ * Update cached areas for this workspace.
+ */
+Blockly.WorkspaceSvg.prototype.recordCachedAreas = function() {
+  this.recordBlocksArea_();
+  this.recordDeleteAreas_();
+};
+
 /**
  * Make a list of all the delete areas for this workspace.
  */
```
rachel-fenichel: @private, here and on `recordBlocksArea_`

rachel-fenichel (approved):
One nit, then LGTM.

PR #1394, opened 2018-03-05
Add for-each block

towerofnix:
### Resolves

Towards LLK/scratch-vm#959. Should be merged before LLK/scratch-vm#963 (do read that PR as well).

### Proposed Changes

Adds a "for each" block to the Control category:

For each (i) in (abc)

Both strings and numbers may go in the "abc" input.

The block is also placed into the block palette, below "repeat until".

### Reason for Changes

Compatibility with Scratch 2.0 projects that include the hacked "for in" block. (See LLK/scratch-vm#959.)

The block is placed into the palette for the sake of testing, but this can easily be reverted if it's preferred to be hidden. I have made an identical change in the GUI's default block palette so I could test the VM code for this, which I can make a PR for, if wanted.

### Test Coverage

No new tests. Manually tested (variable dropdown does behave like a variable dropdown, anything can be typed in the text input).

### DIFF
--- a/blocks_vertical/control.js
+++ b/blocks_vertical/control.js
@@ -296,6 +296,39 @@ Blockly.Blocks['control_repeat_until'] = {
   }
 };
 
+Blockly.Blocks['control_for_each'] = {
+  /**
+   * Block for for-each. This is an obsolete block that is implemented for
+   * compatibility with Scratch 2.0 projects.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "type": "control_for_each",
+      "message0": "for each %1 in %2",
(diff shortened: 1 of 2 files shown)


kchadha (changes requested):
@towerofnix Thanks for the pull request, it looks good! I do think the block should be removed from the palette so we don't create false expectations for anyone currently using the vertical playground.

towerofnix:
@kchadha Sounds good. I've removed the block from the toolbox.

kchadha (approved):

kchadha:
Looks great! I'm approving this PR, and will have a review of the scratch-vm pr sometime today.

PR #1400, opened 2018-03-08
Prevent jumping up a category when switching sprites

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/1221

### Proposed Changes

Prevent a small numeric error when selecting blocks category by scroll position. 

### Reason for Changes

Previously, before switching sprites, the scroll position could be very slight above the recorded scroll position of the current category, so that after switching sprites it would jump up by one category. Here I round the position and use `>=` to make sure we stay on the same category when switching sprites.

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -556,11 +556,11 @@ Blockly.Flyout.prototype.selectCategoryByScrollPosition = function(pos) {
   if (this.scrollTarget) {
     return;
   }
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):

rachel-fenichel:
http://i.imgur.com/k6dFJYI.gif

PR #1404, opened 2018-03-12
Don’t scroll the toolbox to the top on populate

### LINKED ISSUE #1401: Flyout scrolls from the top on switching sprites
### Expected Behavior

Generally, the flyout should not unexpectedly move or scroll.

### Actual Behavior

After https://github.com/LLK/scratch-blocks/pull/1400, the flyout will not change categories unexpectedly when switching sprites. But it still scrolls down from the top to the selected category on switching sprites.

Switching sprites causes the [GUI to call workspace.updateToolbox](https://github.com/LLK/scratch-gui/blob/cbe577fd5722205cf7e307129078a12b431a7261/src/containers/blocks.jsx#L99), which causes the blocks to scroll to the Motion category; then the GUI calls toolbox. setSelectedCategoryByName, which scrolls it down again.  

scrolling-on-switch

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/1401

### Proposed Changes

Add an optional parameter to setSelectedItem for whether or not to scroll to the new category, defaulting to true. Other calls to setSelectedItem work the same, but now the populate function does not cause a scroll.

### Reason for Changes

This prevents the problem where switching sprites scrolled to the top and then back down to the selected category.

There's still scrolling when switching to or from the stage, but that's for a different reason, and will require a different fix.

### DIFF
--- a/core/toolbox.js
+++ b/core/toolbox.js
@@ -183,7 +183,7 @@ Blockly.Toolbox.prototype.createFlyout_ = function() {
 Blockly.Toolbox.prototype.populate_ = function(newTree) {
   this.categoryMenu_.populate(newTree);
   this.showAll_();
-  this.setSelectedItem(this.categoryMenu_.categories_[0]);
+  this.setSelectedItem(this.categoryMenu_.categories_[0], false);
 };
 
 /**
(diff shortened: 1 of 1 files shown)


core/toolbox.js
```diff
@@ -340,8 +340,12 @@ Blockly.Toolbox.prototype.getSelectedItem = function() {
 /**
  * Set the currently selected category.
  * @param {Blockly.Toolbox.Category} item The category to select.
+ * @param {boolean} [shouldScroll=true] Whether or not to scroll to the selected category.
```
rachel-fenichel:
```
@param {boolean=} opt_shouldScroll Whether to scroll to the selected category.  Defaults to true.
```

rachel-fenichel (commented):
LG, except for a change to the annotation.

ericrosenbaum:
Thanks @rachel-fenichel!

Issue #1406, opened 2018-03-12
Flyout scrolls when switching between sprite and stage


ericrosenbaum:
### Expected Behavior

Generally, the flyout should not unexpectedly move or scroll.

### Actual Behavior

Following https://github.com/LLK/scratch-blocks/pull/1404, switching sprites does not cause the flyout to scroll from the top. But when you switch from a sprite to the stage or vice versa, the flyout scrolls to get back to the same category you were in. This is because the stage has no blocks in the motion category, so the absolute positions of the categories are in different places.

PR #1407, opened 2018-03-15
Support for scroll position stability

ericrosenbaum:
### Resolves

Progress on https://github.com/LLK/scratch-blocks/issues/1406, which will also require a [GUI PR](https://github.com/LLK/scratch-gui/pull/1620).

### Proposed Changes

Add some functions to support more stable scroll position behavior when switching sprites. These include:

- get and set scaled position of the flyout
- get name of the current category
- get the scroll offset from the top of the current category
- get position of a category by name

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -595,9 +595,9 @@ Blockly.Flyout.prototype.stepScrollAnimation = function() {
  * @return {number} The current scroll position.
  */
 Blockly.Flyout.prototype.getScrollPos = function() {
-    var pos = this.horizontalLayout_ ?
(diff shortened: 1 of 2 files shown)


mrjacobbloom:
Out of curiosity: will there ever be a situation (microworlds? motion+stage?) where one sprite doesn't have the same categories as another?

rachel-fenichel (approved):

ericrosenbaum:
@mrjacobbloom Good question. The GUI part of this change uses the assumption that the set of categories is always the same across sprites and stage- but there's no dependency on that fact here in   scratch-blocks (this PR just provides helper functions).

Issue #1188, opened 2017-10-27
glowStack() causes executing stack to hide until glow is removed, on iOS/Android


msintov:
### Expected Behavior

The executing script should have a yellow outline.

### Actual Behavior

The executing script disappears.

### Steps to Reproduce

We have observed this behavior on iOS and Android for scripts small and large, but in my reproducible case here, this behavior occurs for a very large script in the iOS simulator.

1. Open the vertical playground in iOS simulator (I've tried iOS 9.3 and iOS 10.0. I believe it also repros on iOS 10.3. Perhaps also iOS 11).
2. Paste this xml in the text box for importing: 
[xml_for_long_script.txt](https://github.com/LLK/scratch-blocks/files/1423307/xml_for_long_script.txt)
3. Import the script by tapping "Import from XML"
4. Tap the green flag block, which is the first block of the long script
5. Tap "Stack glow last clicked block" button

Note: If you remove the last block at the very bottom of the stack and then repeat steps 4 and 5, the script will not disappear, but notice that the glow outside is not yellow but instead is the color of the blocks it surrounds (as you can see if you page down through the stack).

### Operating System and Browser

Xcode iOS simulator 9.3 and 10.0 on macOS 10.11.6

rachel-fenichel:
Is the system under serious memory pressure when this happens?

msintov:
@rachel-fenichel No, I don't believe this is a memory issue. For the reproducible steps listed above, the bug can happen even if the simulator was just booted, and my guess is that those steps don't take a lot of memory. Note that if the svg attribute to create the gaussian blur is not applied, then the script does not disappear.

xmeow:
I can reproduce this bug under chrome58, but works fine on chrom63. Looks like browser depends.

paulkaplan:
Yikes, large stacks cause some seriously strange behaviors
1. Confirmed the "disappear on stack glow" with the very large stack on iOS 11 iPad Air simulator 
2. Confirm the "stack glow changes color" if you scroll down the page. Here is a screenshot. Scrolled halfway down, you can see where it goes from yellow to the color of the block
![screen shot 2017-12-20 at 2 08 57 pm](https://user-images.githubusercontent.com/654102/34224132-cf00278c-e58f-11e7-9729-f1f4c09aa839.png)
3. Also, the dragging shadow gets super strange with large stacks of blocks.
[...]

PR #1413, opened 2018-03-20
Improve visibility of stack glow

### LINKED ISSUE #1378: Improve visibility of stack glow
/cc @carljbowman @ericrosenbaum from a discussion we had in person, wanted to file this so that it doesn't get lost. Just a stub, more to come.

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/1378

### Proposed Changes

Increases the default stack glow size, and also makes it externally settable like the rest of the stack glow params (color, opacity). Depending on background color, a different stack glow size might be better.

I also moved out the "replacement glow size" parameter, which is used in the horizontal grammar, just to be consistent. I did not change its value, however.

### DIFF
--- a/core/colours.js
+++ b/core/colours.js
@@ -96,8 +96,10 @@ Blockly.Colours = {
   "insertionMarkerOpacity": 0.2,
   "dragShadowOpacity": 0.3,
   "stackGlow": "#FFF200",
+  "stackGlowSize": 1.3,
   "stackGlowOpacity": 1,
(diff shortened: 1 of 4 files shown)


ericrosenbaum (approved):
Looks glowy to me!

msintov:
Does this change fix https://github.com/LLK/scratch-blocks/issues/1188 (I assume not)?

PR #1427, opened 2018-04-01
Fix scroll speed on Firefox

### LINKED ISSUE #1190: Script area scrolling is very slow in Firefox
### Expected Behavior

Scrolling with the scroll wheel happens at the same speed as in the block palette.

### Actual Behavior

Scrolling is much slower.

### Steps to Reproduce

Add a block to the scripting area so that the scrolling speed is clear, then use the scroll wheel to scroll.

Firefox 57.0 on Debian (Testing).

towerofnix:
### Resolves

Fixes #1190.

### Proposed Changes

Adds a "line-scroll multiplier" value, which is multiplied when the delta mode is line (0x1) instead of pixel (0x0). This makes scrolling on Firefox bearable (as well as on other browsers which might report such a delta mode).

### Reason for Changes

To fix scrolling on Firefox.

### Test Coverage

Manually tested. Equivalent code tested and merged in LLK/scratch-paint#532.


towerofnix:
Note - this shouldn't be merged yet - I also want to fix scaling speed in this PR (see [this](https://github.com/LLK/scratch-blocks/issues/1190#issuecomment-340214251)). I'll compare this with Chromium to get a feel of what the right speed is.

towerofnix:
Alright, compared with Chromium on my system and adjusted the multiplier. Also made scaling use the multiplier. (This is set for a review :P)

Another thing I noticed - horizontal scrolling needs to be fixed! I can do that in this PR if wanted, or make another separate one later. The fix for that would be the same as [the one I did for scratch-paint](https://github.com/LLK/scratch-paint/pull/263).

core/workspace_svg.js
```diff
@@ -190,6 +190,16 @@ Blockly.WorkspaceSvg.prototype.startScrollY = 0;
...
+/***
+ * Distance to scroll when a mouse wheel event is received and
+ * its delta mode is line (0x1) instead of pixel (0x0). In these
+ * cases, a single "scroll" has a delta of 1, which makes the
+ * workspace scroll very slowly (just one pixel). To compensate,
+ * that delta is multiplied by this value.
+ * @type {number}
+ */
+Blockly.WorkspaceSvg.prototype.lineScrollMultiplier = 15;
+
```
towerofnix: Is there a better place to put this constant, or is putting it here right?

towerofnix:
Seeing as we're very fast approaching the Scratch 3.0 beta release, I'm explicitly bumping this pull request. With the 3.0 beta, many, many users will start using the editor; and a fair chunk of them will be Firefox-users who use a mouse. These users will notice the issue and be rather frustrated, because the mouse wheel only working very slowly is annoying and is extremely apparent - almost all people with a mouse use their mouse wheel to scroll.

I know it's certainly not good practice to bump issues/PRs, but this is in the April milestone and is a simple fix for a rather prominent issue. Equivalent code to what's in this PR has been tested and reviewed by @fsih in the recent PR LLK/scratch-paint#532, so it hopefully won't be too complicated to review this one.

core/workspace_svg.js
rachel-fenichel: Constants.js is a good place for this.

Add @package and @const annotations when you put it there, and give it an all caps name to match conventions for constants.

rachel-fenichel (commented):
One comment on constant placement, but this looks good to me.  @paulkaplan I'll merge it after the constant is moved, unless you object.

towerofnix:
@rachel-fenichel Sounds good, made a commit moving it to constants.js.

PR #1430, opened 2018-04-07
Add "while" block

towerofnix:
### Resolves

Towards LLK/scratch-vm#355 (adds "doWhile"). Should be merged alongside LLK/scratch-vm#1032.

### Proposed Changes

Adds a `control_while` block. (This block isn't added to the block palette because it's a hacked, compatibility-purpose block. Naturally, it's functionally equivalent to "repeat until (not (condition))".)

An empty "while" block

### Reason for Changes

Compatibility with Scratch 2.0 projects. This is a hacked block; in total, [1505 projects use the `while` loop](https://github.com/LLK/scratch-vm/issues/355#issuecomment-379418752).

### Test Coverage

No new tests, but tested manually (see VM PR).

### DIFF
--- a/blocks_vertical/control.js
+++ b/blocks_vertical/control.js
@@ -268,11 +268,52 @@ Blockly.Blocks['control_wait_until'] = {
 
 Blockly.Blocks['control_repeat_until'] = {
   /**
-   * Block for repeat until a condition becomes true.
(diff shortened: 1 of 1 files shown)


kchadha (approved):
Looks good!

PR #1435, opened 2018-04-09
Fix shadow block IDs for procedure call blocks

rachel-fenichel:
### Resolves

https://github.com/LLK/scratch-vm/issues/1011
### Proposed Changes

Add a "generateShadows" attribute to the mutation parsing. Provide it when setting the mutation from the custom procedure editor callbacks. This makes the current code paths happen.

By default, do not generate shadow blocks when creating inputs. XML generated by scratch-blocks (when saving the project) and by scratch-vm (for loading a project) should not set "generateShadows". That means that a shadow block will not auto-generate, and the shadow DOM from the XML will be respected.
### Reason for Changes

See https://github.com/LLK/scratch-vm/issues/1011
### Test Coverage

Tested in the playground.

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -56,6 +56,8 @@ Blockly.ScratchBlocks.ProcedureUtils.callerMutationToDom = function() {
  */
 Blockly.ScratchBlocks.ProcedureUtils.callerDomToMutation = function(xmlElement) {
   this.procCode_ = xmlElement.getAttribute('proccode');
+  this.generateShadows_ =
+      JSON.parse(xmlElement.getAttribute('generateShadows'));
   this.argumentIds_ = JSON.parse(xmlElement.getAttribute('argumentids'));
   this.warp_ = JSON.parse(xmlElement.getAttribute('warp'));
   this.updateDisplay_();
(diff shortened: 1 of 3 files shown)


kchadha (approved):
LG

kchadha (changes requested):
Actually, I don't think this should be merged in as is until we figure out how the VM should cooperate with these changes. Testing with these changes (and the necessary GUI change of adding the generateshadows flag to the mutationToDom call from the custom procedure editor), editing a custom procedure definition with a caller with obscured shadows out on the main workspace generates warnings (repro described and warnings copied below) and incorrect block JSON on the vm, which will produce incorrect serialization on saving a project.

Steps to repro warnings:
1) Make sure you have gui edited with [this line](https://github.com/LLK/scratch-gui/blob/develop/src/containers/custom-procedures.jsx#L78) edited to pass in `true` to the `mutationToDom` call.
2) In your locally built gui (with the above changes) use the `Make a Block` button in the toolbox to create a block with two string or number inputs, we'll call the first `a` and the second `b`.
3) Drag out a call block for your new custom procedure and plug something into input `a`
4) Right click the procedure definition and click edit.
5) In the custom procedure editor, delete input b. Then, add another string/text input.
6) Switch to the stage and then switch back to sprite 1.
7) You should see the following warnings in your console:
```Ignoring non-existent input [some UID] in block procedures_prototype```
```Ignoring non-existent input [same UID as above] in block procedures_call```

rachel-fenichel:
Note that those warnings happen without this change as well.

kchadha:
@rachel-fenichel, should this be closed since it's covered in #1439?

PR #1436, opened 2018-04-10
Replace goog.dom.removeNode with parentNode.removeChild

gnarf:
### Resolves

* Removes a good 10-20% of time spent doing a workspace update when switching sprites.

### Proposed Changes

* replaces `goog.dom.removeNode(node)` with `if (node && node.parentNode) node.parentNode.removeChild(node)` 

### Reason for Changes

image
* this is exactly what the `goog.dom.removeNode` method does, however since it avoids yet-another-function-call-for-every-node-removed we save a ton of time (at least with chrome)

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -934,7 +934,7 @@ Blockly.BlockSvg.prototype.dispose = function(healStack, animate) {
   }
   Blockly.BlockSvg.superClass_.dispose.call(this, healStack);
 
-  goog.dom.removeNode(this.svgGroup_);
+  if (this.svgGroup_ && this.svgGroup_.parentNode) this.svgGroup_.parentNode.removeChild(this.svgGroup_);
   blockWorkspace.resizeContents();
   // Sever JavaScript to DOM connections.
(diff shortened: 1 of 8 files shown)


rachel-fenichel:
I tried to repro this, because I expected that v8 would be doing a much better job of inlining code like this.  If function calls are causing problems on this order of magnitude, we definitely need to be much more careful.  

I'm getting some signs that dispose needs some major work, but I can't match your numbers for the percent of the workspace update that comes from removeNode.  How are you getting that 10-20% number?  

My steps:

- Go to preview.scratch.mit.edu
- Set up a project with two empty sprites
image

For each of the following examples, get a performance report of the switch, then select the "Event (click)" part of the timeline.
image

Switching from one empty sprite to another:
image

Switching from an empty sprite to the stage:
image


Switching from the stage to an empty sprite:
image

gnarf:
I think this came from the slightly less "accurate" `console.profile` recordings I was doing... Honestly, we can probably forget about this one, I think it's a red herring

PR #1438, opened 2018-04-11
use .textContent setter

gnarf:
### Proposed Changes

* setting `node.textContent` will empty the node and create a new text node child

### Reason for Changes

* Slight performance boost - probably not a whole lot, but also more readable IMO


rachel-fenichel (commented):
LG.  We'll want to backport this to Blockly as well.

rschamp (approved):
Thanks @gnarf!

PR #1439, opened 2018-04-11
Fix some problems with shadows in custom procedure calls

rachel-fenichel:
### Resolves

Works on LLK/scratch-vm#1011

Includes https://github.com/LLK/scratch-blocks/pull/1435

### Proposed Changes

Includes all changes from #1435 and also preserves occluded shadow DOMs on procedure call blocks on the workspace.

### Reason for Changes

Bugs with save and load, related to orphaning blocks in the VM.

This doesn't fix all of those bugs.  Notably:
- When you delete an input on a custom procedure, it's actually gone.  The orphaned shadow block needs to be destroyed in the VM.
- Text set on custom blocks in the toolbox is not preserved across toolbox refreshes.

### Test Coverage

Tested in the custom procedures playground:
- create a test block with one text/number input
- create a call of the test block
- set the text in the first (only) input
- drag out a reporter and occlude the text
- edit the custom procedure by adding another input
- hit ok to apply the edit
- remove the reporter from the procedure call
- see that the same text still there

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -244,10 +244,15 @@ Blockly.ScratchBlocks.ProcedureUtils.deleteShadows_ = function(connectionMap) {
   // Get rid of all of the old shadow blocks if they aren't connected.
   if (connectionMap) {
     for (var id in connectionMap) {
-      var block = connectionMap[id];
-      if (block && block.isShadow()) {
-        block.dispose();
-        connectionMap[id] = null;
+      var saveInfo = connectionMap[id];
+      if (saveInfo) {
+        var block = saveInfo['block'];
+        if (block && block.isShadow()) {
+          block.dispose();
+          connectionMap[id] = null;
+          // At this point we know which shadow DOMs are about to be orphaned in
+          // the VM.  What do we do with that information?
+        }
       }
     }
   }
@@ -396,7 +401,7 @@ Blockly.ScratchBlocks.ProcedureUtils.populateArgumentOnCaller_ = function(type,
   }
 
   if (connectionMap && oldBlock) {
-    // Reattach the old block.
+    // Reattach the old block and shadow DOM.
     connectionMap[input.name] = null;
     oldBlock.outputConnection.connect(input.connection);
     if (type != 'b' && this.generateShadows_) {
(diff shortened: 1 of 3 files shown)


blocks_vertical/procedures.js
```diff
@@ -233,10 +244,15 @@ Blockly.ScratchBlocks.ProcedureUtils.deleteShadows_ = function(connectionMap) {
...
-        block.dispose();
-        connectionMap[id] = null;
+      var saveInfo = connectionMap[id];
+      if (saveInfo) {
+        var block = saveInfo['block'];
+        if (block && block.isShadow()) {
+          block.dispose();
+          connectionMap[id] = null;
+          // At this point we know which shadow DOMs are about to be orphaned in
+          // the VM.  What do we do with that information?
```
kchadha, line 254: These get delete events right (thanks to dispose)?

rachel-fenichel: Shadow blocks get delete events.  But an occluded shadow is not a block that's being disposed, and does not get a delete event.

kchadha: Hmmm so my thought in response to this comment is:
- it would be nice to somehow provide info to the VM that the shadow that got deleted was because of a shadow getting obscured (as opposed to the input containing the shadow getting deleted from the block)

That way, the VM can continue to 'ignore' delete events for shadows deleted because they were obscured, and in the case that a shadow was deleted because the input that contains that shadow is deleted from the parent block, we can actually handle that shadow deletion in the vm directly (let me know if this doesn't make any sense)

At any rate, the above is only a nice to have, and if it requires a ton of extra work, it does not need to get done. This logic can feasibly be checked in the VM itself, but I seem to have found a different bug that needs to be fixed before that can happen =/

kchadha (approved):
LGTM and works well!

> This doesn't fix all of those bugs. Notably:
>
> - When you delete an input on a custom procedure, it's actually gone. The orphaned shadow block needs to be destroyed in the VM.

Yep, I tried to implement this yesterday, but discovered another bug in the VM, so will file a separate issue with this.

> - Text set on custom blocks in the toolbox is not preserved across toolbox refreshes.

I think this behavior is okay

blocks_vertical/procedures.js
rachel-fenichel: I'll merge this and we can circle back if needed.  My first thought is that we can fire events when changing the shadow dom, with the text of the shadow dom as the event info, and that should probably be enough.

PR #1441, opened 2018-04-13
Optimize block.setParent during clearing workspace

gnarf:
### Proposed Changes

* Tracks an `isClearing` on workspaces, and when clearing, avoids reattaching the SVG nodes to the workspace

### Reason for Changes
* Saves a lot of useless calls to `appendNode` which are followed up slightly later by `removeNode`

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -272,33 +272,23 @@ Blockly.BlockSvg.prototype.getIcons = function() {
  * @param {Blockly.BlockSvg} newParent New parent block.
  */
 Blockly.BlockSvg.prototype.setParent = function(newParent) {
-  if (newParent === this.parentBlock_) {
+  var oldParent = this.parentBlock_;
+  if (newParent == oldParent) {
     return;
   }
-  var svgRoot = this.getSvgRoot();
-  var oldXY = this.getRelativeToSurfaceXY();
-
-  if (this.parentBlock_ && svgRoot) {
-    // Move this block up the DOM.  Keep track of x/y translations.
-
-    // Avoid moving a block up the DOM if it's currently selected/dragging,
-    // so as to avoid taking things off the drag surface.
-
-    // Also if we have a new parent, we should avoid calling `appendChild` twice with
-    // the same child, so we will just move ourselves later in the newParent block.
-
-    // Also do not move it if the workspace is clearing, we will only `removeChild`
-    // a few cycles later.
-    if (Blockly.selected !== this && !newParent && !this.workspace.isClearing) {
-      this.workspace.getCanvas().appendChild(svgRoot);
-      this.translate(oldXY.x, oldXY.y);
-    }
-  }
-
   Blockly.Field.startCache();
   Blockly.BlockSvg.superClass_.setParent.call(this, newParent);
   Blockly.Field.stopCache();
 
+  var svgRoot = this.getSvgRoot();
+
+  // Bail early if workspace is clearing, or we aren't rendered.
+  // We won't need to reattach ourselves anywhere.
+  if (this.workspace.isClearing || !svgRoot) {
+    return;
+  }
+
+  var oldXY = this.getRelativeToSurfaceXY();
   if (newParent) {
     newParent.getSvgRoot().appendChild(svgRoot);
     var newXY = this.getRelativeToSurfaceXY();
@@ -310,6 +300,17 @@ Blockly.BlockSvg.prototype.setParent = function(newParent) {
         newParent.getColourTertiary());
     }
   }
+  // If we are losing a parent, we want to move our DOM element to the
+  // root of the workspace.
+  else if (oldParent) {
+    // Avoid moving a block up the DOM if it's currently selected/dragging,
+    // so as to avoid taking things off the drag surface.
+    if (Blockly.selected != this) {
+      this.workspace.getCanvas().appendChild(svgRoot);
+      this.translate(oldXY.x, oldXY.y);
+    }
+  }
+
 };
 
 /**
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -272,18 +272,23 @@ Blockly.BlockSvg.prototype.getIcons = function() {
  * @param {Blockly.BlockSvg} newParent New parent block.
  */
 Blockly.BlockSvg.prototype.setParent = function(newParent) {
-  if (newParent == this.parentBlock_) {
+  if (newParent === this.parentBlock_) {
     return;
   }
   var svgRoot = this.getSvgRoot();
+  var oldXY = this.getRelativeToSurfaceXY();
   if (this.parentBlock_ && svgRoot) {
     // Move this block up the DOM.  Keep track of x/y translations.
-    var xy = this.getRelativeToSurfaceXY();
+
     // Avoid moving a block up the DOM if it's currently selected/dragging,
     // so as to avoid taking things off the drag surface.
-    if (Blockly.selected != this) {
+
+    // Also do not move it if the workspace is clearing, no need, we will be deleted
+    // Also, if we have a new parent, we should avoid calling `appendChild` twice with
+    // the same child, so we will just move ourselves later in the newParent block
+    if (Blockly.selected !== this && !newParent && !this.workspace.isClearing) {
       this.workspace.getCanvas().appendChild(svgRoot);
-      this.translate(xy.x, xy.y);
+      this.translate(oldXY.x, oldXY.y);
     }
   }
 
@@ -292,7 +297,6 @@ Blockly.BlockSvg.prototype.setParent = function(newParent) {
   Blockly.Field.stopCache();
 
   if (newParent) {
-    var oldXY = this.getRelativeToSurfaceXY();
     newParent.getSvgRoot().appendChild(svgRoot);
     var newXY = this.getRelativeToSurfaceXY();
     // Move the connections to match the child's new position.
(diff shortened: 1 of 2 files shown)


paulkaplan:
Hmmm something very strange is going on with the travis build, it is trying to pull from llk/scratch-blocks block-setParent instead of your fork. Did this branch get pushed first to LLK and then to your fork?

gnarf:
> Hmmm something very strange is going on with the travis build, it is trying to pull from llk/scratch-blocks block-setParent instead of your fork. Did this branch get pushed first to LLK and then to your fork?

yes - ignore that "push" -- i accidentally did push origin instead of push gnarf -- i deleted that branch afterward, I think we can just ignore the "push" test on travis here and rely on the "pr" test

core/block_svg.js
```diff
@@ -272,18 +272,23 @@ Blockly.BlockSvg.prototype.getIcons = function() {
  * @param {Blockly.BlockSvg} newParent New parent block.
  */
 Blockly.BlockSvg.prototype.setParent = function(newParent) {
-  if (newParent == this.parentBlock_) {
+  if (newParent === this.parentBlock_) {
```
paulkaplan: @rachel-fenichel I think you mentioned at some point preferring double equals in this repo, is that right?

core/block_svg.js
```diff
@@ -272,18 +272,23 @@ Blockly.BlockSvg.prototype.getIcons = function() {
...
-    var xy = this.getRelativeToSurfaceXY();
+
     // Avoid moving a block up the DOM if it's currently selected/dragging,
     // so as to avoid taking things off the drag surface.
-    if (Blockly.selected != this) {
+
+    // Also do not move it if the workspace is clearing, no need, we will be deleted
+    // Also, if we have a new parent, we should avoid calling `appendChild` twice with
+    // the same child, so we will just move ourselves later in the newParent block
+    if (Blockly.selected !== this && !newParent && !this.workspace.isClearing) {
```
paulkaplan: Ditto here with == / ===

paulkaplan (changes requested):
Code looks good other than the == / === style preference in this repo, which @rachel-fenichel can give a bit more context about. 

But I think @rachel-fenichel might be better to give a 👍  or 👎 on the actual idea of blocks using state information about their workspace to make granular changes to their render-ish functions. 

One other thing, could you update the PR with the change you are seeing in timing from this? I'd like to just sanity check what you are seeing to be sure it isn't super setup specific or something.

core/block_svg.js
gnarf: Happy to change, I'm not going to pretend like the difference between `==` and `===` will save us on perf in the long run, especially since both of these are objects / null.  However if you look at the spec language for the `===` and `==` operator it's fairly clear that `null === object` is false much sooner than the `==` case.  I'm not sure why you would ever opt-into type-looseness here.

core/block_svg.js
```diff
@@ -272,18 +272,26 @@ Blockly.BlockSvg.prototype.getIcons = function() {
  * @param {Blockly.BlockSvg} newParent New parent block.
  */
 Blockly.BlockSvg.prototype.setParent = function(newParent) {
-  if (newParent == this.parentBlock_) {
+  if (newParent === this.parentBlock_) {
     return;
   }
   var svgRoot = this.getSvgRoot();
+  var oldXY = this.getRelativeToSurfaceXY();
```
rachel-fenichel: Why pull this out of the if?

core/block_svg.js
```diff
@@ -272,18 +272,26 @@ Blockly.BlockSvg.prototype.getIcons = function() {
...
     // Avoid moving a block up the DOM if it's currently selected/dragging,
     // so as to avoid taking things off the drag surface.
-    if (Blockly.selected != this) {
+
+    // Also if we have a new parent, we should avoid calling `appendChild` twice with
+    // the same child, so we will just move ourselves later in the newParent block.
+
+    // Also do not move it if the workspace is clearing, we will only `removeChild`
+    // a few cycles later.
+    if (Blockly.selected !== this && !newParent && !this.workspace.isClearing) {
```
rachel-fenichel: You can check `this.workspace.isClearing` earlier, which will save the `getRelativeToSurfaceXY` call as well.

core/block_svg.js
rachel-fenichel: Consistency with the rest of the code and not needing to think through the implications of `==` vs `===` all the time.  

But I don't personally have strong opinions on this one.  If you do, I can direct you to someone else with strong opinions and I'll stay on the sidelines.

core/block_svg.js
```diff
@@ -272,18 +272,26 @@ Blockly.BlockSvg.prototype.getIcons = function() {
...
+
   if (this.parentBlock_ && svgRoot) {
     // Move this block up the DOM.  Keep track of x/y translations.
-    var xy = this.getRelativeToSurfaceXY();
+
     // Avoid moving a block up the DOM if it's currently selected/dragging,
     // so as to avoid taking things off the drag surface.
-    if (Blockly.selected != this) {
+
+    // Also if we have a new parent, we should avoid calling `appendChild` twice with
```
rachel-fenichel: What is this comment referring to?

core/workspace.js
```diff
@@ -202,6 +202,7 @@ Blockly.Workspace.prototype.getAllBlocks = function() {
  * Dispose of all blocks in workspace.
  */
 Blockly.Workspace.prototype.clear = function() {
+  this.isClearing = true;
```
rachel-fenichel, line 212: Does this need to be on `Workspace` instead of `WorkspaceSvg`?

If you add a property, add it at the top level with [jsdoc](https://developers.google.com/blockly/guides/modify/web/style-guide#properties) and then set it as needed.  That means it gets into our generated documentation.  This one should be @package.

core/block_svg.js
gnarf: It was the first statement in both branches of the `if`

core/block_svg.js
gnarf: :+1:

core/block_svg.js
gnarf: I check `!newParent` also, but your comments here making me realize how mucked this if/comment has gotten, gonna clean it up.

core/workspace.js
gnarf: I'm not sure which it should be on, but I would assume that other workspaces may want to take advantage of knowing if they are clearing while destroying blocks for similar reasons.

You tell me 😄

core/block_svg.js
gnarf: I'm of the opinion that you should use the rules enforced by your linter/style guide with this particular argument.  I'll switch back to `==` here to be consistent.

Though it is worth noting that eslint doesn't actually have a rule that will ban threequals, if it did, I'd ask that we enable it as it would stop me from making this mistake again in the future :)

core/block_svg.js
rachel-fenichel: The ifs weren't an if/else, though.  It may still be right to pull this out, and to reorder the code in the way that you did, but it won't be doing the same thing as the old code.

core/block_svg.js
```diff
@@ -272,27 +272,24 @@ Blockly.BlockSvg.prototype.getIcons = function() {
...
-  }
-
   Blockly.Field.startCache();
   Blockly.BlockSvg.superClass_.setParent.call(this, newParent);
   Blockly.Field.stopCache();
 
+  var svgRoot = this.getSvgRoot();
+
+  // Bail early if workspace is clearing, or we aren't rendered.
+  // We won't need to reattach ourselves anywhere.
```
rachel-fenichel, line 286: This also won't detach it, since that happened as a result of `attachChild`.  I think that's fine in the case when we're clearing, and if not rendered there was nothing to do anyway.

rachel-fenichel: I meant `appendChild` above.  I've looked at it more and I'm satisfied that not detaching it is fine.

core/workspace.js
rachel-fenichel: Good point.  Leave it on `Workspace`, but add the explicit property + documentation.

core/block_svg.js
```diff
@@ -303,6 +300,17 @@ Blockly.BlockSvg.prototype.setParent = function(newParent) {
         newParent.getColourTertiary());
     }
   }
+  // If we are losing a parent, we want to move our DOM element to the
+  // root of the workspace.
+  else if (oldParent) {
```
rachel-fenichel, line 305: Changing this from two `if`s to one `if/else if` is a good idea--saves us from moving things to the top level and then immediately back, and the reason to do it the other way appears to have faded with tiime.  I'll also apply this to Blockly.

rachel-fenichel (commented):
Add jsdoc and then lgtm.

gnarf:
I think with this latest push of the isClearing prototype docblock and default value, we've solved all the outstanding review comments.  Could @paulkaplan AND @rachel-fenichel take another look and green light if we feel green to merge?

core/workspace.js
```diff
@@ -110,6 +110,12 @@ Blockly.Workspace = function(opt_options) {
  */
 Blockly.Workspace.prototype.rendered = false;
 
+/**
+ * Returns `true` if the workspace is currently in the process of a bulk clear.
+ * @type {boolean}
```
rachel-fenichel, line 115: @package

rachel-fenichel (approved):
@package annotation, then lgtm

PR #1442, opened 2018-04-13
Recycle flyout blocks

gnarf:
### Proposed Changes

* Flyout stores a "recycle bin" of the svg nodes it's about to throw out and reuses them when recreating a flyout

### Reason for Changes

* Avoids a bunch of DOM GC and trash for every sprite swap with swapping toolbars.

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -55,12 +55,12 @@ Blockly.Flyout = function(workspaceOptions) {
   this.workspace_ = new Blockly.WorkspaceSvg(workspaceOptions);
   this.workspace_.isFlyout = true;
 
-  // when we create blocks for this workspace, make the default id
-  // the same as the prototype for easier re-use
+  // When we create blocks for this workspace, instead of using the "optional" id
+  // make the default `id` the same as the `type` for easier re-use.
   var newBlock = this.workspace_.newBlock;
-  this.workspace_.newBlock = function(proto, id) {
-    // this here is workspace
-    return newBlock.call(this, proto, id || proto);
+  this.workspace_.newBlock = function(type, id) {
+    // Use `type` if `id` isn't passed. `this` will be workspace.
+    return newBlock.call(this, proto, id || type);
   };
 
   /**
@@ -482,8 +482,9 @@ Blockly.Flyout.prototype.show = function(xmlList) {
       var tagName = xml.tagName.toUpperCase();
       var default_gap = this.horizontalLayout_ ? this.GAP_X : this.GAP_Y;
       if (tagName == 'BLOCK') {
-        // this is probably a naive assumption, but we are going to assume
-        // that in a flyout, the same ID block is the same
+
+        // We assume that in a flyout, the same block id/type means
+        // the same output SVG nodes.  Check for a recycled element.
         var id = xml.getAttribute('id') || xml.getAttribute('type');
         var recycled = this.recycleBlocks_.findIndex(function(block) {
           return block.id === id;
@@ -491,8 +492,8 @@ Blockly.Flyout.prototype.show = function(xmlList) {
         var curBlock;
 
         if (recycled > -1) {
-          // delete the recycled element out of the recycle bin and
-          // assign it to curBlock
+          // Delete the recycled element out of the recycle bin and
+          // assign it to curBlock.
           curBlock = this.recycleBlocks_.splice(recycled, 1)[0];
         } else {
           var curBlock = Blockly.Xml.domToBlock(xml, this.workspace_);
@@ -544,7 +545,7 @@ Blockly.Flyout.prototype.show = function(xmlList) {
   this.listeners_.push(Blockly.bindEvent_(this.svgBackground_, 'mouseover',
       this, deselectAll));
 
-  // Clean out the old recycle bin
+  // Clean out the old recycle bin.
   var oldBlocks = this.recycleBlocks_;
   this.recycleBlocks_ = [];
   for (var i = 0; i < oldBlocks.length; i++) {
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -55,6 +55,14 @@ Blockly.Flyout = function(workspaceOptions) {
   this.workspace_ = new Blockly.WorkspaceSvg(workspaceOptions);
   this.workspace_.isFlyout = true;
 
+  // when we create blocks for this workspace, make the default id
+  // the same as the prototype for easier re-use
+  var newBlock = this.workspace_.newBlock;
+  this.workspace_.newBlock = function(proto, id) {
+    // this here is workspace
+    return newBlock.call(this, proto, id || proto);
+  };
+
   /**
    * Is RTL vs LTR.
    * @type {boolean}
@@ -126,6 +134,14 @@ Blockly.Flyout = function(workspaceOptions) {
    * @package
    */
   this.scrollTarget = null;
+
+  /**
+   * A recycle bin for blocks.
+   * @type {!Array.<!Blockly.Block>}
+   * @private
+   */
+  this.recycleBlocks_ = [];
+
 };
 
 /**
@@ -466,7 +482,21 @@ Blockly.Flyout.prototype.show = function(xmlList) {
       var tagName = xml.tagName.toUpperCase();
       var default_gap = this.horizontalLayout_ ? this.GAP_X : this.GAP_Y;
       if (tagName == 'BLOCK') {
-        var curBlock = Blockly.Xml.domToBlock(xml, this.workspace_);
+        // this is probably a naive assumption, but we are going to assume
+        // that in a flyout, the same ID block is the same
+        var id = xml.getAttribute('id') || xml.getAttribute('type');
+        var recycled = this.recycleBlocks_.findIndex(function(block) {
+          return block.id === id;
+        });
+        var curBlock;
+
+        if (recycled > -1) {
+          // delete the recycled element out of the recycle bin and
+          // assign it to curBlock
+          curBlock = this.recycleBlocks_.splice(recycled, 1)[0];
+        } else {
+          var curBlock = Blockly.Xml.domToBlock(xml, this.workspace_);
+        }
         if (curBlock.disabled) {
           // Record blocks that were initially disabled.
           // Do not enable these blocks as a result of capacity filtering.
@@ -513,6 +543,13 @@ Blockly.Flyout.prototype.show = function(xmlList) {
 
   this.listeners_.push(Blockly.bindEvent_(this.svgBackground_, 'mouseover',
       this, deselectAll));
+
+  // Clean out the old recycle bin
+  var oldBlocks = this.recycleBlocks_;
+  this.recycleBlocks_ = [];
+  for (var i = 0; i < oldBlocks.length; i++) {
+    oldBlocks[i].dispose(false, false);
+  }
 
   this.workspace_.setResizesEnabled(true);
   this.reflow();
@@ -617,7 +654,7 @@ Blockly.Flyout.prototype.clearOldBlocks_ = function() {
   var oldBlocks = this.workspace_.getTopBlocks(false);
   for (var i = 0, block; block = oldBlocks[i]; i++) {
     if (block.workspace == this.workspace_) {
-      block.dispose(false, false);
+      this.recycleBlock(block);
     }
   }
   // Delete any background buttons from a previous showing.
@@ -806,4 +843,17 @@ Blockly.Flyout.prototype.placeNewBlock_ = function(oldBlock) {
 
   block.moveBy(finalOffsetMainWs.x, finalOffsetMainWs.y);
   return block;
+};
+
+/**
+ * Put a previously created block into the recycle bin, used during large
+ * workspace swaps to limit the number of new dom elements we need to create
+ *
+ * @param {!Blockly.BlockSvg} block The block to recycle.
+ */
+Blockly.WorkspaceSvg.prototype.recycleBlock = function(block) {
+  block.rendered = false;
+  var g = block.getSvgRoot();
+  g.setAttribute('transform', '');
+  this.recycleBlocks_.push(block);
 };
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -484,7 +484,7 @@ Blockly.Flyout.prototype.show = function(xmlList) {
       if (tagName == 'BLOCK') {
 
         // We assume that in a flyout, the same block id/type means
-        // the same output SVG nodes.  Check for a recycled element.
+        // the same output BlockSvg.  Check for a recycled element.
         var id = xml.getAttribute('id') || xml.getAttribute('type');
         var recycled = this.recycleBlocks_.findIndex(function(block) {
           return block.id === id;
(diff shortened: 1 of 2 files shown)


paulkaplan:
Wow! Testing locally I'm seeing sprite switching times go from 1000ms to 300ms on my machine.

gnarf:
After:
```
blocks.jsx:105 updateToolboxXML: 163.908935546875ms
blocks.jsx:105 updateToolboxXML: 161.132080078125ms
blocks.jsx:105 updateToolboxXML: 197.967041015625ms
blocks.jsx:105 updateToolboxXML: 148.363037109375ms
blocks.jsx:105 updateToolboxXML: 141.632080078125ms
```

Before:
```
blocks.jsx:105 updateToolboxXML: 669.8251953125ms
blocks.jsx:105 updateToolboxXML: 625.43994140625ms
blocks.jsx:105 updateToolboxXML: 767.947021484375ms
blocks.jsx:105 updateToolboxXML: 801.824951171875ms
blocks.jsx:105 updateToolboxXML: 707.60791015625ms
blocks.jsx:105 updateToolboxXML: 748.876953125ms
```

paulkaplan:
It looks like there is an issue with the blocks that contain menus specific to the sprite like the "play sound" and "switch costume" blocks, they are not updating when switching sprites.

I'm not sure exactly what is going on there, they may have the same ID but should not have the same ID, or something else.

gnarf:
> I'm not sure exactly what is going on there, they may have the same ID but should not have the same ID, or something else.

https://github.com/LLK/scratch-gui/pull/1787

Yeah - the shadows for these blocks were getting recycled, changing them to have a targetId based id like the others fixed it.

gnarf:
I'm pretty sure this is causing a bit of a bug with the flyout scrollbar at the moment too.

gnarf:
I still want the review, perhaps it will help me solve the bugs ;)

core/flyout_base.js
```diff
@@ -55,6 +55,14 @@ Blockly.Flyout = function(workspaceOptions) {
   this.workspace_ = new Blockly.WorkspaceSvg(workspaceOptions);
   this.workspace_.isFlyout = true;
 
+  // When we create blocks for this workspace, instead of using the "optional" id
+  // make the default `id` the same as the `type` for easier re-use.
+  var newBlock = this.workspace_.newBlock;
+  this.workspace_.newBlock = function(type, id) {
+    // Use `type` if `id` isn't passed. `this` will be workspace.
+    return newBlock.call(this, proto, id || type);
```
rachel-fenichel: What is `proto` here?

core/flyout_base.js
```diff
@@ -466,7 +482,22 @@ Blockly.Flyout.prototype.show = function(xmlList) {
       var tagName = xml.tagName.toUpperCase();
       var default_gap = this.horizontalLayout_ ? this.GAP_X : this.GAP_Y;
       if (tagName == 'BLOCK') {
-        var curBlock = Blockly.Xml.domToBlock(xml, this.workspace_);
+
+        // We assume that in a flyout, the same block id/type means
+        // the same output SVG nodes.  Check for a recycled element.
```
rachel-fenichel: You're assuming that the same id/type means the same output `Blockly.BlockSvg` object, not just the same SVG nodes.  Please update the comment.

core/flyout_base.js
```diff
@@ -466,7 +482,22 @@ Blockly.Flyout.prototype.show = function(xmlList) {
       var tagName = xml.tagName.toUpperCase();
       var default_gap = this.horizontalLayout_ ? this.GAP_X : this.GAP_Y;
       if (tagName == 'BLOCK') {
-        var curBlock = Blockly.Xml.domToBlock(xml, this.workspace_);
+
+        // We assume that in a flyout, the same block id/type means
```
rachel-fenichel: Pull all of this code out into a helper function.  It'll make merges easier.

core/flyout_base.js
```diff
@@ -466,7 +482,22 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
+          return block.id === id;
+        });
+        var curBlock;
+
+        if (recycled > -1) {
+          // Delete the recycled element out of the recycle bin and
+          // assign it to curBlock.
+          curBlock = this.recycleBlocks_.splice(recycled, 1)[0];
+        } else {
+          var curBlock = Blockly.Xml.domToBlock(xml, this.workspace_);
```
rachel-fenichel: No need to redeclare `curBlock`.

core/flyout_base.js
```diff
@@ -466,7 +482,22 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
+        var id = xml.getAttribute('id') || xml.getAttribute('type');
+        var recycled = this.recycleBlocks_.findIndex(function(block) {
+          return block.id === id;
+        });
+        var curBlock;
+
+        if (recycled > -1) {
+          // Delete the recycled element out of the recycle bin and
+          // assign it to curBlock.
+          curBlock = this.recycleBlocks_.splice(recycled, 1)[0];
```
rachel-fenichel: @kchadha @paulkaplan Recycling a block this way means that there will be no delete or create events fired.  Will the VM handle that correctly?

core/flyout_base.js
```diff
@@ -514,6 +545,13 @@ Blockly.Flyout.prototype.show = function(xmlList) {
   this.listeners_.push(Blockly.bindEvent_(this.svgBackground_, 'mouseover',
       this, deselectAll));
 
+  // Clean out the old recycle bin.
```
rachel-fenichel: Make this a separate function (I suggest `emptyRecycleBin`).

core/flyout_base.js
```diff
@@ -807,3 +845,15 @@ Blockly.Flyout.prototype.placeNewBlock_ = function(oldBlock) {
   block.moveBy(finalOffsetMainWs.x, finalOffsetMainWs.y);
   return block;
 };
+
+/**
+ * Put a previously created block into the recycle bin, used during large
+ * workspace swaps to limit the number of new dom elements we need to create
+ *
+ * @param {!Blockly.BlockSvg} block The block to recycle.
```
rachel-fenichel, line 867: @private if possible, and end the name with an underscore if so.

rachel-fenichel (commented):
Quick notes
- using `type` instead of `id` assumes that the first block of a given type in the flyout is the same as the first block of that type the last time the flyout was shown.  Is that true?  If not, I'd rather just use the ID and make sure that the IDs generated in scratch-vm are correct.

However, that means that anyone else using the library will need to know to set their IDs correctly in the flyout in order to get this performance benefit.  That would belong in the documentation.  

More broadly, is ID the right thing to use for this comparison?  Do we care about keeping values the same on fields when we do these switches?  What about fields that have different possible values when you switch sprites?  (I think that last is solved by https://github.com/LLK/scratch-gui/pull/1787).

- @kchadha an issue with potential variables: the potential variable map is still getting cleared (flyout_base line 674, in `clearOldBlocks_`), but not being regenerated on any blocks that are recycled--since making the potential variables happens in XML parsing.

- background rectangles, checkboxes and mouse listeners look like they're getting reset correctly.

core/flyout_base.js
gnarf: `type` but i changed that in a followup commit and forgot to change that, sorry. will fix

gnarf:
> Quick notes

<3 thanks - this will help me catch a lot of the bugs I'm seeing I think :)

gnarf:
Tracked down the "scrollbar bug" -- setting translate wasn't enough for the recycle process - 1f6638a

kchadha:
@gnarf, The most recent changes look good to me! @rachel-fenichel, there was a new property added `"dynamic" = "true"` which we can use to specify which blocks should not be recycled...

rachel-fenichel:
API questions:
- should the default be that blocks are recycled, or not recycled?
- should the designation be by block type or actual block in the xml?

Default: recycle:
- Pro: Good performance with no extra work
- Con: If someone uses a variable and doesn't do anything special, they will have a world of problems with variables and not much information on why.

Default: no recycle:
- Pro: Guaranteed correctness if you do no work
- Con: Worse performance if you do not additional work

By block type:
- Pro: can have an explicit list of types in one place, with no need to edit the toolbox xml
- Pro: could also put the information on the block itself, in the block definition
- Con: maybe there are uses of the same block type where some need to be recycled and some shouldn't be--but I can't think of any offhand

By block:
- Pro: All information is in the toolbox XML
- Con: All information has to be in the toolbox xml (especially a problem if the default is no recycle, so lots of blocks need to be marked)

gnarf:
@rachel-fenichel I've been asking these questions for myself too, thanks for the pros/cons list! I don't have much to add to your list so I'll just summarize my opinions here:

> should the default be that blocks are recycled, or not recycled?

I believe the default should be to recycle.  This is a pretty big performance improvement, avoiding a lot of dom manip and GC, and we should encourage a "recyclable" toolbox by default.  I feel like with perf improvements, we should opt-out not opt-in.

I also think a near-future goal should be to find a way to reuse even more of these blocks by updating their shadows, wrangling the variables problem, or whatever we need to do to make more things recyclable.

We could probably even save even more time swapping workspaces by adding some recycling ability to the main workspaces.  We can reuse old blocks to avoid the GC thrash/double DOM Manip (delete + add rather than mutate).  We'd just need to figure out how to do so while also updating for new children/ids/etc.

> should the designation be by block type or actual block in the xml?

There is something very sweet sounding about the toolbox "dont recycle" flag being per block type. Since the block types / templates are defined in this repo, we wouldn't have to update multiple repos to tweak recycling settings.  I think having an override on the XML is also a decent (easily written in code) option.

picklesrus:
One other consideration wrt defaults
Thinking specifically about the use case of extensions, I tend to lean toward correctness for those blocks :-/  Perhaps different defaults for different types of extensions (e.g. core and official ones vs unofficial ones)?  I think it is worth asking someone who has been talking to extension developers  what the expectations are for understanding the blocks this level of detail. I don't have a great sense for how hard that would be to implement/expose in  the extension api or whether the distinction between per block type or per block in the xml matters there either.

thisandagain:
Re: extensions. For all "core" and "team" extensions I think I would be ok with having blocks default to "recycle" and using the `dynamic = true` setting to resolve specific issues. For 3rd party extensions (both "official" and "unofficial") I think these performance details are going to be confusing / difficult to describe and my recommendation would be to default to not recycling.

That said, much of this work regarding extensions is still in process and due to the high potential for confusing bugs to occur, I have a preference that we default to not recycling and then explicitly set specific blocks to `dynamic = false` (?) to enable this behavior. This will cause a lot of clutter in the  block definitions, but I think we need to be striving for stability and need to be cautious so close to launch.

mrjacobbloom:
>For 3rd party extensions (both "official" and "unofficial") I think these performance details are going to be confusing / difficult to describe and my recommendation would be to default to not recycling.

How possible is it, when importing an extension, to make sure that a block is static and mark it as recyclable? That'd at least include blocks with static text and no inputs, right? It'd be cool if the default for 3rd-party extensions was to recycle as much as you can verify is recyclable and let the developer push that further if they so desire

rachel-fenichel:
I strongly prefer that we 
- default to not recyclable
- set recyclable as a property on block definitions, not xml
- do not allow an override in the toolbox xml

My reasoning:
- Correctness, correctness, correctness.  It's more important than performance to me.  Developers are not going to be able to debug some of the problems that come up with recycling going wrong (we're going to have to dig deep on some of them), and they also may not know that their weird problems are because of recycling.  On top of that, they often won't get in touch to say what is broken--instead they just muddle through, but lose trust in the library.
- Per-block should be sufficient, and makes it easier to describe to developers which blocks are safe to recycle and which aren't (e.g. "Blocks with just text are safe to recycle" or "Blocks with non-dynamic dropdown menus are safe to recycle" or "Blocks with dynamic dropdowns are unsafe" or "Blocks with variables are unsafe").
- By blockly's rules, anything that is added to the xml spec has to be supported indefinitely--we commit to being able to load any xml from any workspace that has ever been saved, as long as you provide the block definitions.  Adding to the XML spec is therefore not something we do lightly.
  - I strongly advise this for Scratch-Blocks as well, but at minimum it should be an explicit decision.
- Yes, we should figure out what to do about variables so that it is safe to recycle blocks with variable references, but that's going to be tricky and I'd prefer to do it after we land basic recycling of everything easy.

gnarf:
> I strongly prefer that we
> 
> default to not recyclable
> set recyclable as a property on block definitions, not xml
> do not allow an override in the toolbox xml

Agreed! I'm thinking rework this to add `recycle: true` property to the blocks defined here in scratch-blocks, that way extensions can opt-in.

It would be nice to get sane defaults without any effort somehow from the extensions API though.  I don't think we want to be teaching extension developers in the long run to be adding `recycle: true` to save ~2-3ms on a macbook pro(let alone ipad) per block they don't mark per sprite swap do we?

(previously stated 10-20ms was a made up number, not supported by recent testing, but given that we recycle about 100 blocks in my test and it's saving 200-300ms due to GC/destroy/rebuild I updated that number -- still 2-3ms = adding 10 blocks in your extension = 1 frame wasted at 30fps)

rachel-fenichel:
What if we make the recyclable property true/false/auto-if-not-defined in the block definition spec?

Developers who know that they want a block to be recycled can mark as true.
Developers who know that a block is not recyclable can mark as false.
Developers who don't know what to do just don't put anything down, and scratch-blocks will decide whether to make it recyclable or not.

As a first pass, `auto` will just mean recyclable is false.  As we come up with good rules of thumb for recyclability, we can add those into the `auto` code path.  That could look like inspecting to see if all of the fields are of types that we know are pretty safe.

@picklesrus thoughts, and is the above proposal overengineering in terms of what it adds to the block spec?

rachel-fenichel:
Normally when we put something into the block definition we can just go inspect it directly from the block.  In this case we want to know something about the block without actually having an instance of it--need to decide whether to try to recycle at the point that we have only the block's type.  Does that mean that the workspace gets an object with the mapping from block type to recyclability, which it builds up as blocks of each type are used for the first time?

cwillisf:
For the purposes of Scratch 3.0 specifically: in the long run, I don't think it will matter much if the `recycle` property is default true, false, or auto. The plan is that eventually, all Scratch 3.0 blocks will be generated by the "extension" system (and similar systems for things like variables and "my blocks"), which means that we'll only need to set the property in one place. If there's complex logic to decide when a Scratch 3.0 extension block is recyclable, at least it will be contained in one place.

For non-Scratch 3.0 consumers of scratch-blocks, I think I like the proposal of a "true/false/auto-if-not-defined" property. Like @rachel-fenichel suggested, I think it makes sense for "auto" to mean "don't recycle" initially, and we can make it smarter as we have time to do so without sacrificing correctness.

gnarf:
> In this case we want to know something about the block without actually having an instance of it

We have the recycled `BlockSvg` (we can't recycle it if we hadn't created it yet anyway).  We could mark the block itself recyclable during creation?

rschamp:
Merging this into the performance branch for smoke testing! @gnarf Please keep this branch around so work can continue on it for the merge into develop if all goes well.

PR #1452, opened 2018-04-17
Switch to Blockly.Field.fromJson

rachel-fenichel:
### Resolves

None

### Proposed Changes

See https://github.com/google/blockly/pull/1594 and https://github.com/google/blockly/pull/1592

This is the combination of those two, applied to scratch-blocks.

### Reason for Changes

Make it easier to add fields without changing core blockly/scratch-blocks.  close gap in API between blockly and scratch-blocks.

### Test Coverage

jsunit tests pass.  A few new tests added.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -1408,76 +1408,20 @@ Blockly.Block.prototype.interpolate_ = function(message, args, lastDummyAlign) {
             case 'input_dummy':
               input = this.appendDummyInput(element['name']);
               break;
-            case 'field_label':
-              field = Blockly.Block.newFieldLabelFromJson_(element);
-              break;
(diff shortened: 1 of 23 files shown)


picklesrus:
Code looks good. IIUC, the only way to verify you've got them all is to interact with a block that has the field type and make sure it still works.  Do all the types exist in the vertical or horizontal playground? 

If not, seems worth being mindful of the scratch 3 smoke testing and release schedule when merging. Seems like this would be a reasonable thing to add to the smoke test list.

rachel-fenichel:
Flyouts load correctly in both horizontal and vertical playgrounds.

kchadha (approved):
LGTM

PR #1453, opened 2018-04-17
Local build script

rachel-fenichel:
### Resolves
None

### Proposed Changes

See https://github.com/google/blockly/pull/1772

### Reason for Changes

Ergonomics
### Test Coverage

Used the new built file in the playground to test.


local_build/local_build.sh
```diff
@@ -0,0 +1,70 @@
...
+# Future work:
+# - Trim down Google's Apache licenses, to match the output of build.py.
+# - Generate other compressed files generated by build.py normally.
+# - Add a good error message if multiple versions of the closure compiler were
+#   found.
+# - Add a flag for generating horizontal (still default to vertical).
+
+# Find the Closure Compiler.
+if [ -f "$(npm root)/google-closure-compiler/compiler.jar" ]; then
+  COMPILER="$(npm root)/google-closure-compiler/compiler.jar"
```
thisandagain: If we are going to use this, I would recommend adding https://www.npmjs.com/package/google-closure-compiler to the `devDependencies` in the `package.json` file.

local_build/local_build.sh
```diff
@@ -0,0 +1,70 @@
...
+  echo "Download from this URL, and place jar file in current directory."
+  echo "https://dl.google.com/closure-compiler/compiler-latest.zip"
+  exit 1
+fi
+
+echo Using $COMPILER as the compiler.
+rm local_blockly_compressed_vertical.js 2> /dev/null
+echo Compiling Scratch-Blocks..
+java -jar $COMPILER \
+  --js='../core/**.js' \
```
thisandagain: Is there a reason we should have this inside an enclosing `local_build` directory that the script must be run from? My instinct would be to have this at the top level along with the `build.py` file.

thisandagain (approved):
It works! Thanks @rachel-fenichel. Two fairly minor questions for you.

local_build/local_build.sh
rachel-fenichel: I agree that putting it at the top level would match what we're already doing, but I'm not stoked on how many files are at the top level.  Moving the existing files seems like a pain, so I'm just trying to avoid adding new ones.  

This adds three (closure-compiler, the built file, and this file).  

If you want it at the top level, I'll move it.  Just let me know.

local_build/local_build.sh
mzgoddard: I think we should require use of `google-closure-compiler` in `devDependencies` so builds are deterministic between different computers because they will use the same compiler version.

rachel-fenichel: FYI I haven't actually tested this branch of the if--I pulled it from our compiler tests and it looked reasonable.  Happy to add `google-closure-compiler` to `devDependencies`, though.  Give me a sec.

local_build/local_build.sh
thisandagain: I think I'd rather move it to the top level (even though I agree it's pretty crowded). Needing to `cd` into `local_build` feels like it may cause confusion.

mzgoddard:
@thisandagain @rachel-fenichel I took a shot at this kind of thing last week and got a working version with dirty changes in the python script that I want to clean up. https://github.com/mzgoddard/scratch-blocks/tree/local-closure Thought I should mention it here and ask for advice on how I may proceed. I don't want to block this PR with my work.

rachel-fenichel:
Updated, please take a look.

@mzgoddard I'll check that out!  Do you want me to look now our wait until you clean it up a bit?

mzgoddard:
@rachel-fenichel probably after I clean it up a bit.

thisandagain (approved):

thisandagain:
Thanks for making those changes @rachel-fenichel. I think this is good to land.

PR #1458, opened 2018-04-19
Update cleanup.sh to handle more files automatically

rachel-fenichel:
### Resolves

None
### Proposed Changes

Improves the cleanup script for helping blockly->scratch-blocks merges.  It now deals with most files that are not in core.
- Handle more files
- Clean up output

If I'm missing out on best practices for scripts like this, please let me know.

### Reason for Changes

Ergonomics.

### Test Coverage

I pulled from blockly (`git pull https://github.com/google/blockly.git develop`) and ran this.

### DIFF
--- a/cleanup.sh
+++ b/cleanup.sh
@@ -4,29 +4,92 @@
 # Removes files and directories that scratch-blocks doesn't want.
 # Rachel Fenichel (fenichel@google.com)
 
-# On separate lines so that a failure to find one doesn't block removal of the other directories.
-git rm -rf accessible
-git rm -rf demos
-git rm -rf tests/generators
-git rm -rf appengine
-git rm -rf blocks
-git rm blockly_compressed.js
-git rm blockly_uncompressed.js
-git rm blocks_compressed.js
-git rm -f tests/playground.html core/block_render_svg.js
+# Note: 'ours' is scratch-blocks, 'theirs' is blockly.
 
+# Formatting helpers.
+indent() { sed 's/^/  /'; }
+indent_more() { sed 's/^/\t/'; }
+empty_lines() { printf '\n\n'; }
+
+
+empty_lines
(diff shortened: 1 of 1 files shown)


kchadha:
@rachel-fenichel, looks okay to me. I think there should be more comments about usage up at the top, though it *does* look like using it incorrectly should not break things. I'm worried about weird edge cases and accidentally `rm -rf`ing the wrong thing if this is run this from the wrong directory or something. Maybe there should be a check and early bail out if this is run from the wrong directory...?

thisandagain (approved):
This looks great. Thanks @rachel-fenichel. 😄

rachel-fenichel:
@kchadha note that it's `git rm -rf`, so failure won't actually `rm -rf` all the things.  Also that it's done in the context of a git merge, so you usually can just back out with `git merge --abort` (which is what I did repeatedly while testing it).

What additional comments/checks/bail outs do you want me to add?  Happy to do so, but I don't know how to check that I'm in the right place.

kchadha:
@rachel-fenichel ah, oops yep re: `git rm -rf`.  The fact that it's done in the context of a merge should be documented somewhere (within comments in the script itself, a --help, in the README of this repo, or maybe even in the name of the script) b/c there's nothing really stopping users from running `./cleanup.sh` at any other point.

The script also has `cd`s which assume that you're running this script from the directory it's located in. You can add a check at the top which checks if the current working directory is the same as the directory that the script itself is located in. I only did a quick google search for this, but this might be useful:
https://stackoverflow.com/questions/59895/getting-the-source-directory-of-a-bash-script-from-within

FWIW, this PR is good to go, these suggestions don't need to block this merge!

rachel-fenichel:
Cool, I'm going to merge this and put those concerns into an issue to follow up on, because these scripts weren't intended to take a lot of time this week.

re figuring out what directory you're running from: yup, makes sense.

PR #1463, opened 2018-04-20
BlockSvg.isRecyclable

gnarf:
Follow-up that catches most of the non-recyclable blocks (I think).


gnarf:
This targets the temporary "performance" branch we are using for smoke testing - we will still have the opportunity to do a follow-up PR/review of the whole flyout recycle PR before landing on develop - @rschamp do you want to just land this (it fixes one or two buggy toolbox blocks)?

gnarf:
merging to perf branch

PR #1464, opened 2018-04-20
Recycle blocks in Flyouts

gnarf:
Re-opening this from #1442 to point at develop for now for better overall PR review / discussion

### Proposed Changes

* Adds the ability for toolbox blocks to be reused between toolbox show calls.  

### Reason for Changes

* Each recyclable block saves us 2-3ms on my macbook pro.

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -276,27 +276,24 @@ Blockly.BlockSvg.prototype.getIcons = function() {
  * @param {Blockly.BlockSvg} newParent New parent block.
  */
 Blockly.BlockSvg.prototype.setParent = function(newParent) {
-  if (newParent == this.parentBlock_) {
+  var oldParent = this.parentBlock_;
+  if (newParent == oldParent) {
     return;
   }
-  var svgRoot = this.getSvgRoot();
-  if (this.parentBlock_ && svgRoot) {
-    // Move this block up the DOM.  Keep track of x/y translations.
-    var xy = this.getRelativeToSurfaceXY();
-    // Avoid moving a block up the DOM if it's currently selected/dragging,
-    // so as to avoid taking things off the drag surface.
-    if (Blockly.selected != this) {
-      this.workspace.getCanvas().appendChild(svgRoot);
-      this.translate(xy.x, xy.y);
-    }
-  }
-
   Blockly.Field.startCache();
   Blockly.BlockSvg.superClass_.setParent.call(this, newParent);
   Blockly.Field.stopCache();
 
+  var svgRoot = this.getSvgRoot();
+
+  // Bail early if workspace is clearing, or we aren't rendered.
+  // We won't need to reattach ourselves anywhere.
+  if (this.workspace.isClearing || !svgRoot) {
+    return;
+  }
+
+  var oldXY = this.getRelativeToSurfaceXY();
   if (newParent) {
-    var oldXY = this.getRelativeToSurfaceXY();
     newParent.getSvgRoot().appendChild(svgRoot);
     var newXY = this.getRelativeToSurfaceXY();
     // Move the connections to match the child's new position.
@@ -307,6 +304,17 @@ Blockly.BlockSvg.prototype.setParent = function(newParent) {
           newParent.getColourTertiary());
     }
   }
+  // If we are losing a parent, we want to move our DOM element to the
+  // root of the workspace.
+  else if (oldParent) {
+    // Avoid moving a block up the DOM if it's currently selected/dragging,
+    // so as to avoid taking things off the drag surface.
+    if (Blockly.selected != this) {
+      this.workspace.getCanvas().appendChild(svgRoot);
+      this.translate(oldXY.x, oldXY.y);
+    }
+  }
+
 };
 
 /**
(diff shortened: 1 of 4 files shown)


core/block_svg.js
```diff
@@ -272,27 +272,24 @@ Blockly.BlockSvg.prototype.getIcons = function() {
  * @param {Blockly.BlockSvg} newParent New parent block.
  */
 Blockly.BlockSvg.prototype.setParent = function(newParent) {
-  if (newParent == this.parentBlock_) {
+  var oldParent = this.parentBlock_;
```
gnarf: this is a totally separate optimization here (setParent) that will get landed separately - ignore for now, it's just a base difference (i'm on "performance")

core/workspace.js
```diff
@@ -110,6 +110,13 @@ Blockly.Workspace = function(opt_options) {
  */
 Blockly.Workspace.prototype.rendered = false;
 
+/**
```
gnarf: this also is the diff from another PR (which exists in my "performance" base branch, please ignore for now)

rachel-fenichel (commented):
How has this been behaving on the performance branch?

We may want to use the xml representation of the block as the key for recycling (possibly without the ID), rather than just the type.  Then blocks that are being recycled have the same contents at a deep level, instead of just the same type.

The main argument against doing that is if getting the XML representation of a block is expensive.  I haven't benchmarked it.

rschamp:
There are still a few issues with custom procedures and variables on the performance branch which are either caused by this PR or https://github.com/LLK/scratch-gui/pull/1784. We should figure out which.  I think since this PR will make the gains of https://github.com/LLK/scratch-gui/pull/1784 less noticeable, we should try reverting https://github.com/LLK/scratch-gui/pull/1784 out of the performance branch and re-test. What do you think @gnarf?

gnarf:
I think that the GUI PR is the more stable one -- This is probably the culprit -- I'm going to be looking into that today - will update with more info when I have it.

I was gonna do something about it last week but the develop builds were in some weird state that made it hard for me to work on.

gnarf:
so this branch ended up picking up the `block.setParent  during clearing workspace` commit when I rebased from performance and develop --- Just keeping them together

Freshly minted commit 821b13e is "recycle" in its entirety (squashed down to a single commit)

gnarf:
> There are still a few issues with custom procedures

I did make some changes for the `procedures_call` - they will no longer be recycled which seemed to clear up a lot of the smoke test errors.

gnarf:
just rebased on latest merges

gnarf:
Alright - rather than split these back up into two PR's again, could I get a final 👀 from y'all again on landing this?  Remember you can compare these changes (and a few from gui as well) here:

* https://llk.github.io/scratch-gui/performance/#208914317 - includes this branch (flyout-recycle) as well as https://github.com/LLK/scratch-gui/pull/1973
* https://llk.github.io/scratch-gui/#208914317 - "baseline" develop build

core/block_svg.js
```diff
@@ -1385,3 +1393,42 @@ Blockly.BlockSvg.prototype.scheduleSnapAndBump = function() {
...
 };
+
+/**
+ * Determine if this block can be recycled, basic non-dynamic shadows and no variables
+ * @return {boolean} true if the block can be recycled
+ */
+Blockly.BlockSvg.prototype.isRecyclable = function() {
+
+  // The procedures_call type is always dynamic as it can be mutated at runtime.
+  if (this.type === 'procedures_call') {
```
rachel-fenichel: What about other things that can be mutated at runtime?  The general rule is probably "if mutationToDom and domToMutation are defined on the block, it's not recyclable".

core/procedures.js
```diff
@@ -237,6 +237,7 @@ Blockly.Procedures.flyoutCategory = function(workspace) {
     //   <mutation ...></mutation>
     // </block>
     var block = goog.dom.createDom('block');
+    block.setAttribute('id', 'proccode:' + mutation.proccode );
```
rachel-fenichel: Unrelated change?

core/block_svg.js
gnarf: Thanks! I'm sure this would catch other similar extension blocks in the future. -- Running a quick test to see what else might match this.

core/procedures.js
gnarf: This was one of the first approaches I took, giving the procedure_call's id's - I don't think it hurts anything, but also is probably not necessary - can remove it.

core/block_svg.js
gnarf: the only other blocks this caught were already `false` by some other check, so I updated accordingly.

rachel-fenichel (commented):
Looks good from my side.

core/block_svg.js
```diff
@@ -1385,3 +1393,42 @@ Blockly.BlockSvg.prototype.scheduleSnapAndBump = function() {
...
   }, Blockly.BUMP_DELAY);
 };
+
+/**
+ * Determine if this block can be recycled, basic non-dynamic shadows and no variables
+ * @return {boolean} true if the block can be recycled
+ */
+Blockly.BlockSvg.prototype.isRecyclable = function() {
+
+  // If the block needs to parse mutations, it's probably safest to never recycle.
```
kchadha, line 1403: +1

rschamp (approved):
Thank you @gnarf!

PR #1469, opened 2018-04-23
Move all block strings into messages.js

thisandagain:
## Changes
- Moves all block messages and menus into `msg/messages.js`
- Modifies the `build.py` script to not include localization scripts (a more thorough removal of existing `i18n` infrastructure will come later)

## Adds
- New Node.js utility for converting messages into JSON that can be used by Transifex
- New Node.js utility for converting languages provided by Transifex back into JavaScript
- New NPM script for executing both utilities (`npm run translate`)

## Removes
- Unused block definitions (e.g. "pen", "music", and "video motion" blocks that have been migrated to extensions
- Unused Blockly messages

## Notes
- This PR is pretty big so I recommend going through it commit by commit
- You'll need to look at the full diff on `msg/messages.js`, `msg/json/en.json`, and `msg/js/en.js`


build.py
```diff
@@ -502,4 +502,4 @@ def exclude_horizontal(item):
   Gen_compressed(search_paths_vertical, search_paths_horizontal).start()
 
   # This is run locally in a separate thread.
-  Gen_langfiles().start()
+  # Gen_langfiles().start()
```
thisandagain, line 505: This is gross, but temporary until I can fully remove all of the old `i18n` fixtures from Blockly which will be fairly invasive.

msg/js/en.js
```diff
@@ -3,429 +3,195 @@
...
-Blockly.Msg.MATH_HUE = "230";
-Blockly.Msg.LOOPS_HUE = "120";
-Blockly.Msg.LISTS_HUE = "260";
-Blockly.Msg.LOGIC_HUE = "210";
-Blockly.Msg.VARIABLES_HUE = "330";
-Blockly.Msg.TEXTS_HUE = "160";
-Blockly.Msg.PROCEDURES_HUE = "290";
-Blockly.Msg.COLOUR_HUE = "20";
\ No newline at end of file
+Blockly.Msg["CONTROL_FOREVER"] = "forever";
```
thisandagain, line 8: I'm enclosing the keys here in bracket notation property accessors for safety.

msg/messages.js
```diff
@@ -20,1220 +20,275 @@
...
-/// For more context, see
-/// [[Translating:Blockly#infrequent_message_types]].\n{{Identical|Item}}
-Blockly.Msg.VARIABLES_DEFAULT_NAME = 'item';
-/// button text - Button that sets a calendar to today's date.\n{{Identical|Today}}
-Blockly.Msg.TODAY = 'Today';
-
-// Context menus.
-/// context menu - Make a copy of the selected block (and any blocks it contains).\n{{Identical|Duplicate}}
+// Control blocks
+Blockly.Msg.CONTROL_FOREVER = "forever";
```
thisandagain, line 34: Where possible I've tried to match the opcode pattern (e.g. `CATEGORY_OPCODE`). For menu items I'm using the convention of `CATEGORY_OPCODE_MENUITEM`.

package.json
```diff
@@ -17,21 +17,25 @@
     "test:unit": "node tests/jsunit/test_runner.js",
     "test:lint": "eslint .",
     "test": "npm run test:lint && npm run test:setup && npm run test:unit",
-    "version": "json -f package.json -I -e \"this.repository.sha = '$(git log -n1 --pretty=format:%H)'\""
+    "version": "json -f package.json -I -e \"this.repository.sha = '$(git log -n1 --pretty=format:%H)'\"",
+    "translate": "node i18n/js_to_json.js && node i18n/json_to_js.js"
```
thisandagain, line 21: Eventually this will also include methods to push and pull from Transifex.

rachel-fenichel:
@AnmAtAnm for blockly-side comments on messages.

package.json
chrisgarrity: Just FYI, we usually only push on freeze.

i18n/json_to_js.js
```diff
@@ -0,0 +1,49 @@
+const async = require('async');
+const fs = require('fs');
+const path = require('path');
+const glob = require('glob');
+
+// Globals
+const PATH_INPUT = path.resolve(__dirname, '../msg/json/*.json');
```
chrisgarrity, line 7: We could pull directly from Transifex, but if we let scratch-l10n manage the pull it can also run tests to check that translators didn't mess up the json. If scratch-l10n exports the .js files, maybe scratch-blocks can just handle things in webpack.

One thing to note is that developers need transifex credentials to push/pull translations.

chrisgarrity (approved):
This is a good starting point. I think we'll want to isolate the transifex dependencies from scratch-blocks as much as possible.

i18n/json_to_js.js
thisandagain: These fixtures are still needed when the locales are eventually pulled in from `scratch-l10n` there will just be some path manipulation needed.

AnmAtAnm:
Looks good.

Blocks and strings look good in the various block definition files. I notice your `messages.js` does not include translator comments. Obviously the comments for Blockly blocks don't apply, but I would expect translators to possibly need additional context surround some strings, particularly those that require %1 arguments.

The `js_to_json` and `json_to_js` look like pretty accurate recreations of the relevant parts of `build.py`, though I didn't run it.

rachel-fenichel (approved):

AnmAtAnm:
FYI - We're [updating](https://github.com/google/blockly/pull/1829) Blockly code to use array key references to all `Msg` constants, and removing the `@export` annotations. This is critical for apps that either compile those constants, or rely on runtime manipulation of the constants. `@export` was duplicating the string table, instead of just renaming it.

Further, Rachel pointed out that the end goal is run-time language swapping.  For this to work, the blocks will need to use the `"%{BKY_MESSAGE_ID}"` notation, in order to defer the string lookup until block creation time. Direct `Blockly.Msg` references, as seen in this PR, will be reified once when the block definition is read/executed, and will not pick up later `Blockly.Msg` changes.

PR #1471, opened 2018-04-24
Use the local closure compiler in build.py

mzgoddard:
### Resolves

Depending on a remote online compiler service.

### Proposed Changes

Use `google-closure-compiler` installed by npm to closure compile all of the intermediate assets build.py uses that webpack consumes.

- Test for library and compiler instead by npm is available
- Check that library tool can be loaded and that the compiler will give expected output
- Failing to load the library's calcdeps or run the compiler, fallback to the previous compilation configuration using the remote closure-compiler server
- Pass a `closure_env` object to `Gen_compressed` and `Gen_uncompressed` so they can use the closure environment state to build locally or remotely
- Format javascript chunks in `Gen_uncompressed` with a helper to replace python format expressions with values in the `closure_env`
- Reorganize compressed compiling so the compiling part itself can be swapped between compiling locally and remotely
- Configure options for the compressed builds so they can compile locally and remotely
- The remote compiler takes additional options that do not exist for the command line compiler interface
  - The compressed blockly files have some distinct local and remote options that do_compile_local and do_compile_remote filter out
  - The remote compiles add options for the remote server
  - Use the compilation_level value SIMPLE as its the more up to date name, the remote handler renames the value to SIMPLE_OPTIMIZATIONS for the remote compiler

### Reason for Changes

1. The remote compiler at closure-compiler.appspot.com is rate limited. Building more than a certain number of times in one hour can easily deplete the available count per IP, leaving a developer iterating scratch-blocks to work around not being able to run changes until the next hour.
2. Documentation on the remote compiler regarding compiler and library version are either hard to find (I couldn't find any) or not available. If those versions changed we may not know until a change breaks a build of scratch-blocks.
3. We cannot control what version of the closure library and closure compiler are used by the remote service. The locally installed closure library version is not the version the remote compiler uses, the library output is different.


build.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python2.7
+#!/usr/bin/env python
```
rachel-fenichel: What changes did you make to make this work with either version of python?

build.py
```diff
@@ -136,15 +157,15 @@ def run(self):
 
     provides = []
     for dep in calcdeps.BuildDependenciesFromFiles(self.search_paths):
-      if not dep.filename.startswith(os.pardir + os.sep):  # '../'
+      if not dep.filename.startswith(self.closure_env["closure_root"] + os.sep):  # '../'
```
rachel-fenichel: Check that the comment at the end of the line is still accurate.

build.py
```diff
@@ -464,42 +566,74 @@ def exclude_horizontal(item):
...
+          closure_root, closure_library, "closure", "bin", "calcdeps.py"))
+    except ImportError:
+      if os.path.isdir(os.path.join(os.path.pardir, "closure-library-read-only")):
+        # Dir got renamed when Closure moved from Google Code to GitHub in 2014.
+        print("Error: Closure directory needs to be renamed from"
+              "'closure-library-read-only' to 'closure-library'.\n"
+              "Please rename this directory.")
+      elif os.path.isdir(os.path.join(os.path.pardir, "google-closure-library")):
+        # When Closure is installed by npm, it is named "google-closure-library".
+        #calcdeps = import_path(os.path.join(
```
rachel-fenichel: Remove dead code.

build.py
```diff
@@ -464,42 +566,74 @@ def exclude_horizontal(item):
 
 if __name__ == "__main__":
   try:
+    closure_dir = CLOSURE_DIR_NPM
+    closure_root = CLOSURE_ROOT_NPM
+    closure_library = CLOSURE_LIBRARY_NPM
+    closure_compiler = CLOSURE_COMPILER_NPM
+
+    # Load calcdeps from the local library
     calcdeps = import_path(os.path.join(
```
rachel-fenichel, line 590: Why do this before you validate the compiler you're using?

build.py
```diff
@@ -259,43 +300,94 @@ def gen_generator(self, language):
...
+    json_data = do_compile(params, target_filename)
+
+    if self.report_errors(target_filename, filenames, json_data):
+      self.write_output(target_filename, remove, json_data)
+      self.report_stats(target_filename, json_data)
+
+  def do_compile_local(self, params, target_filename):
+      filter_keys = ["use_closure_library"]
+
+      dash_params = [("--" + arg, value) if arg != "js_file" else (value,) for (arg, value) in params]
```
rachel-fenichel: Add some comments--these five lines are super dense.

rachel-fenichel (commented):
Looks reasonable.  Also get someone on scratch-team to take a look and make sure it won't cause them any issues.

rschamp (approved):
This looks great from my perspective! Thank you! Building on trains and planes! ✈️ 🚂 🛩

rschamp:
I think @rachel-fenichel's comments should be addressed before merging though!

build.py
mzgoddard: I didn't. I'm not sure what you mean by either version of python. `/usr/bin/env python` will run the version of `python` that first appears in the user's path.

mzgoddard: I probably made this change out of habit. I can change it back.

build.py
mzgoddard: Sure. If its helpful this code was already commented out before this PR. Its just now in a block that was indented more.

build.py
mzgoddard: The order does not really matter. We need to check that the calcdeps module imports and that the compiler does what we expect. If either calcdeps fails to import or the compiler fails to give the expected output the build cannot be done locally.

build.py
rachel-fenichel: I ask because there's an explicit check for python 2.x at the beginning of the script: https://github.com/LLK/scratch-blocks/pull/1471/files#diff-2955d5257f635de1df53f55a171ca5c7R39

Issue #1472, opened 2018-04-25
Change language on "mouse-pointer" blocks


thisandagain:
## Overview
Scratch 3.0 will work on a wide variety of devices including Chromebooks, Windows Surface, iPads, and Android tablets. Many of these devices either exclusively support touch or support a mixture of touch and mouse interaction. Because of this, we would like to retire use of the terms "mouse-pointer" and "clicked" and use the more generic terms "pointer" and "tapped". This will impact a large number of blocks, but feels important in order to properly support touch platforms moving forward.

#### Move
- [x] `go to (pointer)`
- [x] `glide (1) secs to (pointer)`
- [x] `point towards (pointer)`

#### Events
- [x] `when this sprite tapped`
- [x] `when stage tapped`

#### Sensing
- [x] `<pointer down?>`
- [x] `(pointer x)`
- [x] `(pointer y)`
- [x] `(distance to (pointer))`

#### Details
- [x] The `(distance to (pointer))` block should be moved down to below the `(pointer y)` block. It's current positioning in the "Sensing" category doesn't make much sense (particularly for a block that isn't used very heavily).

---
Moved from LLK/scratch-vm#1015
/cc @thisandagain @speakkvisually @kenny2github

ericrosenbaum:
I wonder if we will see any confusion related to the word "touching" on touch devices?

`touching (sprite)?`
`touching (color)?`
`(color) is touching (color)?`

Not sure if it would make sense to rename these though. One idea would be to change just the first two like this (though these are longer, unfortunately):

`this sprite is touching (sprite)?`
`this sprite is touching (color)?`

thisandagain:
@ericrosenbaum I share the concern. Should we file that as a follow-up item to #1473 (I mention a few other concerns in that PR as well)? It would be good to talk through these issues with @ntlrsk.

thisandagain:
Between negative feedback from resource partners that came in today (major changes to blocks coming in too late) and some of the questions / concerns around this I have closed the PR and recommend we table this for discussion post release.

/cc @carljbowman @ntlrsk @ericrosenbaum

PR #1476, opened 2018-04-26
Revert "mouse-pointer" changes in messages

thisandagain:
### Proposed Changes

Revert some changes that were made as part of #1472 

### Reason for Changes

Prepare `en.json` for localization


chrisgarrity (approved):
👍

PR #1477, opened 2018-04-26
Decompose functions in core/zoom_controls.js

rachel-fenichel:
### Resolves

Scratch blocks version of https://github.com/google/blockly/pull/1828

### Proposed Changes

Make some helper functions to create the zoom controls.

### Reason for Changes

More work for easing merges.   Makes it easier to mark which code is deliberately different in scratch blocks and blockly.

### Test Coverage

### DIFF
--- a/core/zoom_controls.js
+++ b/core/zoom_controls.js
@@ -121,104 +121,11 @@ Blockly.ZoomControls.prototype.top_ = 0;
  * @return {!Element} The zoom controls SVG group.
  */
 Blockly.ZoomControls.prototype.createDom = function() {
-  var workspace = this.workspace_;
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):
LG

paulkaplan:
It might be worth adding a comment about how/why the blockly/scratch-blocks implementations differ (looks like scratch-blocks doesn't use a clip-path, but I'm not exactly sure why. Unless you happen to know off-hand, I wouldn't worry about it)

rachel-fenichel:
I don't actually know--I think y'all just decided to use individual sprites, and I don't actually know why we use a clip-path.

PR #1478, opened 2018-04-26
Only use Blockly.Msg inside the init functions, since it doesn't exist yet in dist

paulkaplan:
Basically, our load order in the final webpack step that bundles together messages/blockly/blocks_vertical is putting messages at the end instead of in the middle. This fixes that while we wait on a real fix to the load order stuff.

### DIFF
--- a/blocks_vertical/looks.js
+++ b/blocks_vertical/looks.js
@@ -146,16 +146,6 @@ Blockly.Blocks['looks_hide'] = {
   }
 };
 
-Blockly.Blocks['looks_effect_menu_options'] = [
-  [Blockly.Msg.LOOKS_EFFECT_COLOR, 'COLOR'],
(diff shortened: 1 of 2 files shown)


rachel-fenichel:
@AnmAtAnm to discuss the % syntax for messages, here or offline.

rachel-fenichel (approved):

AnmAtAnm:
The `%{BKY_MESSAGEID}` syntax will defer the value lookup, which will disconnect the `Blockly.Msg` load ordering issue that looks like the problem that triggered this change.

Further, as noted in [this comment](https://github.com/LLK/scratch-blocks/pull/1469#issuecomment-385106088), the `%{..}` notation, instead of direct `Blockly.Msg` references, will allow the Block to pick up the latest value of `Blockly.Msg`. For instance, if the language is changed at run time. At least that is the intent; there may be minor issues with respect to drop down fields, like these because of the way all drop down fields share a single backing data structure.

PR #1479, opened 2018-04-26
Lint that caused conflicts with Blockly; mark things protected

rachel-fenichel:
### Resolves

None
### Proposed Changes

- Marks things protected if they're used by subclasses (solves build errors)
- Fix bad annotations on various functions
- Pulls over lots of lint changes (often indentation) directly from Blockly

Should have nothing that changes behaviour.  All of this is to reduce the area of merge conflicts so that I can get to a full merge soon.
### Reason for Changes

Ergonomics (of the merge); lint (getting ready to upgrade eslint); correct visibility of properties (important under advanced compilation).
### Test Coverage

Played in the vertical playground for a while.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,7 +6,16 @@
         "linebreak-style": ["error", "unix"],
         "max-len": ["error", 120, 4],
         "no-trailing-spaces": ["error", { "skipBlankLines": true }],
-        "no-unused-vars": ["error", {"args": "after-used", "varsIgnorePattern": "^_"}],
+        "no-unused-vars": [
+            "error",
+            {
+                "args": "after-used",
+                # Ignore vars starting with an underscore.
+                "varsIgnorePattern": "^_",
+                # Ignore arguments starting with an underscore.
+                "argsIgnorePattern": "^_"
+            }
+        ],
         "no-use-before-define": ["error"],
         "quotes": ["off"], # Blockly mixes single and double quotes
         "semi": ["error", "always"],
--- a/core/comment.js
+++ b/core/comment.js
@@ -79,7 +79,8 @@ Blockly.Comment.prototype.drawIcon_ = function(group) {
         'class': 'blocklyIconSymbol',
         'd': 'm6.8,10h2c0.003,-0.617 0.271,-0.962 0.633,-1.266 2.875,-2.405' +
           '0.607,-5.534 -3.765,-3.874v1.7c3.12,-1.657 3.698,0.118 2.336,1.25' +
-          '-1.201,0.998 -1.201,1.528 -1.204,2.19z'},
+          '-1.201,0.998 -1.201,1.528 -1.204,2.19z'
+      },
       group);
   // Dot of question mark.
   Blockly.utils.createSvgElement('rect',
--- a/core/comment.js
+++ b/core/comment.js
@@ -70,21 +70,27 @@ Blockly.Comment.prototype.drawIcon_ = function(group) {
   // Circle.
   Blockly.utils.createSvgElement('circle',
       {'class': 'blocklyIconShape', 'r': '8', 'cx': '8', 'cy': '8'},
-       group);
+      group);
   // Can't use a real '?' text character since different browsers and operating
   // systems render it differently.
   // Body of question mark.
   Blockly.utils.createSvgElement('path',
-      {'class': 'blocklyIconSymbol',
-       'd': 'm6.8,10h2c0.003,-0.617 0.271,-0.962 0.633,-1.266 2.875,-2.405' +
-       ' 0.607,-5.534 -3.765,-3.874v1.7c3.12,-1.657 3.698,0.118 2.336,1.25' +
-       ' -1.201,0.998 -1.201,1.528 -1.204,2.19z'},
-       group);
+      {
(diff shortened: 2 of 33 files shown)


core/comment.js
```diff
@@ -70,21 +70,27 @@ Blockly.Comment.prototype.drawIcon_ = function(group) {
...
-      {'class': 'blocklyIconSymbol',
-       'd': 'm6.8,10h2c0.003,-0.617 0.271,-0.962 0.633,-1.266 2.875,-2.405' +
-       ' 0.607,-5.534 -3.765,-3.874v1.7c3.12,-1.657 3.698,0.118 2.336,1.25' +
-       ' -1.201,0.998 -1.201,1.528 -1.204,2.19z'},
-       group);
+      {
+        'class': 'blocklyIconSymbol',
+        'd': 'm6.8,10h2c0.003,-0.617 0.271,-0.962 0.633,-1.266 2.875,-2.405' +
+          '0.607,-5.534 -3.765,-3.874v1.7c3.12,-1.657 3.698,0.118 2.336,1.25' +
+          '-1.201,0.998 -1.201,1.528 -1.204,2.19z'},
```
kchadha: should `},` here be on a separate line to match the other blocks like this?

core/flyout_base.js
```diff
@@ -289,7 +289,8 @@ Blockly.Flyout.prototype.init = function(targetWorkspace) {
       Blockly.bindEventWithChecks_(this.svgGroup_, 'wheel', this, this.wheel_));
   // Dragging the flyout up and down (or left and right).
   Array.prototype.push.apply(this.eventWrappers_,
-      Blockly.bindEventWithChecks_(this.svgGroup_, 'mousedown', this, this.onMouseDown_));
+      Blockly.bindEventWithChecks_(
+          this.svgBackground_, 'mousedown', this, this.onMouseDown_));
```
kchadha: Was the switch from `this.svgGroup_` to `this.svgBackground_` here intentional?

core/scrollbar.js
```diff
@@ -30,6 +30,7 @@ goog.provide('Blockly.ScrollbarPair');
 goog.require('goog.dom');
 goog.require('goog.events');
 
+
```
kchadha: Extra newline

core/warning.js
```diff
@@ -57,21 +57,27 @@ Blockly.Warning.prototype.collapseHidden = false;
...
+      },
+      group);
   // Dot of exclamation point.
   Blockly.utils.createSvgElement('rect',
-      {'class': 'blocklyIconSymbol',
-       'x': '7', 'y': '11', 'height': '2', 'width': '2'},
-       group);
+      {
+        'class': 'blocklyIconSymbol',
+        'x': '7', 'y': '11', 'height': '2', 'width': '2'
```
kchadha: Should these be on separate lines to match new style in other places?

kchadha (commented):
Looks mostly good. The comments are mostly further minor style update suggestions, and one possible bug.

.eslintrc
```diff
@@ -6,7 +6,16 @@
         "linebreak-style": ["error", "unix"],
         "max-len": ["error", 120, 4],
         "no-trailing-spaces": ["error", { "skipBlankLines": true }],
-        "no-unused-vars": ["error", {"args": "after-used", "varsIgnorePattern": "^_"}],
+        "no-unused-vars": [
+            "error",
+            {
+                "args": "after-used",
+                # Ignore vars starting with an underscore.
+                "varsIgnorePattern": "^_",
```
rachel-fenichel, line 14: @kchadha are you folks okay with this convention for ignorable argument names/variables, or do you prefer something else?  No strong preference from our side.  This is mostly for defining event handlers, where sometimes they don't use the incoming event but I want to be clear that it exists and is documented.

core/comment.js
rachel-fenichel: Yes.  Done.

core/flyout_base.js
rachel-fenichel: No.  Good catch.

Though I'm not sure why the two repos diverge on this line, and nothing in the history is really helpful.

core/warning.js
rachel-fenichel: Done.  I suspect that I didn't catch every use of createSvgElement in Blockly, so there may be some inconsistency that gets worked out over time.

This doesn't cause lint errors (as long as the indentation is correct) and we don't have a written preference for style anywhere.  I've been going with 
```
  Blockly.utils.createSvgElement('rect',
      {
        'class': 'blocklyIconSymbol',
        'x': '7',
        'y': '11',
        'height': '2',
        'width': '2'
      },
      group);
```
for anything that can't fit the object in one line, and 

```
    var tspanElement = Blockly.utils.createSvgElement('tspan',
        {'dy': '1em', 'x': Blockly.Bubble.BORDER_WIDTH}, paragraph);
```
if it does fit.

.eslintrc
kchadha: Not sure about this one since it could get confusing with other repos where we sometimes use that convention to mark things private.
@paulkaplan any thoughts on this?

kchadha (approved):
Looks good! Thanks for making these changes! <3

PR #1481, opened 2018-04-27
Store and get category length

ericrosenbaum:
### Resolves

Support for https://github.com/LLK/scratch-gui/issues/1638

### Proposed Changes

- When recording the scroll position of the top of each blocks category, also store the length of the category.
- Add a function to get the length of a category by name.

### Reason for Changes

The GUI will use the category length to prevent a problem where the scroll position jumps when switching sprites between a sprite and the stage, while in a category that changes length (such as motion).

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -544,7 +544,7 @@ Blockly.Flyout.prototype.recordCategoryScrollPositions_ = function() {
     }
   }
   var numCategories = this.categoryScrollPositions.length;
-  for (var i=0; i<numCategories - 1; i++) {
+  for (var i = 0; i < numCategories - 1; i++) {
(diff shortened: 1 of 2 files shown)


core/flyout_base.js
```diff
@@ -543,6 +543,14 @@ Blockly.Flyout.prototype.recordCategoryScrollPositions_ = function() {
       });
     }
   }
+  var numCategories = this.categoryScrollPositions.length;
+  for (var i=0; i<numCategories - 1; i++) {
```
rachel-fenichel: nit: spaces around = and <

rachel-fenichel (commented):
One nit then lg.

ericrosenbaum:
thanks @rachel-fenichel!

PR #1484, opened 2018-04-30
Add scroll blocks

towerofnix:
### Resolves

Towards LLK/scratch-vm#355 (adds scrolling-related blocks). Should be merged alongside LLK/scratch-vm#1099.

### Proposed Changes

Adds five new blocks: `motion_scroll_right`, `motion_scroll_up`, `motion_align_scene`, `motion_xscroll`, and `motion_yscroll`.

The five new blocks: "scroll right (10)", "scroll up (10)", "x scroll", "y scroll", "align scene (middle)". The "align scene" block's dropdown menu is open, showing its options: "bottom-left", "bottom-right", "middle", "top-left", and "top-right".

### Reason for Changes

Compatibility with Scratch 2.0 projects. These are hacked blocks; in total, [around 130 projects use scroll blocks](https://github.com/LLK/scratch-vm/issues/355#issuecomment-379418752).

### Test Coverage

No new tests, but tested manually.

### DIFF
--- a/blocks_vertical/motion.js
+++ b/blocks_vertical/motion.js
@@ -478,3 +478,110 @@ Blockly.Blocks['motion_direction'] = {
     });
   }
 };
+
+Blockly.Blocks['motion_scroll_right'] = {
+  /**
+   * Block to scroll the stage right. Does not actually do anything. This is
+   * an obsolete block that is implemented for compatibility with Scratch 2.0
+   * projects.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": Blockly.Msg.MOTION_SCROLLRIGHT,
+      "args0": [
+        {
+          "type": "input_value",
+          "name": "DISTANCE"
+        }
+      ],
+      "category": Blockly.Categories.motion,
+      "extensions": ["colours_motion", "shape_statement"]
+    });
+  }
+};
+
+Blockly.Blocks['motion_scroll_up'] = {
+  /**
+   * Block to scroll the stage up. Does not actually do anything. This is an
+   * obsolete block that is implemented for compatibility with Scratch 2.0
+   * projects.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": Blockly.Msg.MOTION_SCROLLUP,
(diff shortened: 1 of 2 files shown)


blocks_vertical/motion.js
```diff
@@ -478,3 +478,110 @@ Blockly.Blocks['motion_direction'] = {
...
+   * Block to report the stage's scroll position's X value. Does not actually
+   * do anything. This is an obsolete block that is implemented for
+   * compatibility with Scratch 2.0 projects.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": Blockly.Msg.MOTION_XSCROLL,
+      "category": Blockly.Categories.motion,
+      "checkboxInFlyout": true,
```
towerofnix: Does this make sense? This block never appears in the flyout/palette at all. Is it okay to omit `checkboxInFlyout` entirely?

towerofnix: Seeing as the counter blocks don't have `checkboxInFlyout`, it doesn't look to be required.

kchadha: @towerofnix, that's right, you don't need the `checkboxInFlyout` here (especially since this block doesn't appear in the flyout).

blocks_vertical/motion.js
```diff
@@ -478,3 +478,110 @@ Blockly.Blocks['motion_direction'] = {
...
+   * Block to report the stage's scroll position's Y value. Does not actually
+   * do anything. This is an obsolete block that is implemented for
+   * compatibility with Scratch 2.0 projects.
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": Blockly.Msg.MOTION_YSCROLL,
+      "category": Blockly.Categories.motion,
+      "checkboxInFlyout": true,
```
kchadha: Same here as above

kchadha (changes requested):
Minor change, please remove checkboxInFlyout property since these blocks aren't being used in the flyout.

towerofnix:
Right, thanks - I meant to remove those but forgot! They're gone now.

kchadha (approved):

PR #1488, opened 2018-05-01
Add micellaneous no-op obsolete blocks

towerofnix:
### Resolves

Towards LLK/scratch-vm#355 (adds "hide all sprites", "set stretch to", "change stretch by", and "user id"). Should be merged alongside LLK/scratch-vm#1102.

### Proposed Changes

Adds four new blocks: `looks_hideallsprites`, `looks_setstretchto`, `looks_changestretchby`, and `sensing_userid`.

The four new blocks: "hide all sprites", "set stretch to (10)", "change stretch by (10)", and "user id".

(Note -- this screenshot is a little out of date. It's now "set stretch to (10) %", which [matches](https://en.scratch-wiki.info/wiki/Set_Stretch_to_()_(block)) 1.4.)

### Reason for Changes

Compatibility with Scratch 2.0 projects. In total, [around 350 projects use the change/set stretch blocks](https://github.com/LLK/scratch-vm/issues/355#issuecomment-379418752). I don't have stats on how often "hide all sprites" and "user id" were used.

### Test Coverage

No new tests, but tested manually.


blocks_vertical/looks.js
```diff
@@ -281,6 +298,49 @@ Blockly.Blocks['looks_size'] = {
...
+      "category": Blockly.Categories.looks,
+      "extensions": ["colours_looks", "shape_statement"]
+    });
+  }
+};
+
+Blockly.Blocks['looks_setstretchto'] = {
+  /**
+   * Block to set stretch. Does not actually do anything. This is an obsolete
+   * block that is implemented for compatibility with Scratch 2.0 projects.
```
towerofnix: This is not exactly true :)

I say "Scratch 2.0 projects", but really it's very very old pre-1.4 projects that had the `setStretchTo` and `changeStretchBy` blocks. [They don't exist in the scratch-flash source code](https://github.com/LLK/scratch-flash/blob/develop/src/Specs.as), and there are [projects that use them from 2010](https://scratch.mit.edu/projects/1119715/). (The Scratch wiki [implies](https://en.scratch-wiki.info/wiki/Set_Stretch_to_()_(block)) they existed in 1.4 (though nonfunctional) and were removed altogether in 2.0.)

kchadha: Thanks for the comment and the research. =] Can you add a mention for 1.4 projects here then? Or feel free to change all of these comments to say something like `for compatibility with legacy blocks` instead.

kchadha (changes requested):
Minor change request in comments.

blocks_vertical/looks.js
towerofnix: Okay, there's a comment here explaining this now :)

kchadha (approved):
Thanks for updating the comment!

PR #1491, opened 2018-05-02
New script to pull from Blockly and do basic cleanup

rachel-fenichel:
### Resolves

Ergonomics

### Proposed Changes

This script pulls from blockly, runs cleanup.sh to handle basic problems, shows the status, and optionally aborts the merge.

If run without prompts it'll just do all the steps and not abort the merge.
If run with prompts it'll ask the user what they want to do at various points.  This gives the user a chance to look at intermediate status, and to abort if they don't like the state.

Usage:
`./pull_from_blockly.sh`
`./pull_from_blockly.sh with_prompts`

This has the same problem as cleanup.sh of needing to be run in the directory it lives in, I think.

It does not do any commits or create any new branches, so in its worst case all effects can be backed out with `git merge --abort`.

### Reason for Changes

Formalize parts of the merge process

### Test Coverage

Ran it and went through the various y/n options, as well as running it without prompts.

### DIFF
--- a/pull_from_blockly.sh
+++ b/pull_from_blockly.sh
@@ -1,0 +1,150 @@
+#!/bin/bash
+
+# Pull from Blockly into Scratch Blocks and do basic cleanup.
+# Rachel Fenichel (fenichel@google.com)
+
+BOLD='\e[1m'
+NOBOLD='\e[21m'
+
+# Formatting helper.
+empty_lines() { printf '\n\n'; }
+bold_echo() {
+  echo -e "${BOLD}$1${NOBOLD}"
+}
+
+stop_on_fail() {
+  # Fail if any command fails.
+  set -e
+  # Even if you're piping the output.
+  set -o pipefail
+}
+
+# Undo the effects of start_failing.
+continue_on_fail() {
+  set +e
+  set +o pipefail
+}
+
+# Prompt for y/n and put the result in $prompt_result
+# The first argument specifies the text to use in the prompt.
+# The second argument specifies which value to use if we're skipping prompts.
+prompt() {
+  if [ $with_prompts ]
+    then
+      if [ $2 = true ]
+        then
+          paren_text="(Y/n)"
+      else
+        paren_text="(y/N)"
+      fi
+      # Prompt the user and retry if they try any funny business.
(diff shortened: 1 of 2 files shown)


pull_from_blockly.sh
```diff
@@ -0,0 +1,150 @@
...
+
+
+# Ask the user for confirmation, then pull from Blockly's develop branch.
+# The default is to do the pull.
+pull_from_develop_fn() {
+  empty_lines
+  prompt "Do you want to pull from develop?" true
+  if [ $prompt_result = false ]
+    then
+      bold_echo "You don't want to pull from develop.  Why are you running this script?"
```
kchadha, line 63: 😄

pull_from_blockly.sh
```diff
@@ -0,0 +1,150 @@
...
+
+# Ask the user for confirmation, then possibly abort the merge.
+# The default is to *not* abort the merge.
+# Used to clean up the repo instead of leaving it in a bad state.
+prompt_for_merge_abort() {
+  empty_lines
+  prompt "Do you want to abort this merge?" false
+  if [ $prompt_result = false ]
+    then
+      bold_echo "Skipping merge abort."
```
kchadha: Should these be flipped? If `$prompt_result` is false above, that means the user said `N` or pressed enter (defaulting to N) right?

kchadha: This print out confused me for a second. Maybe say 'Continuing with merge' or `Merge was not aborted` instead?

pull_from_blockly.sh
```diff
@@ -0,0 +1,150 @@
...
+  else
+    bold_echo "Running git merge --abort"
+    git merge --abort
+    bold_echo "Current status"
+    sleep .5
+    git status
```
kchadha: You could use `display_status_fn` here

kchadha (approved):
Looks good! Also this script is awesome =]

Had some minor comments, but they're okay to ignore.

rachel-fenichel:
I made the changes you requested.  The phrasing on the merge-not-aborted method is still weird, but I think we'll deal.  I'll merge it when travis finishes.

PR #1493, opened 2018-05-02
More lint

rachel-fenichel:
### Resolves

🎶Dashing through the code, linting as I go...🎶

Further lines are left as an exercise for the singer.

### Proposed Changes

Lots of lint.

I'll probably return to inject.js to break createDom up into smaller pieces, which will also help with mergeability with Blockly--not all of the filter/glow stuff is shared.

### Reason for Changes

Lint and merge conflicts.

### DIFF
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -2,7 +2,7 @@
  * @license
  * Visual Blocks Editor
  *
- * Copyright 2011 Google Inc.
+ * Copyright 2017 Google Inc.
  * https://developers.google.com/blockly/
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -39,6 +39,7 @@ goog.require('goog.events');
 goog.require('goog.math.Rect');
 goog.require('goog.userAgent');
 
+
 /**
  * Class for a flyout.
  * @param {!Object} workspaceOptions Dictionary of options for the workspace.
@@ -51,7 +52,7 @@ Blockly.VerticalFlyout = function(workspaceOptions) {
 
   Blockly.VerticalFlyout.superClass_.constructor.call(this, workspaceOptions);
   /**
-   * Flyout should be laid out horizontally vs vertically.
+   * Flyout should be laid out vertically.
    * @type {boolean}
    * @private
    */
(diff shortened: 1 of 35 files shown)


rachel-fenichel:
This should cover all lint in core now.

Kenny2github:
:musical_note: 
Dashing through the code, on an input/output stream...
O'er the lines we go, linting fast as steam!
Errors fail a check, PR can't be merged...
What fun it is to write and see a linting fail submerged!

|: chorus :|
Linting code, linting code,
Linting to be merged,
Oh! What fun it is to ride
On a file with lint fails purged

A day or two ago, I thought I'd jump right in,
And soon I realized: That was total sin!
For linting was a mess, misfortune seemed their lot,
They merged a pull with failing checks and then they got upsot!

|: chorus :|
:musical_note:

core/flyout_vertical.js
```diff
@@ -51,7 +52,7 @@ Blockly.VerticalFlyout = function(workspaceOptions) {
 
   Blockly.VerticalFlyout.superClass_.constructor.call(this, workspaceOptions);
   /**
-   * Flyout should be laid out horizontally vs vertically.
+   * Flyout should be laid out vertically.
```
kchadha, line 55: Above comment is now misleading since `true` means horizontal... maybe change to `Whether the flyout should be laid out horizontally. Vertical by default.`

Or is this just a change to match upstream?

core/workspace_svg.js
```diff
@@ -425,7 +425,7 @@ Blockly.WorkspaceSvg.prototype.createDom = function(opt_backgroundClass) {
     bottom = this.addTrashcan_(bottom);
   }
   if (this.options.zoomOptions && this.options.zoomOptions.controls) {
-    bottom = this.addZoomControls_(bottom);
+    this.addZoomControls_(bottom);
```
kchadha, line 428: Does `bottom` not get used after this line?

kchadha (approved):
Changes look good. Had some minor questions in the comments.

core/flyout_vertical.js
rachel-fenichel: Yup, it's a change to match upstream.  I read it as "This subclass is always laid out vertically (so horizontalLayout is false)."

core/workspace_svg.js
rachel-fenichel: Correct.  If we were adding more controls below (above?  I don't remember) we could use the result of addZoomControls_, but in this case we're not going to.

PR #1496, opened 2018-05-03
Add new indent rule to .eslintrc and increase the minimum version

rachel-fenichel:
### Resolves

None
### Proposed Changes

Use the [new indent rules](https://eslint.org/docs/user-guide/migrating-to-4.0.0#indent-rewrite) added in eslint 4.  Require a newer version of eslint, since the rules are not compatible across the migration.

### Reason for Changes

Automate linting, consistency between blockly and scratch-blocks.

### Test Coverage

I ran eslint . and got no errors or warnings.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,28 @@
     "rules": {
         "curly": ["error", "multi-line"],
         "eol-last": ["error"],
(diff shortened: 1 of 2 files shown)


rschamp (approved):
It lints! LGTM!

PR #1502, opened 2018-05-07
Fixed lack of category identifiers

### LINKED ISSUE #1509: Some sound blocks do not have a category label
I noticed that a couple of blocks in the Sound category do not have a category label.  I don't know if this actually affects much (other than being unable to properly theme the blocks via custom css), but I though you should be aware.

**These blocks include:**
play sound (sound) until done
change [pitch] effect by (10)
set [pitch] effect to (100)
clear sound effects

I opened a pull request (#1502) that includes a quick fix for this issue.

ZenithRogue:
When I was changing CSS via a userscript, I noticed that a few blocks were not being affected.  Looking at the code I noticed that those specific blocks did not have the `"category": Blockly.Categories.sound,` property.  This simple change should fix the problem.  (There is one more block with this issue, and that is the "When backdrop switches to" block.  However, looking at it, it appears that this block does have a category definition, making that cause unknown)

Resolves #1509

### DIFF
--- a/blocks_vertical/sound.js
+++ b/blocks_vertical/sound.js
@@ -95,6 +95,7 @@ Blockly.Blocks['sound_playuntildone'] = {
           "name": "SOUND_MENU"
         }
       ],
+      "category": Blockly.Categories.sound,
       "extensions": ["colours_sounds", "shape_statement"]
     });
   }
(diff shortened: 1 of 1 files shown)


ZenithRogue:
Any reason this can't be merged?

Kenny2github:
The ST just haven't gotten around to looking at this pull. Please note, however, that you should (actually, *must*, but you've already opened it) open an issue for problems like this before sending in pulls - and the issue must be labelled "help wanted" before the ST will accept outside pulls for it.

thisandagain (approved):
🎉 Thanks.

PR #1503, opened 2018-05-07
chore(webpack): update to webpack 4

mzgoddard:
### Proposed Changes

- Update from webpack 1 to webpack 4
- Use full loader (`imports` becomes `imports-loader`) names in `shim/`
- Add `shim/gh-pages.js` entry (versions of webpack after 1 do not allow entry to be empty)

### Reason for Changes

Sibling of https://github.com/LLK/scratch-vm/pull/1104

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- "4"
+- "8"
 - "stable"
 sudo: required
 addons:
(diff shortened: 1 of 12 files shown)


mzgoddard:
Uh. Webpack dropped support for Node 4 a while ago. To update a modern webpack travis needs to build on a more recent node. At least Node 6 if I recall correctly.

Primarily this is because webpack updated its js syntax to ES2015 which Node 4 doesn't support all of.

mzgoddard:
The build failure is because travis is using Node 4 for scratch-blocks.

rschamp (approved):

PR #1510, opened 2018-05-09
Change local compiler script name

mzgoddard:
### Proposed Changes

Use a path to the local compiler's startup script under node_modules instead of assuming the developer will have node_modules/.bin in there executable PATH.

### Reason for Changes

This removes an assumption about how the local compiler runs so it should work for more developers with less (unexplained setup). The remaining bit for local builds would be need for a JDK. This is mentioned as an error by the startup script before build.py starts trying to use the remote compiler but could possibly be better explained.

### DIFF
--- a/build.py
+++ b/build.py
@@ -49,9 +49,9 @@ CLOSURE_LIBRARY = "closure-library"
 CLOSURE_COMPILER = REMOTE_COMPILER
 
 CLOSURE_DIR_NPM = "node_modules"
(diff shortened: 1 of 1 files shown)


rschamp (approved):
This works for me, but I can't say I actually understand which situation this change will fix.  So I think @rachel-fenichel should also give a 👍 before we merge this.

rachel-fenichel:
I also don't understand this.  Can you explain more?  

Does it solve or exacerbate #1523 ?

PR #1512, opened 2018-05-09
Fix the check to compile local or remote in do_compile

mzgoddard:
### Proposed Changes

Fix the check to compile local or remote in do_compile

### Reason for Changes

All the compilations since the local compile ability has actually been local compiles.

### DIFF
--- a/build.py
+++ b/build.py
@@ -322,7 +322,7 @@ class Gen_compressed(threading.Thread):
     self.do_compile(params, target_filename, filenames, remove)
 
   def do_compile(self, params, target_filename, filenames, remove):
-    do_compile = self.do_compile_remote if self.closure_env == REMOTE_COMPILER else self.do_compile_local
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -322,7 +322,7 @@ def gen_generator(self, language):
     self.do_compile(params, target_filename, filenames, remove)
 
   def do_compile(self, params, target_filename, filenames, remove):
-    do_compile = self.do_compile_remote if self.closure_env == REMOTE_COMPILER else self.do_compile_local
+    do_compile = self.do_compile_remote if self.closure_env["closure_compiler"] == REMOTE_COMPILER else self.do_compile_local
```
rachel-fenichel: Long line--prefer
```
    if self.closure_env["closure_compiler"] == REMOTE_COMPILER:
      do_compile = self.do_compile_remote
    else:
      do_compile = self.do_compile_local
```

PR #1513, opened 2018-05-09
Fix the uncompressed node_modules BLOCKLY_BOOT check

mzgoddard:
### Proposed Changes

Complete a check added to uncompressed for the local compile work

### Reason for Changes

Make sure the checked value is within the range that states the substring is inside the BLOCKLY_DIR


thisandagain:
@mzgoddard Can you check out why CI is failing?

cwillisf:
`build.py` will be removed soon

PR #1514, opened 2018-05-09
Split events.js into multiple files

rachel-fenichel:
### Resolves

Needs to happen to pull in workspace comments from blockly.

### Proposed Changes

Split events.js into multiple files, and have a separate provide for each event type

Add base classes for most event types.

Add requires where needed.

### Reason for Changes

I made these changes on the Blockly side so that adding events doesn't mean putting even more code into events.js.  The application here is that the scratch-blocks-specific events are in scratch_events.js.  A developer adding a new event type still needs to add some constants to events.js and add an entry in fromJson, but this is a step closer to not needing to do that.

This change also makes merging easier.

### Test Coverage

Current tests pass, and some have been updated to refer to the new event types.
The vertical playground works, and all event types show up in the console.


core/variable_map.js
```diff
@@ -26,7 +26,8 @@
 
 goog.provide('Blockly.VariableMap');
 
-goog.require('Blockly.VariableModel');
```
kchadha: This file still uses Blockly.VariableModel...

rachel-fenichel: Fixed.

core/events.js
```diff
@@ -347,879 +353,6 @@ Blockly.Events.fromJson = function(json, workspace) {
...
- * Class for a block drag event. Fired when block dragged into or out of
- * the blocks UI.
- * @param {Blockly.Block} block The moved block.  Null for a blank event.
- * @extends {Blockly.Events.Abstract}
- * @constructor
- */
-Blockly.Events.DragOutside = function(block) {
-  if (!block) {
-    return;  // Blank event to be populated by fromJson.
-  }
```
kchadha, line 888: I noticed that this check/early exit got removed in the refactor, is it no longer necessary?

core/events.js
```diff
@@ -347,879 +353,6 @@ Blockly.Events.fromJson = function(json, workspace) {
...
-};
-goog.inherits(Blockly.Events.DragOutside, Blockly.Events.Abstract);
-
-/**
- * Class for a block drag event.
- * @param {Blockly.Block} block The moved block.  Null for a blank event.
- * @extends {Blockly.Events.Abstract}
- * @constructor
- */
-Blockly.Events.BlockDragOutside = Blockly.Events.DragOutside;
```
kchadha, line 899: This extra layer got removed in the refactor, I assume that's okay?

core/ui_events.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {string} element One of 'selected', 'comment', 'mutator', etc.
+ * @param {*} oldValue Previous value of element.
+ * @param {*} newValue New value of element.
+ * @extends {Blockly.Events.Abstract}
+ * @constructor
+ */
+Blockly.Events.Ui = function(block, element, oldValue, newValue) {
+  Blockly.Events.Ui.superClass_.constructor.call(this);
+  this.blockId = block ? block.id : null;
+  this.workspaceId = block? block.workspace.id : null;
```
kchadha: should have space between `block` and `?`

rachel-fenichel: Fixed.

core/events.js
rachel-fenichel: Yes.  The extra layer exists for the block events because they used to be called by their short forms (e.g. Create instead of BlockCreate).  When we added more events (VarCreate, etc.) we renamed the block events to be clearer.  But we couldn't get rid of the old classes in case people were relying on them.

These new events don't have people who would be relying on them having short names, so no need for the indirection.

core/events.js
rachel-fenichel: Added back in, for this and EndBlockDrag.

kchadha (approved):
Looks good!

Issue #1429, opened 2018-04-06
Scroll to category by ID instead of name


ericrosenbaum:
Currently we use the name of a category to access its scroll position when scrolling to it. This happens when you click a blocks category menu button, or when the GUI triggers positioning or scrolling, such as when switching sprites or adding an extension.

Once we have the new blocks generation system implemented for all blocks categories, we should use the category ID instead of the name to store and access the category position. This will prevent bugs due to language changes and name mismatches. For example, currently, if the name of an extension displayed in the extension library is different from the name displayed in the category menu, the auto-scroll on extension load fails.

PR #1517, opened 2018-05-11
Scroll to category by id

ericrosenbaum:
### Resolves

Support for https://github.com/LLK/scratch-blocks/issues/1429
GUI blocks menu will be broken after this change until https://github.com/LLK/scratch-gui/pull/2041 is merged.
  
### Proposed Changes

- Add an id to each category in the default toolbox
- Store the category ids in the object recording category scroll positions, which required adding `getCategoryByIndex`
- Add several functions to access categories by id, to replace the ones that accessed categories by name

### Reason for Changes

This change is necessary because the category names will change due to localization, so we need to use stable ids. 

It also provides support for auto-scrolling to an extension category (on adding it) even if the extension name shown in the library does not match the name shown in the category menu.

### Question

I've left in the functions that access categories by name. Should we:

A) Leave them in (messy, but fewer things break)?
B) Leave them in but stubbed out, or with a deprecation warning?
C) Remove them entirely?

(If it's B or C I can do that in another PR)

### DIFF
--- a/core/toolbox.js
+++ b/core/toolbox.js
@@ -502,10 +502,10 @@ Blockly.Toolbox.prototype.scrollToCategoryById = function(id) {
 /**
  * Get a category by its index.
  * @param  {number} index The index of the category.
- * @return {Blockly.Toolbox.Category} the category.
+ * @return {Blockly.Toolbox.Category} the category, or null if there are no categories.
  */
 Blockly.Toolbox.prototype.getCategoryByIndex = function(index) {
-  if (!this.categoryMenu_.categories_) return;
+  if (!this.categoryMenu_.categories_) return null;
   return this.categoryMenu_.categories_[index];
 };
 
--- a/core/toolbox.js
+++ b/core/toolbox.js
@@ -345,11 +345,18 @@ Blockly.Toolbox.prototype.getSelectedCategoryName = function() {
 };
 
 /**
+ * @return {string} The id of the currently selected category.
+ */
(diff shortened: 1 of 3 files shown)


core/toolbox.js
```diff
@@ -344,12 +344,19 @@ Blockly.Toolbox.prototype.getSelectedCategoryName = function() {
   return this.selectedItem_.name_;
 };
 
+/**
+ * @return {string} The id of the currently selected category.
+ */
```
rachel-fenichel: For all new functions here: if they're intended to be public, mark that explicitly.  Otherwise mark them package or private.

core/toolbox.js
```diff
@@ -439,6 +483,32 @@ Blockly.Toolbox.prototype.scrollToCategoryByName = function(name) {
...
+  }
+};
+
+/**
+ * Get a category by its index.
+ * @param  {number} index The index of the category.
+ * @return {Blockly.Toolbox.Category} the category.
+ */
+Blockly.Toolbox.prototype.getCategoryByIndex = function(index) {
+  if (!this.categoryMenu_.categories_) return;
```
rachel-fenichel: If you're trying to return null, do that explicitly and also describe it in the return description.

rachel-fenichel:
If you think/know that no one else is using those functions, I recommend just removing them.  I'm fine with that being a separate PR.

rachel-fenichel (approved):

PR #1518, opened 2018-05-11
Fix remote builds with build.py

rachel-fenichel:
This has two changes to make building and using the uncompressed version of scratch blocks work, when doing a remote build:
- get rid of a check for `node_modules` that was unnecessary and always set `dir` wrong
- correctly check for the remote compiler

The second one is the same as #1512 with a long line broken apart.

I need this so that I can add workspace comments, because I need to be able to build and test them as I add new classes.

### DIFF
--- a/build.py
+++ b/build.py
@@ -134,11 +134,7 @@ window.BLOCKLY_BOOT = function() {
       alert('Error: Closure not found.  Read this:\\n' +
             'developers.google.com/blockly/guides/modify/web/closure');
     }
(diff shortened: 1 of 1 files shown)


kchadha:
Changes look okay to me. Any idea why the build is failing on travis?

kchadha (approved):
LG pending travis build fix

rachel-fenichel:
Oh no, tests pass locally but not on Travis :(

PR #1519, opened 2018-05-12
Feature/merge workspace comments

rachel-fenichel:
Pull in workspace comment code and dependencies from blockly.

I'll add comments in palces that changes are not obviously workspace comment related.

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -179,8 +179,11 @@ Blockly.BlockDragger.prototype.startBlockDrag = function(currentDragDeltaXY) {
   // surface.
   this.draggingBlock_.moveToDragSurface_();
 
-  if (this.workspace_.toolbox_) {
-    this.workspace_.toolbox_.addDeleteStyle();
+  var toolbox = this.workspace_.getToolbox();
+  if (toolbox) {
+    var style = this.draggingBlock_.isDeletable() ? 'blocklyToolboxDelete' :
+        'blocklyToolboxGrab';
+    toolbox.addStyle(style);
   }
 };
 
@@ -249,8 +252,11 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
   }
   this.workspace_.setResizesEnabled(true);
 
-  if (this.workspace_.toolbox_) {
-    this.workspace_.toolbox_.removeDeleteStyle();
+  var toolbox = this.workspace_.getToolbox();
+  if (toolbox) {
+    var style = this.draggingBlock_.isDeletable() ? 'blocklyToolboxDelete' :
+        'blocklyToolboxGrab';
+    toolbox.removeStyle(style);
   }
   Blockly.Events.setGroup(false);
 
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -239,8 +239,6 @@ Blockly.copy_ = function(toCopy) {
     var xml = toCopy.toXmlWithXY();
   } else {
     var xml = Blockly.Xml.blockToDom(toCopy);
-    // Copy only the selected block and internal blocks.
-    Blockly.Xml.deleteNext(xml);
     // Encode start position in XML.
     var xy = toCopy.getRelativeToSurfaceXY();
     xml.setAttribute('x', toCopy.RTL ? -xy.x : xy.x);
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -61,7 +61,6 @@ goog.require('Blockly.constants');
 goog.require('Blockly.inject');
 goog.require('Blockly.utils');
 goog.require('goog.color');
-goog.require('goog.userAgent');
 
 
 // Turn off debugging when compiled.
@@ -230,33 +229,41 @@ Blockly.onKeyDown_ = function(e) {
 };
 
 /**
- * Copy a block onto the local clipboard.
- * @param {!Blockly.Block} block Block to be copied.
+ * Copy a block or workspace comment onto the local clipboard.
+ * @param {!Blockly.Block | !Blockly.WorkspaceComment} toCopy Block or Workspace Comment
+ *    to be copied.
  * @private
  */
-Blockly.copy_ = function(block) {
-  var xmlBlock = Blockly.Xml.blockToDom(block);
-  // Encode start position in XML.
-  var xy = block.getRelativeToSurfaceXY();
-  xmlBlock.setAttribute('x', block.RTL ? -xy.x : xy.x);
-  xmlBlock.setAttribute('y', xy.y);
-  Blockly.clipboardXml_ = xmlBlock;
-  Blockly.clipboardSource_ = block.workspace;
+Blockly.copy_ = function(toCopy) {
+  if (toCopy.isComment) {
+    var xml = toCopy.toXmlWithXY();
+  } else {
+    var xml = Blockly.Xml.blockToDom(toCopy);
+    // Copy only the selected block and internal blocks.
+    Blockly.Xml.deleteNext(xml);
+    // Encode start position in XML.
+    var xy = toCopy.getRelativeToSurfaceXY();
+    xml.setAttribute('x', toCopy.RTL ? -xy.x : xy.x);
+    xml.setAttribute('y', xy.y);
+  }
+  Blockly.clipboardXml_ = xml;
+  Blockly.clipboardSource_ = toCopy.workspace;
 };
 
 /**
- * Duplicate this block and its children.
- * @param {!Blockly.Block} block Block to be copied.
+ * Duplicate this block and its children, or a workspace comment.
+ * @param {!Blockly.Block | !Blockly.WorkspaceComment} toDuplicate Block or
+ *     Workspace Comment to be copied.
  * @private
  */
-Blockly.duplicate_ = function(block) {
+Blockly.duplicate_ = function(toDuplicate) {
   // Save the clipboard.
   var clipboardXml = Blockly.clipboardXml_;
   var clipboardSource = Blockly.clipboardSource_;
 
   // Create a duplicate via a copy/paste operation.
-  Blockly.copy_(block);
-  block.workspace.paste(Blockly.clipboardXml_);
+  Blockly.copy_(toDuplicate);
+  toDuplicate.workspace.paste(Blockly.clipboardXml_);
 
   // Restore the clipboard.
   Blockly.clipboardXml_ = clipboardXml;
--- a/core/bubble.js
+++ b/core/bubble.js
@@ -275,47 +275,52 @@ Blockly.Bubble.prototype.createDom_ = function(content, hasResize) {
 };
 
 /**
+ * Return the root node of the bubble's SVG group.
+ * @return {Element} The root SVG node of the bubble's group.
+ */
+Blockly.Bubble.prototype.getSvgRoot = function() {
+  return this.bubbleGroup_;
+};
+
+/**
+ * Expose the block's ID on the bubble's top-level SVG group.
+ * @param {string} id ID of block.
+ */
+Blockly.Bubble.prototype.setSvgId = function(id) {
+  if (this.bubbleGroup_.dataset) {
+    this.bubbleGroup_.dataset.blockId = id;
+  }
+};
+
+/**
  * Handle a mouse-down on bubble's border.
  * @param {!Event} e Mouse down event.
  * @private
  */
 Blockly.Bubble.prototype.bubbleMouseDown_ = function(e) {
-  this.promote_();
-  Blockly.Bubble.unbindDragEvents_();
-  if (Blockly.utils.isRightButton(e)) {
-    // No right-click.
-    e.stopPropagation();
-    return;
-  } else if (Blockly.utils.isTargetInput(e)) {
-    // When focused on an HTML text input widget, don't trap any events.
-    return;
+  var gesture = this.workspace_.getGesture(e);
+  if (gesture) {
+    gesture.handleBubbleStart(e, this);
   }
-  // Left-click (or middle click)
-  this.workspace_.startDrag(e, new goog.math.Coordinate(
-      this.workspace_.RTL ? -this.relativeLeft_ : this.relativeLeft_,
-      this.relativeTop_));
-
-  Blockly.Bubble.onMouseUpWrapper_ = Blockly.bindEventWithChecks_(document,
-      'mouseup', this, Blockly.Bubble.bubbleMouseUp_);
-  Blockly.Bubble.onMouseMoveWrapper_ = Blockly.bindEventWithChecks_(document,
-      'mousemove', this, this.bubbleMouseMove_);
-  Blockly.hideChaff();
-  // This event has been handled.  No need to bubble up to the document.
-  e.stopPropagation();
 };
 
 /**
- * Drag this bubble to follow the mouse.
- * @param {!Event} e Mouse move event.
+ * Show the context menu for this bubble.
+ * @param {!Event} _e Mouse event.
  * @private
  */
-Blockly.Bubble.prototype.bubbleMouseMove_ = function(e) {
-  this.autoLayout_ = false;
-  var newXY = this.workspace_.moveDrag(e);
-  this.relativeLeft_ = this.workspace_.RTL ? -newXY.x : newXY.x;
-  this.relativeTop_ = newXY.y;
-  this.positionBubble_();
-  this.renderArrow_();
+Blockly.Bubble.prototype.showContextMenu_ = function(_e) {
+  // NOP on bubbles, but used by the bubble dragger to pass events to
+  // workspace comments.
+};
+
+/**
+ * Get whether this bubble is deletable or not.
+ * @return {boolean} True if deletable.
+ * @package
+ */
+Blockly.Bubble.prototype.isDeletable = function() {
+  return false;
 };
 
 /**
@@ -609,4 +614,51 @@ Blockly.Bubble.prototype.dispose = function() {
   this.workspace_ = null;
   this.content_ = null;
   this.shape_ = null;
+};
+
+/**
+ * Move this bubble during a drag, taking into account whether or not there is
+ * a drag surface.
+ * @param {?Blockly.BlockDragSurfaceSvg} dragSurface The surface that carries
+ *     rendered items during a drag, or null if no drag surface is in use.
+ * @param {!goog.math.Coordinate} newLoc The location to translate to, in
+ *     workspace coordinates.
+ * @package
+ */
+Blockly.Bubble.prototype.moveDuringDrag = function(dragSurface, newLoc) {
+  if (dragSurface) {
+    dragSurface.translateSurface(newLoc.x, newLoc.y);
+  } else {
+    this.moveTo(newLoc.x, newLoc.y);
+  }
+  if (this.workspace_.RTL) {
+    this.relativeLeft_ = this.anchorXY_.x - newLoc.x - this.width_;
+  } else {
+    this.relativeLeft_ = newLoc.x - this.anchorXY_.x;
+  }
+  this.relativeTop_ = newLoc.y - this.anchorXY_.y;
+  this.renderArrow_();
+};
+
+/**
+ * Return the coordinates of the top-left corner of this bubble's body relative
+ * to the drawing surface's origin (0,0), in workspace units.
+ * @return {!goog.math.Coordinate} Object with .x and .y properties.
+ */
+Blockly.Bubble.prototype.getRelativeToSurfaceXY = function() {
+  return new goog.math.Coordinate(
+      this.anchorXY_.x + this.relativeLeft_,
+      this.anchorXY_.y + this.relativeTop_);
+};
+
+/**
+ * Set whether auto-layout of this bubble is enabled.  The first time a bubble
+ * is shown it positions itself to not cover any blocks.  Once a user has
+ * dragged it to reposition, it renders where the user put it.
+ * @param {boolean} enable True if auto-layout should be enabled, false
+ *     otherwise.
+ * @package
+ */
+Blockly.Bubble.prototype.setAutoLayout = function(enable) {
+  this.autoLayout_ = enable;
 };
--- a/core/bubble_dragger.js
+++ b/core/bubble_dragger.js
@@ -1,0 +1,282 @@
+/**
+ * @license
+ * Visual Blocks Editor
+ *
+ * Copyright 2018 Google Inc.
+ * https://developers.google.com/blockly/
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Methods for dragging a bubble visually.
+ * @author fenichel@google.com (Rachel Fenichel)
+ */
+'use strict';
+
+goog.provide('Blockly.BubbleDragger');
+
+goog.require('Blockly.Bubble');
+goog.require('Blockly.Events.CommentMove');
+goog.require('Blockly.WorkspaceCommentSvg');
+
+goog.require('goog.math.Coordinate');
+goog.require('goog.asserts');
+
+
+/**
+ * Class for a bubble dragger.  It moves things on the bubble canvas around the
+ * workspace when they are being dragged by a mouse or touch.  These can be
+ * block comments, mutators, warnings, or workspace comments.
+ * @param {!Blockly.Bubble|!Blockly.WorkspaceCommentSvg} bubble The item on the
+ *     bubble canvas to drag.
+ * @param {!Blockly.WorkspaceSvg} workspace The workspace to drag on.
+ * @constructor
+ */
+Blockly.BubbleDragger = function(bubble, workspace) {
+  /**
+   * The item on the bubble canvas that is being dragged.
+   * @type {!Blockly.Bubble|!Blockly.WorkspaceCommentSvg}
+   * @private
+   */
+  this.draggingBubble_ = bubble;
+
+  /**
+   * The workspace on which the bubble is being dragged.
+   * @type {!Blockly.WorkspaceSvg}
+   * @private
+   */
+  this.workspace_ = workspace;
+
+  /**
+   * Which delete area the mouse pointer is over, if any.
+   * One of {@link Blockly.DELETE_AREA_TRASH},
+   * {@link Blockly.DELETE_AREA_TOOLBOX}, or {@link Blockly.DELETE_AREA_NONE}.
+   * @type {?number}
+   * @private
+   */
+  this.deleteArea_ = null;
+
+  /**
+   * Whether the bubble would be deleted if dropped immediately.
+   * @type {boolean}
+   * @private
+   */
+  this.wouldDeleteBubble_ = false;
+
+  /**
+   * The location of the top left corner of the dragging bubble's body at the
+   * beginning of the drag, in workspace coordinates.
+   * @type {!goog.math.Coordinate}
+   * @private
+   */
+  this.startXY_ = this.draggingBubble_.getRelativeToSurfaceXY();
+
+  /**
+   * The drag surface to move bubbles to during a drag, or null if none should
+   * be used.  Block dragging and bubble dragging use the same surface.
+   * @type {?Blockly.BlockDragSurfaceSvg}
+   * @private
+   */
+  this.dragSurface_ =
+      Blockly.utils.is3dSupported() && !!workspace.getBlockDragSurface() ?
+      workspace.getBlockDragSurface() : null;
+};
+
+/**
+ * Sever all links from this object.
+ * @package
+ */
+Blockly.BubbleDragger.prototype.dispose = function() {
+  this.draggingBubble_ = null;
+  this.workspace_ = null;
+  this.dragSurface_ = null;
+};
+
+/**
+ * Start dragging a bubble.  This includes moving it to the drag surface.
+ * @package
+ */
+Blockly.BubbleDragger.prototype.startBubbleDrag = function() {
+  if (!Blockly.Events.getGroup()) {
+    Blockly.Events.setGroup(true);
+  }
+
+  this.workspace_.setResizesEnabled(false);
+  this.draggingBubble_.setAutoLayout(false);
+  if (this.dragSurface_) {
+    this.moveToDragSurface_();
+  }
+
+  this.draggingBubble_.setDragging && this.draggingBubble_.setDragging(true);
+
+  var toolbox = this.workspace_.getToolbox();
+  if (toolbox) {
+    var style = this.draggingBubble_.isDeletable() ? 'blocklyToolboxDelete' :
+        'blocklyToolboxGrab';
+    toolbox.addStyle(style);
+  }
+};
+
+/**
+ * Execute a step of bubble dragging, based on the given event.  Update the
+ * display accordingly.
+ * @param {!Event} e The most recent move event.
+ * @param {!goog.math.Coordinate} currentDragDeltaXY How far the pointer has
+ *     moved from the position at the start of the drag, in pixel units.
+ * @package
+ */
+Blockly.BubbleDragger.prototype.dragBubble = function(e, currentDragDeltaXY) {
+  var delta = this.pixelsToWorkspaceUnits_(currentDragDeltaXY);
+  var newLoc = goog.math.Coordinate.sum(this.startXY_, delta);
+
+  this.draggingBubble_.moveDuringDrag(this.dragSurface_, newLoc);
+
+  if (this.draggingBubble_.isDeletable()) {
+    this.deleteArea_ =  this.workspace_.isDeleteArea(e);
+    this.updateCursorDuringBubbleDrag_();
+  }
+};
+
+/**
+ * Shut the trash can and, if necessary, delete the dragging bubble.
+ * Should be called at the end of a bubble drag.
+ * @return {boolean} whether the bubble was deleted.
+ * @private
+ */
+Blockly.BubbleDragger.prototype.maybeDeleteBubble_ = function() {
+  var trashcan = this.workspace_.trashcan;
+
+  if (this.wouldDeleteBubble_) {
+    if (trashcan) {
+      goog.Timer.callOnce(trashcan.close, 100, trashcan);
+    }
+    // Fire a move event, so we know where to go back to for an undo.
+    this.fireMoveEvent_();
+    this.draggingBubble_.dispose(false, true);
+  } else if (trashcan) {
+    // Make sure the trash can is closed.
+    trashcan.close();
+  }
+  return this.wouldDeleteBubble_;
+};
+
+/**
+ * Update the cursor (and possibly the trash can lid) to reflect whether the
+ * dragging bubble would be deleted if released immediately.
+ * @private
+ */
+Blockly.BubbleDragger.prototype.updateCursorDuringBubbleDrag_ = function() {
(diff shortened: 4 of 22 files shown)


core/block_dragger.js
```diff
@@ -179,8 +179,11 @@ Blockly.BlockDragger.prototype.startBlockDrag = function(currentDragDeltaXY) {
   // surface.
   this.draggingBlock_.moveToDragSurface_();
 
-  if (this.workspace_.toolbox_) {
-    this.workspace_.toolbox_.addDeleteStyle();
+  var toolbox = this.workspace_.getToolbox();
```
rachel-fenichel, line 182: We changed the API for toolbox style from `addDeleteStyle` to `addStyle(style)`.  We use the new version in the bubble dragger, so needed to delete the old version and use the new one here as well.

core/blockly.js
```diff
@@ -61,7 +61,6 @@ goog.require('Blockly.constants');
 goog.require('Blockly.inject');
 goog.require('Blockly.utils');
 goog.require('goog.color');
-goog.require('goog.userAgent');
```
rachel-fenichel, line 64: Not used anywhere in this file.

core/blockly.js
```diff
@@ -230,33 +229,41 @@ Blockly.onKeyDown_ = function(e) {
 };
 
 /**
- * Copy a block onto the local clipboard.
- * @param {!Blockly.Block} block Block to be copied.
+ * Copy a block or workspace comment onto the local clipboard.
```
rachel-fenichel, line 232: Copy now has to handle a workspace comment as well as a block.

core/bubble.js
```diff
@@ -274,48 +274,53 @@ Blockly.Bubble.prototype.createDom_ = function(content, hasResize) {
   return this.bubbleGroup_;
 };
 
+/**
```
rachel-fenichel, line 277: Some bubble functions had to change to make block comments and workspace comments both drag with the bubble dragger.

You can play with block comments by changing the injection struct in `vertical_playground.html` to have `comments: true`.

core/bubble_dragger.js
```diff
@@ -0,0 +1,282 @@
+/**
```
rachel-fenichel, line 1: The bubble dragger moves things that live on the bubble canvas instead of the block canvas.

Things on the bubble canvas are always in front of things on the block canvas.  This ensures that blocks don't look like they're inside bubbles, among other things.  Ask @rachel-fenichel if you want more details.

The bubble dragger has a lot of code that's similar to the block dragger.  I'd like to eventually combine them more, but structuring it this way makes it easier to experiment with bubble behaviour for things like clicks and drags.

core/css.js
```diff
@@ -434,6 +434,12 @@ Blockly.Css.CONTENT = [
     'cursor: url("<<<PATH>>>/handdelete.cur"), auto;',
   '}',
 
+  '.blocklyToolboxGrab {',
```
rachel-fenichel, line 437: New style added to distinguish between deletable and undeletable blocks over the toolbox.  Used in the block dragger.

CSS styles for the new comments are below.

core/workspace_svg.js
```diff
@@ -247,6 +249,14 @@ Blockly.WorkspaceSvg.prototype.useWorkspaceDragSurface_ = false;
  */
 Blockly.WorkspaceSvg.prototype.isDragSurfaceActive_ = false;
 
+/**
```
rachel-fenichel, line 252: Moved up a little bit to match where it is in blockly/get rid of a merge conflict.  I don't remember whether it's used in this change.

core/workspace_svg.js
```diff
@@ -961,6 +981,18 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
   if (this.currentGesture_) {
     this.currentGesture_.cancel();  // Dragging while pasting?  No.
   }
+  if (xmlBlock.tagName.toLowerCase() == 'comment') {
```
rachel-fenichel, line 984: Paste can now paste blocks and ws comments.

core/workspace_svg.js
```diff
@@ -1255,17 +1318,19 @@ Blockly.WorkspaceSvg.prototype.onMouseWheel_ = function(e) {
  */
 Blockly.WorkspaceSvg.prototype.getBlocksBoundingBox = function() {
   var topBlocks = this.getTopBlocks(false);
+  var topComments = this.getTopComments(false);
```
rachel-fenichel, line 1322: The bounding box now has more than blocks.

The name of this function may eventually need to change, but probably not because it's public.  It's really "get the bounding box of things in the workspace that impact the size of the workspace".

core/blockly.js
```diff
@@ -230,33 +229,41 @@ Blockly.onKeyDown_ = function(e) {
...
-  xmlBlock.setAttribute('y', xy.y);
-  Blockly.clipboardXml_ = xmlBlock;
-  Blockly.clipboardSource_ = block.workspace;
+Blockly.copy_ = function(toCopy) {
+  if (toCopy.isComment) {
+    var xml = toCopy.toXmlWithXY();
+  } else {
+    var xml = Blockly.Xml.blockToDom(toCopy);
+    // Copy only the selected block and internal blocks.
+    Blockly.Xml.deleteNext(xml);
```
kchadha: This is a change from the previous behavior of copying the entire stack starting at the copy point. Any particular reason for this change?

There also seems to be some strange things going on with positioning if you're copying different blocks from a single stack (see GIF below):

copy_positioning

core/contextmenu.js
```diff
@@ -404,4 +404,99 @@ Blockly.ContextMenu.wsExpandOption = function(hasCollapsedBlocks, topBlocks) {
...
+ * @package
+ */
+Blockly.ContextMenu.workspaceCommentOption = function(ws, e) {
+  // Helper function to create and position a comment correctly based on the
+  // location of the mouse event.
+  var addWsComment = function() {
+    var comment = new Blockly.WorkspaceCommentSvg(
+        ws, Blockly.Msg.WORKSPACE_COMMENT_DEFAULT_TEXT,
+        Blockly.WorkspaceCommentSvg.DEFAULT_SIZE,
+        Blockly.WorkspaceCommentSvg.DEFAULT_SIZE);
```
kchadha, line 460: While playing around with the vertical_playground, I noticed that the size of the workspace comment is very small compared to the size of the blocks. I'm not sure what size we're aiming for, but it looks like it's easy enough to change! =]

core/gesture.js
```diff
@@ -68,9 +69,17 @@ Blockly.Gesture = function(e, creatorWorkspace) {
    * How far the mouse has moved during this drag, in pixel units.
    * (0, 0) is at this.mouseDownXY_.
    * @type {goog.math.Coordinate}
```
kchadha, line 71: looks like this type is no longer correct..

core/gesture.js
```diff
@@ -605,10 +695,34 @@ Blockly.Gesture.prototype.handleBlockStart = function(e, block) {
...
 /* Begin functions defining what actions to take to execute clicks on each type
  * of target.  Any developer wanting to add behaviour on clicks should modify
  * only this code. */
 
+/**
+ * Execute a bubble click.
+ * @private
+ */
+Blockly.Gesture.prototype.doBubbleClick_ = function() {
+  // TODO (#1673): Consistent handling of single clicks.
```
kchadha: This comment should probably be updated to indicate that it's referencing a blockly issue

core/workspace.js
```diff
@@ -55,6 +56,16 @@ Blockly.Workspace = function(opt_options) {
    * @private
    */
   this.topBlocks_ = [];
+  /**
+   * @type {!Array.<!Blockly.WorkspaceComment>}
+   * @private
+   */
+  this.topComments_ = [];
```
kchadha, line 63: Is top comments really just all workspace comments? I assume that comments can't be nested/connected the way blocks can..

core/workspace.js
```diff
@@ -193,6 +204,61 @@ Blockly.Workspace.prototype.getTopBlocks = function(ordered) {
...
+ * @package
+ */
+Blockly.Workspace.prototype.addTopComment = function(comment) {
+  this.topComments_.push(comment);
+
+  // Note: If the comment database starts to hold block comments, this may need
+  // to move to a separate function.
+  if (this.commentDB_[comment.id]) {
+    console.warn('Overriding an existing comment on this workspace, with id "' +
+        comment.id + '"');
```
kchadha, line 219: Does this basically happen every time we change the (text) contents of an existing comment? Or is this covering some case I'm missing?

core/workspace.js
```diff
@@ -193,6 +204,61 @@ Blockly.Workspace.prototype.getTopBlocks = function(ordered) {
...
+/**
+ * Finds the top-level comments and returns them.  Comments are optionally sorted
+ * by position; top to bottom (with slight LTR or RTL bias).
+ * @param {boolean} ordered Sort the list if true.
+ * @return {!Array.<!Blockly.WorkspaceComment>} The top-level comment objects.
+ * @package
+ */
+Blockly.Workspace.prototype.getTopComments = function(ordered) {
+  // Copy the topComments_ list.
+  var comments = [].concat(this.topComments_);
```
kchadha, line 247: Is it preferred to use `[].concat(x)` instead of `x.slice(0)`?

core/workspace.js
```diff
@@ -193,6 +204,61 @@ Blockly.Workspace.prototype.getTopBlocks = function(ordered) {
...
+  if (ordered && comments.length > 1) {
+    var offset = Math.sin(goog.math.toRadians(Blockly.Workspace.SCAN_ANGLE));
+    if (this.RTL) {
+      offset *= -1;
+    }
+    comments.sort(function(a, b) {
```
kchadha, line 253: I noticed that `getAllBlocks` doesn't have any sort of sorting, what's the use case for it in comments?

core/workspace.js
```diff
@@ -217,6 +283,9 @@ Blockly.Workspace.prototype.clear = function() {
   while (this.topBlocks_.length) {
     this.topBlocks_[0].dispose();
   }
+  while (this.topComments_.length) {
+    this.topComments_[this.topComments_.length - 1].dispose();
```
kchadha, line 287: I noticed that blocks are disposed starting at the beginning of the list and comments are disposed starting at the end of the list. Are there pros and cons here?

core/workspace.js
rachel-fenichel: It's currently all workspace comments, yes.  Block comments *are* nested in a similar way to blocks.  If we find a way to merge block and workspace comments, this may be important.

core/blockly.js
rachel-fenichel: Copying blocks from a single stack/positioning: that looks like it's repeatedly bumping blocks to not be within the snap radius of other blocks, at a guess.  

Change from previous behaviour wrt copying the entire stack: I think that's something scratch and blockly do differently--I'll check, and make sure I'm not accidentally pulling in blockly behaviour.

core/workspace.js
rachel-fenichel: No, this is when you actually create a new workspace comment.  It's for tracking the comment object, not its contents. 
 And if we let comments move from blocks to workspace and vice versa, this will be called when a block comment becomes a workspace comment.

core/workspace_comment.js
```diff
@@ -0,0 +1,370 @@
...
+goog.require('goog.math.Coordinate');
+
+
+/**
+ * Class for a workspace comment.
+ * @param {!Blockly.Workspace} workspace The block's workspace.
+ * @param {string} content The content of this workspace comment.
+ * @param {number} height Height of the comment.
+ * @param {number} width Width of the comment.
+ * @param {string=} opt_id Optional ID.  Use this ID if provided, otherwise
```
kchadha, line 43: The doc for this param should probably include the fact that if the given ID collides with a comment already on the workspace, a new one will be generated (as that might not be the expectation).

core/workspace_comment.js
```diff
@@ -0,0 +1,370 @@
...
+   */
+  this.content_ = content;
+
+  /**
+   * @package
+   * @type {boolean}
+   */
+  this.isComment = true;
+
+  Blockly.WorkspaceComment.fireCreateEvent(this);
```
kchadha, line 112: Playing around with comments in the vertical_playground, I think it would be useful to have the comment create event include the comment's contents in a top level property (instead of having to find it in the xml).

kchadha: Should this also be in a `Blockly.Events.isEnabled()` check like in `dispose` below?

core/workspace_comment.js
```diff
@@ -0,0 +1,370 @@
...
+ * Move a comment by a relative offset.
+ * @param {number} dx Horizontal offset, in workspace units.
+ * @param {number} dy Vertical offset, in workspace units.
+ * @package
+ */
+Blockly.WorkspaceComment.prototype.moveBy = function(dx, dy) {
+  var event = new Blockly.Events.CommentMove(this);
+  this.xy_.translate(dx, dy);
+  event.recordNew();
+  Blockly.Events.fire(event);
```
kchadha, line 193: Should there be an events isEnabled check here?

core/workspace_comment.js
```diff
@@ -0,0 +1,370 @@
...
+  element.setAttribute('x', Math.round(this.xy_.x));
+  element.setAttribute('y', Math.round(this.xy_.y));
+  element.setAttribute('h', this.height_);
+  element.setAttribute('w', this.width_);
+  return element;
+};
+
+/**
+ * Encode a comment subtree as XML, but don't serialize the XY coordinates.
+ * This method avoids some expensive metrics-related calls that are made in
```
kchadha: Should the comments here also indicate that width and height are also not encoded here? And maybe another indication that they are encoded in the comments for `toXMLWithXY`

kchadha: Is Math.round a particularly expensive call?

core/workspace_comment.js
kchadha: Actually I'm not sure this will matter much because we'll probably use the xml for the x, y, width, and height too

core/workspace_comment.js
```diff
@@ -0,0 +1,370 @@
...
+ */
+Blockly.WorkspaceComment.fromXml = function(xmlComment, workspace) {
+  var info = Blockly.WorkspaceComment.parseAttributes(xmlComment);
+
+  var comment = new Blockly.WorkspaceComment(
+      workspace, info.content, info.h, info.w, info.id);
+
+  var commentX = parseInt(xmlComment.getAttribute('x'), 10);
+  var commentY = parseInt(xmlComment.getAttribute('y'), 10);
+  if (!isNaN(commentX) && !isNaN(commentY)) {
```
kchadha: Would it make sense to translate in only one direction if only one is provided? Instead of not using either direction if both aren't provided...

core/workspace_comment.js
```diff
@@ -0,0 +1,370 @@
...
+ * @package
+ */
+Blockly.WorkspaceComment.fromXml = function(xmlComment, workspace) {
+  var info = Blockly.WorkspaceComment.parseAttributes(xmlComment);
+
+  var comment = new Blockly.WorkspaceComment(
+      workspace, info.content, info.h, info.w, info.id);
+
+  var commentX = parseInt(xmlComment.getAttribute('x'), 10);
+  var commentY = parseInt(xmlComment.getAttribute('y'), 10);
```
kchadha: Looks like you're calling `parseInt` on these attributes twice (it should already be stored in `info` as part of `parseAttributes` below.

core/workspace_comment_render_svg.js
```diff
@@ -0,0 +1,465 @@
...
+      },
+      this.deleteGroup_);
+};
+
+/**
+ * Handle a mouse-down on comment's resize corner.
+ * @param {!Event} e Mouse down event.
+ * @private
+ */
+Blockly.WorkspaceCommentSvg.prototype.resizeMouseDown_ = function(e) {
```
kchadha, line 256: I noticed that resizing a comment doesn't seem to emit an event. I think this might be necessary for the vm to keep up-to-date information about the comment's position/shape (esp. for saving and loading a project).

core/workspace_comment_render_svg.js
```diff
@@ -0,0 +1,465 @@
...
+  //this.promote_();
```
kchadha: cruft?

core/workspace.js
rachel-fenichel: Actually it does have sorting in blockly.  That just hasn't been merged into scratch blocks yet.  The use case is to be able to sort the comments as they go into generated code.

core/workspace.js
rachel-fenichel: Doing it from the end should avoid a bunch of array shuffling.

core/workspace_comment_svg.js
```diff
@@ -0,0 +1,591 @@
...
+/**
+ * Dispose of this comment.
+ * @package
+ */
+Blockly.WorkspaceCommentSvg.prototype.dispose = function() {
+  if (!this.workspace) {
+    // The comment has already been deleted.
+    return;
+  }
+  // If this comment is being dragged, unlink the mouse events.
```
kchadha: should this say "deleted" instead of "dragged"?

core/workspace_comment_svg.js
```diff
@@ -0,0 +1,591 @@
...
+  if (this.workspace.options.readOnly) {
+    return;
+  }
+  // Save the current workspace comment in a variable for use in closures.
+  var comment = this;
+  var menuOptions = [];
+
+  if (this.isDeletable() && this.isMovable()) {
+    menuOptions.push(Blockly.ContextMenu.commentDuplicateOption(comment));
+    menuOptions.push(Blockly.ContextMenu.commentDeleteOption(comment));
```
kchadha, line 177: Should these be separated out into different checks? E.g. can a comment be moveable but not deletable or vice-versa?

core/workspace_comment_svg.js
```diff
@@ -0,0 +1,591 @@
...
+ * @package
+ */
+Blockly.WorkspaceCommentSvg.prototype.select = function() {
+  if (Blockly.selected == this) {
+    return;
+  }
+  var oldId = null;
+  if (Blockly.selected) {
+    oldId = Blockly.selected.id;
+    // Unselect any previously selected block.
```
kchadha: block or comment

core/workspace_comment_svg.js
```diff
@@ -0,0 +1,591 @@
...
+Blockly.WorkspaceCommentSvg.prototype.unselect = function() {
+  if (Blockly.selected != this) {
+    return;
+  }
+  var event = new Blockly.Events.Ui(null, 'selected', this.id, null);
+  event.workspaceId = this.workspace.id;
+  Blockly.Events.fire(event);
+  Blockly.selected = null;
+  this.removeSelect();
+  this.blurFocus();
```
kchadha: It looks like this is getting called twice because it's also called in `this.removeSelect()`

core/workspace_comment_svg.js
```diff
@@ -0,0 +1,591 @@
...
+ * @param {number} dx Horizontal offset, in workspace units.
+ * @param {number} dy Vertical offset, in workspace units.
+ * @package
+ */
+Blockly.WorkspaceCommentSvg.prototype.moveBy = function(dx, dy) {
+  var event = new Blockly.Events.CommentMove(this);
+  // TODO: Do I need to look up the relative to surface XY position here?
+  var xy = this.getRelativeToSurfaceXY();
+  this.translate(xy.x + dx, xy.y + dy);
+  this.xy_ = new goog.math.Coordinate(xy.x + dx, xy.y + dy);
```
kchadha: why not use the translate function of `goog.math.Coordinate`?

kchadha: also, `this.xy_` is getting set (to the same thing) twice because it's also getting set in translate...

kchadha: Can we reuse the code in the WorkspaceComment class?

core/workspace_comment_svg.js
```diff
@@ -0,0 +1,591 @@
...
+ * Decode an XML comment tag and create a rendered comment on the workspace.
+ * @param {!Element} xmlComment XML comment element.
+ * @param {!Blockly.Workspace} workspace The workspace.
+ * @param {number=} opt_wsWidth The width of the workspace, which is used to
+ *     position comments correctly in RTL.
+ * @return {!Blockly.WorkspaceCommentSvg} The created workspace comment.
+ * @package
+ */
+Blockly.WorkspaceCommentSvg.fromXml = function(xmlComment, workspace,
+    opt_wsWidth) {
```
kchadha, line 540: Does it make sense to call the superclass fromXml first and do the rendering specific stuff after?

core/workspace_comment_svg.js
```diff
@@ -0,0 +1,591 @@
...
+  return comment;
+};
+
+/**
+ * Encode a comment subtree as XML with XY coordinates.
+ * @param {boolean=} opt_noId True if the encoder should skip the comment id.
+ * @return {!Element} Tree of XML elements.
+ * @package
+ */
+Blockly.WorkspaceCommentSvg.prototype.toXmlWithXY = function(opt_noId) {
```
kchadha, line 575: same question here about calling superclass version of this function first (or maybe even just in the LTR case)?

core/workspace_svg.js
kchadha: Hmmm does this need to be enabled in the vertical_playground? I had trouble getting comment copy-paste working.

core/workspace_comment_svg.js
rachel-fenichel: That's happening on line 583.

core/workspace_svg.js
kchadha: The only way I got it working was to click on a ws comment, and then click out of the ws completely (so that it stays selected, but the text area isn't selected), and then copy-paste

core/workspace_svg.js
```diff
@@ -1018,6 +1050,37 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
...
+    var comment = Blockly.WorkspaceCommentSvg.fromXml(xmlComment, this);
+    // Move the duplicate to original position.
+    var commentX = parseInt(xmlComment.getAttribute('x'), 10);
+    var commentY = parseInt(xmlComment.getAttribute('y'), 10);
+    if (!isNaN(commentX) && !isNaN(commentY)) {
+      if (this.RTL) {
+        commentX = -commentX;
+      }
+      // Offset workspace comment.
+      // TODO: #1719 properly offset comment such that it's not interfereing with any blocks
```
kchadha: reference blockly issue

core/workspace_svg.js
```diff
@@ -1018,6 +1050,37 @@ Blockly.WorkspaceSvg.prototype.paste = function(xmlBlock) {
...
+      commentX += 50;
+      commentY += 50;
+      comment.moveBy(commentX, commentY);
+    }
+  } finally {
+    Blockly.Events.enable();
+  }
+  if (Blockly.Events.isEnabled()) {
+    // TODO: Fire a Workspace Comment Create event
```
kchadha: Why is this still a todo?

core/workspace_svg.js
```diff
@@ -1352,6 +1417,11 @@ Blockly.WorkspaceSvg.prototype.showContextMenu_ = function(e) {
         topBlocks));
   }
 
+  // Option to add a workspace comment.
+  if (this.options.comments) {
```
kchadha, line 1422: 👍

core/ws_comment_events.js
```diff
@@ -0,0 +1,410 @@
...
+
+/**
+ * Type of this event.
+ * @type {string}
+ */
+Blockly.Events.CommentDelete.prototype.type = Blockly.Events.COMMENT_DELETE;
+
+/**
+ * Encode the event as JSON.
+ * TODO (#1266): "Full" and "minimal" serialization.
```
kchadha: Blockly issue (I assume this one actually conflicts with a scratch-blocks thing rn)

core/ws_comment_events.js
```diff
@@ -0,0 +1,410 @@
...
+ */
+Blockly.Events.CommentDelete.prototype.fromJson = function(json) {
+  Blockly.Events.CommentDelete.superClass_.fromJson.call(this, json);
+};
+
+/**
+ * Run a creation event.
+ * @param {boolean} forward True if run forward, false if run backward (undo).
+ */
+Blockly.Events.CommentDelete.prototype.run = function(forward) {
```
kchadha, line 280: Does it make sense to call CommentCreate.run with the opposite of `forward` as an arg? Or is that confusing?

core/ws_comment_events.js
```diff
@@ -0,0 +1,410 @@
...
+  var workspace = this.getEventWorkspace_();
+  if (forward) {
+    var comment = workspace.getCommentById(this.commentId);
+    if (comment) {
+      comment.dispose(false, false);
+    } else {
+      // Only complain about root-level block.
+      console.warn("Can't uncreate non-existent comment: " + this.commentId);
```
kchadha: 'undelete'?

kchadha: JK -- delete?

core/ws_comment_events.js
```diff
@@ -0,0 +1,410 @@
...
+/**
+ * Class for a comment move event.  Created before the move.
+ * @param {Blockly.WorkspaceComment} comment The comment that is being moved.
+ *     Null for a blank event.
+ * @extends {Blockly.Events.CommentBase}
+ * @constructor
+ */
+Blockly.Events.CommentMove = function(comment) {
+  if (!comment) {
+    return;  // Blank event to be populated by fromJson.
```
kchadha, line 306: to be populated by recordNew?

kchadha: actually maybe nvm...

core/ws_comment_events.js
```diff
@@ -0,0 +1,410 @@
...
+ * @param {!goog.math.Coordinate} xy The location before the move, in workspace
+ *     coordinates.
+ */
+Blockly.Events.CommentMove.prototype.setOldCoordinate = function(xy) {
+  this.oldCoordinate_ = xy;
+};
+
+/**
+ * Encode the event as JSON.
+ * TODO (#1266): "Full" and "minimal" serialization.
```
kchadha: blockly issue

kchadha (commented):
Lots of comments throughout the code (I think most of them are minor).

I'm going to attempt to summarize some notes from testing in the vertical_playground here:

### Workspace Comments: 
#### Functionality/UX Notes
- comments render really small compared to block size
- resizing an unselected comment selects text on all comments
- Duplicating a comment should probably attach the comment svg to the mouse like duplicating a block does (so that the user can decide where to place it, instead of the offset positioning used for copy-paste)
- Duplication/copy-pasting a duplicated/copy-pasted comment seems to place the new comment really far away from the original (gets further the more you duplicate duplicated things)
- resizing a comment should fire an event
- (question for the future) should the comment text area be scrollable? Is this a can of worms? (looks like 2.0 comments are completely un-scrollable, not even using the cursor + arrow keys in the text)
- default text in comment should probably get selected when the comment is created for easy editing


#### Rendering Notes (probably all to be saved for the near future, but figured I'd write them here)
- there should probably be some (more) padding around the text in the comments
- text can get hidden behind resize icon
- resize icon renders a little outside the ws comment box (when comment is unselected) / highlight border (when comment is selected)
- min height and width make it so that the top part of the next line of text is just barely visible (probably shouldn't be visible)
image
- I can't tell for sure, but it seems the drag shadow is proportionally large compared to the drag shadow for blocks...


### Block Comments:
- (?) icon is in a weird place on the block
- for some reason bubble comment is being rendered without the bubble (no bubble border or link tying the bubble comment to the ? icon like in Blockly)
- should the block comment also have a context menu?
- is it possible for the comment bubble to open when it gets created (instead of just the ? icon popping up on the block)
- It looks like the comment icon has a color overlay (depending on the block color):
image
- Comment positioning is not necessarily remembered when comment icon toggle is used to show/hide the comment

rachel-fenichel:
### Functionality/UX Notes
> comments render really small compared to block size

No specific bug, but I know we'll be doing another pass on tweaking appearance.

> resizing an unselected comment selects text on all comments

https://github.com/google/blockly/issues/1860

> Duplicating a comment should probably attach the comment svg to the mouse like duplicating a block does (so that the user can decide where to place it, instead of the offset positioning used for copy-paste)

https://github.com/LLK/scratch-blocks/issues/1526

> Duplication/copy-pasting a duplicated/copy-pasted comment seems to place the new comment really far away from the original (gets further the more you duplicate duplicated things)

https://github.com/google/blockly/issues/1861

> resizing a comment should fire an event

https://github.com/google/blockly/issues/1862

> (question for the future) should the comment text area be scrollable? Is this a can of worms? (looks like 2.0 comments are completely un-scrollable, not even using the cursor + arrow keys in the text)

Will leave for later discussion.

> default text in comment should probably get selected when the comment is created for easy editing

https://github.com/google/blockly/issues/1863


### Rendering Notes 
> there should probably be some (more) padding around the text in the comments

No specific bug, but I know we'll be doing another pass on tweaking appearance.

> text can get hidden behind resize icon

https://github.com/google/blockly/issues/1864

> resize icon renders a little outside the ws comment box (when comment is unselected) / highlight border (when comment is selected)

https://github.com/google/blockly/issues/1865

> min height and width make it so that the top part of the next line of text is just barely visible (probably shouldn't be visible)

Probably depends on what font you're using, at a minimum.  I'm not sure we need to solve this for the general case.

> I can't tell for sure, but it seems the drag shadow is proportionally large compared to the drag shadow for blocks...

https://github.com/LLK/scratch-blocks/issues/1527


### Block Comments:
> (?) icon is in a weird place on the block

https://github.com/LLK/scratch-blocks/issues/1528

> for some reason bubble comment is being rendered without the bubble (no bubble border or link tying the bubble comment to the ? icon like in Blockly)

https://github.com/LLK/scratch-blocks/issues/1529

> should the block comment also have a context menu?

https://github.com/LLK/scratch-blocks/issues/1530

> is it possible for the comment bubble to open when it gets created (instead of just the ? icon popping up on the block)

https://github.com/LLK/scratch-blocks/issues/1531

> It looks like the comment icon has a color overlay (depending on the block color):

Leaving for later discussion

> Comment positioning is not necessarily remembered when comment icon toggle is used to show/hide the comment

https://github.com/LLK/scratch-blocks/issues/1532

core/workspace_svg.js
rachel-fenichel: https://github.com/google/blockly/issues/1867

core/blockly.js
rachel-fenichel: Removed the `deleteNext` call.

core/gesture.js
rachel-fenichel: Type was correct but code was wrong.

rachel-fenichel: Also fixing upstream: https://github.com/google/blockly/pull/1868

core/gesture.js
rachel-fenichel: Done.

core/workspace_comment.js
rachel-fenichel: Yeah I think there's a question of how much data should be duplicated between the JSON and the XML.  I don't have answers but I don't think we want to duplicate all of it.

core/workspace_comment.js
rachel-fenichel: Math.round isn't especially expensive.  I think it had been calling getMetrics before.  And the version of it for the rendered comments is more expensive.

rachel-fenichel: Update comment.

core/workspace_comment.js
rachel-fenichel: Good catch.

core/workspace_comment.js
rachel-fenichel: If one of them is missing I think it's a serialization problem, and we shouldn't do much extra work to cover it up.

core/workspace_comment_render_svg.js
rachel-fenichel: Bug filed.

core/workspace_comment_render_svg.js
rachel-fenichel: Deleted.

core/workspace_comment_svg.js
rachel-fenichel: Done.

core/workspace_comment_svg.js
rachel-fenichel: If it's not deletable we don't let you duplicate it (because then can you delete that one?) or delete it.

If it's not movable I guess maybe it's okay to allow duplication, but then can you move that comment?  Similarly, probably okay to allow deletion.  

This matches the logic for blocks in both blockly and scratch blocks.  If we want to change it, let's make that a separate issue and handle it for all of the cases.

core/workspace_comment_svg.js
rachel-fenichel: Done.

core/workspace_comment_svg.js
rachel-fenichel: Removed.

core/workspace_comment_svg.js
rachel-fenichel: Removed extra setting of `this.xy_`.

I'd like to keep the updates of `this.xy_` and the actual transform attribute very tightly coupled.

core/workspace_comment_svg.js
rachel-fenichel: No, because the moveBy can depend on the size of the workspace itself if it's rendered.

The model we have for building blocks (using the headless constructor and then adding stuff if it's rendered) has led to a lot of issues with initialization and ordering, even though it looks like a tidier way to do it.

I'm avoiding that model here, and ideally will eventually fix the block initialization as well.

core/workspace_svg.js
rachel-fenichel: Done

core/workspace_svg.js
rachel-fenichel: Done.

core/workspace_svg.js
rachel-fenichel: Note that this is enabled on scratch blocks but not blockly.

core/ws_comment_events.js
rachel-fenichel: Done.

core/ws_comment_events.js
rachel-fenichel: I think that's a good idea.  I made an issue for it rather than doing it right now. 
 https://github.com/google/blockly/issues/1871

core/ws_comment_events.js
rachel-fenichel: Done.

core/ws_comment_events.js
rachel-fenichel: Yeah, if there's no comment it'll be populated by `fromJson`.  If there's a comment we record the old coordinate but still need to call `recordNew` later.

core/ws_comment_events.js
rachel-fenichel: Done.

core/workspace.js
rachel-fenichel: Quick googling says it doesn't really matter.  Might just be a style thing from pxt-blockly.

core/workspace_comment.js
rachel-fenichel: Doesn't need one--it checks inside `fire`.  Normally we use that check to avoid constructing the event itself if it's unnecessary.  No harm is done by avoiding it.

core/workspace_comment.js
rachel-fenichel: Done, here and in block.js + block_svg.js for good measure.

core/workspace.js
kchadha: Mostly I meant to ask what's the use case for comment IDs colliding?

core/workspace.js
kchadha: yeah, that was my assumption. Should blocks be switched to doing it from the end as well?

core/workspace_comment.js
kchadha: SG, we can always revisit this later if need be.

tests/vertical_playground.html
```diff
@@ -62,7 +62,7 @@
 
         // Create main workspace.
         workspace = Blockly.inject('blocklyDiv', {
-          comments: false,
+          comments: true,
```
kchadha, line 65: 👍

kchadha (approved):
Looks good to go! Thanks for making the changes and for triaging the comments/filing all the bugs. =D

PR #1521, opened 2018-05-12
Use localeCompare for variable and procedure name comparison

### LINKED ISSUE #1504: Variable/broadcast menus sort "name10" before "name2"
### Expected Behavior

They should sort "10" after "2", for example "name1", "name2", "name10".

### Actual Behavior

They sort "10" before "2", for example "name1", "name10", "name 2".

A variables drop-down showing the above A broadcast drop-down

### Steps to Reproduce

Create three variables or broadcasts (or lists) and observe how they're sorted in menus.

Firefox 61.0b2, Debian (testing).

---

I also noticed the block palette has the same issue, for both variables and custom blocks:

In the block palette: "variable1", "variable10", "variable2"; "block1", "block10", "block2"

joshyrobot:
Fixes #1504 by using `localeCompare` with the numeric option enabled.

### DIFF
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -108,7 +108,7 @@ Blockly.Procedures.sortProcedureMutations_ = function(mutations) {
     var procCodeA = a.getAttribute('proccode');
     var procCodeB = b.getAttribute('proccode');
 
-    return Blockly.compareStrings(procCodeA, procCodeB);
+    return Blockly.scratchBlocksUtils.compareStrings(procCodeA, procCodeB);
   });
 
   return newMutations;
@@ -123,7 +123,7 @@ Blockly.Procedures.sortProcedureMutations_ = function(mutations) {
  * @private
  */
 Blockly.Procedures.procTupleComparator_ = function(ta, tb) {
-  return Blockly.compareStrings(ta[0], tb[0]);
+  return Blockly.scratchBlocksUtils.compareStrings(ta[0], tb[0]);
 };
 
 /**
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -108,13 +108,10 @@ Blockly.Procedures.sortProcedureMutations_ = function(mutations) {
     var procCodeA = a.getAttribute('proccode');
     var procCodeB = b.getAttribute('proccode');
 
-    if (procCodeA < procCodeB) {
-      return -1;
-    } else if (procCodeA > procCodeB) {
-      return 1;
-    } else {
-      return 0;
-    }
+    return procCodeA.localeCompare(procCodeB, undefined, {
+      sensitivity: 'base',
+      numeric: true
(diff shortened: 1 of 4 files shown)


thisandagain:
Thanks @joshyrobot! We'll take a look.

core/procedures.js
```diff
@@ -108,13 +108,10 @@ Blockly.Procedures.sortProcedureMutations_ = function(mutations) {
...
     var procCodeB = b.getAttribute('proccode');
 
-    if (procCodeA < procCodeB) {
-      return -1;
-    } else if (procCodeA > procCodeB) {
-      return 1;
-    } else {
-      return 0;
-    }
+    return procCodeA.localeCompare(procCodeB, undefined, {
```
kchadha: @joshyrobot, I think you can just use localeCompare with only the first argument and leave the other arguments out (here and all the other lines changed). It seems to do the right thing, and then we don't have to worry about these extra options.

kchadha (changes requested):
@joshyrobot, I requested a minor cleanup change in the comments.

Thanks!

rachel-fenichel:
That function looks like it belongs in either utils or scratchBlocksUtils.  I would prefer scratchBlocksUtils for now, and we can move it if it moves to Blockly.

(Sorry for the drive-by review)

core/scratch_blocks_utils.js
```diff
@@ -101,3 +101,16 @@ Blockly.scratchBlocksUtils.isShadowArgumentReporter = function(block) {
...
 };
+
+/**
+ * Compare strings with natural number sorting.
+ * @param {string} str1 First input.
+ * @param {string} str2 Second input.
+ * @return {number} -1, 0, or 1 to signify greater than, equality, or less than.
+ */
+Blockly.scratchBlocksUtils.compareStrings = function(str1, str2) {
+  return str1.localeCompare(str2, undefined, {
```
kchadha: I noticed that passing in `[]` for the locales argument instead of 'undefined' seems to have the same effect. I think that's preferred over passing in `undefined` as a value, and unfortunately passing in `null` seems to throw an error.

kchadha:
@joshyrobot Thanks for the catch! I realized I forgot to add an example like `a10` vs. `a2` in my test case for the base usage of `localeCompare` without the extra arguments. Thanks for refactoring the code so that the call to localeCompare is in one place. That way, if we decide to change it in the future, it only needs to happen there. I had one more minor change request in the comments, but otherwise everything looks good. 

It should be noted that localeCompare is listed as having the `locales` and `options` features missing on some mobile browsers. We're not using the `locales` arg here, but we are using `options`.

kchadha (approved):
@joshyrobot Thanks for the changes and also for looking into other possible options!

Issue #1490, opened 2018-05-02
Missing menu strings for next/previous backdrop and others....


chrisgarrity:
### Expected Behavior

The default values in menu lists (e.g. `switch backdrop to %1`) should be translatable.

### Actual Behavior

It looks like the `next backdrop` string is in Msgs (and Transifex) because it's a block, but neither of the menu values is in the table.

Translators are finding other missing menu items. Adding them here:
- [x] next backdrop
- [x] previous backdrop
- [x] myself (create clone of myself)
- [x] last (listindex)
- [x] all (listindex)
- [x] random (listindex)
- [ ] backdrop1 (several places)
- [ ] costume1
- [x] timer (when timer greater than)
- [ ] Sprite1 (sensing)
- [ ] Stage (sensing)

PR #1522, opened 2018-05-13
Add various missing translation strings

thisandagain:
### Resolves
Resolves GH-1490

### Proposed Changes

- Add translations for various missing strings
- Note: `backdrop1`, `costume1`, `Sprite1`, etc. are defined by the VM and will need to be discussed.


chrisgarrity (approved):
👍 LGTM

PR #1533, opened 2018-05-16
Add ordered option to getDescendants and all uses.

rachel-fenichel:
### Resolves

Ergonomics, mergeability

### Proposed Changes

Add an "ordered" parameter to getDescendants.
### Reason for Changes

Ergonomics, mergeability

### DIFF
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -268,6 +268,9 @@ Blockly.Workspace.prototype.getTopComments = function(ordered) {
 Blockly.Workspace.prototype.getAllBlocks = function(ordered) {
   if (ordered) {
     // Slow, but ordered.
+    // This gets all levels of descendants because getDescendants 
+    // is called recuusively.  They are added to a new list, not the
+    // list that it's iterating over.
     var topBlocks = this.getTopBlocks(true);
     var blocks = [];
     for (var i = 0; i < topBlocks.length; i++) {
@@ -275,6 +278,8 @@ Blockly.Workspace.prototype.getAllBlocks = function(ordered) {
     }
   } else {
     // Fast, but in no particular order.
+    // This gets all of levels of descendants by always adding to the 
+    // list that it's iterating over.
     var blocks = this.getTopBlocks(false);
     for (var i = 0; i < blocks.length; i++) {
       blocks.push.apply(blocks, blocks[i].getChildren(false));
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -260,19 +260,31 @@ Blockly.Workspace.prototype.getTopComments = function(ordered) {
 };
 
 /**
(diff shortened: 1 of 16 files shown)


core/workspace.js
```diff
@@ -260,19 +260,31 @@ Blockly.Workspace.prototype.getTopComments = function(ordered) {
...
+    var topBlocks = this.getTopBlocks(true);
+    var blocks = [];
+    for (var i = 0; i < topBlocks.length; i++) {
+      blocks.push.apply(blocks, topBlocks[i].getDescendants(true));
+    }
+  } else {
+    // Fast, but in no particular order.
+    var blocks = this.getTopBlocks(false);
+    for (var i = 0; i < blocks.length; i++) {
+      blocks.push.apply(blocks, blocks[i].getChildren(false));
```
kchadha: Does this actually get all the levels of nesting? I'm a little confused about why getDescendants is being used above (which seem to flatten all levels of nesting, but getChildren is being used in the unordered case.. it seems to only go one layer deep... (unless the `_childBlocks` property is just a list of all children (including arbitrary levels of nesting and next blocks)

rachel-fenichel: I think you're right and this doesn't get all the levels of nesting.  I'll ask the person who actually wrote this code, but I think that's just a bug.  Good catch.

rachel-fenichel: Actually, it *does* do the correct thing.  I think that's because it keeps adding to the array as it iterates over it.  So anything that it added earlier will eventually have `getChildren` called on it later.  
The loop ends when calling `getChildren` no longer adds anything to the list, and it finishes iterating through all of the remaining items that had been added previously.  But because of the way that it's iterating, and because it's going through the list as it builds it, there's no ordering.

The other version (ordered) has one array that it iterates over (`topBlocks`) and another that it actually pushes things to (`blocks`).  `getDescendants` is called on every item in `topBlocks`, and it recurses all the way through all of its children before returning.  The same structure applies inside `getDescendants`:

```
Blockly.Block.prototype.getDescendants = function(ordered) {
  var blocks = [this];
  var childBlocks = this.getChildren(ordered);
  for (var child, i = 0; child = childBlocks[i]; i++) {
    blocks.push.apply(blocks, child.getDescendants(ordered));
  }
  return blocks;
};
```

I just tested this in the playground and it works.  I agree that it's not obvious from just reading the code.

kchadha: Ah, okay. Maybe there should be a comment explaining this (but I don't think it's strictly necessary for this PR). Thank you for the explanation!

kchadha (approved):

rachel-fenichel:
I added some comments.  Merging now.

rachel-fenichel:
I added some comments to help clarify that code.  Merging now.

PR #1536, opened 2018-05-17
Add ability to overwrite default block colours

### LINKED ISSUE #1535: Unable to overwrite default block colours
### Expected Behavior

When colour options are passed into the option parameter when initializing a workspace using `Blockly.inject`, block colours for core scratch blocks, that use colours from the `Blockly.Colours` should use instead the new colours

for the motion blocks using 
```
primary: '#ff0000',
secondary: '#00ff00',
tertiary: '#0000ff',
```

we should get

screen shot 2018-05-17 at 1 05 35 pm

### Actual Behavior

Core blocks don't inherit primary (fill) or tertiary (border) colours from the initialization options, secondary colours do when there is a shadow type, but it is because they inherit from Blockly.Colours correctly

screen shot 2018-05-17 at 1 05 18 pm

### Steps to Reproduce

Initialize a scratch workspace using the following options: 

```
colours: {
    fieldShadow: 'rgba(255, 255, 255, 0.3)',
    dragShadowOpacity: 0.6,
    motion: {
        primary: '#ff0000',
        secondary: '#00ff00',
        tertiary: '#0000ff',
    },
}
```

vincentbriglia:
### Resolves

fixes #1535

### Proposed Changes

This pull request adds the ability to parametrize the core scratch block colours from the `Blockly.inject` call

### Reason for Changes

scratch-blocks already has the ability to change the colours of each category, this further allows the customization of block primary, secondary and tertiary colours, should other projects using scratch-blocks wish to have their own colour scheme

### Test Coverage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ common.pyc
 *.komodoproject
 /nbproject/private/
 
+# Editor
+.vscode
+
 /accessible/*
 /dist
 /msg/json/synonyms.json
@@ -31,3 +34,4 @@ common.pyc
 /gh-pages/_site
 /*compiler*.jar
 /local_blockly_compressed_vertical.js
+
--- a/.gitignore
+++ b/.gitignore
@@ -34,4 +34,3 @@ common.pyc
 /gh-pages/_site
 /*compiler*.jar
 /local_blockly_compressed_vertical.js
-
--- a/core/options.js
+++ b/core/options.js
@@ -127,7 +127,13 @@ Blockly.Options = function(options) {
           Blockly.Colours.hasOwnProperty(colourProperty)) {
         // If a property is in both colours option and Blockly.Colours,
         // set the Blockly.Colours value to the override.
-        Blockly.Colours[colourProperty] = colours[colourProperty];
+        // Ensure references are kept by extending a property when it's an object
+        var colourPropertyValue = colours[colourProperty];
+        if (goog.isObject(colourPropertyValue)) {
+          goog.object.extend(Blockly.Colours[colourProperty], colourPropertyValue);
+        } else {
+          Blockly.Colours[colourProperty] = colourPropertyValue;
+        }
       }
     }
   }
--- a/core/options.js
+++ b/core/options.js
@@ -127,8 +127,8 @@ Blockly.Options = function(options) {
           Blockly.Colours.hasOwnProperty(colourProperty)) {
         // If a property is in both colours option and Blockly.Colours,
         // set the Blockly.Colours value to the override.
-        // Ensure to override old Blockly category colours by reference instead
-        // of overriding them, as theya re used by reference elsewhere.
+        // Override Blockly category color object properties with those
+        // provided.
         var colourPropertyValue = colours[colourProperty];
         if (goog.isObject(colourPropertyValue)) {
           for (var colourSequence in colourPropertyValue) {
--- a/core/options.js
+++ b/core/options.js
@@ -127,10 +127,16 @@ Blockly.Options = function(options) {
           Blockly.Colours.hasOwnProperty(colourProperty)) {
         // If a property is in both colours option and Blockly.Colours,
         // set the Blockly.Colours value to the override.
-        // Ensure references are kept by extending a property when it's an object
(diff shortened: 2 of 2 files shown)


ZenithRogue:
You can rather easily do this with a userscript. Please see https://github.com/nitrocipher/theme3

mrjacobbloom:
~~I think this is talking about changing the colors for a specific block without changing the rest of its category, which isn't possible with the current attributes system~~
Never mind, I didn't look at the original issue

ZenithRogue:
@mrjacobbloom if you look at the issue,  it appears to be about the category

vincentbriglia:
The colours defined for a category (in the xml definition when you pass the toolbox) are different from the colours you define for the blocks in that category. The categoy dictates the colour of the toolbox, whereas the colours in `Blockly.Colours` defines those of the blocks themselves

- see for toolbox category colour definition of the `motion` category [here](https://github.com/LLK/scratch-blocks/blob/develop/blocks_vertical/default_toolbox.js#L32) 
- see for block colour definition of the blocks in the `motion` category [here](https://github.com/LLK/scratch-blocks/blob/develop/core/colours.js#L28) 

so to clarify, this change is unrelated to changing the colours of the toolbox category, but pertains solely to allowing a parametrized override of the default color of the blocks that inherit from the the `Blockly.Colours` object.

hope that clarifies this

kchadha (commented):
These changes look okay to me.

One thing I'm hesitant about is that we're having the old color scheme extend some arbitrary object. I'm wondering if that could result in any bugs down the line.

@rachel-fenichel, do you also want to take a look since I'm relatively unfamiliar with this part of the code base?

rachel-fenichel:
If the object is just supposed to have primary, secondary, and tertiary, I'd rather you check the input object for those things and set them directly on the existing object, rather than extending it.  I agree with @kchadha's concerns here.

.gitignore
```diff
@@ -13,6 +13,9 @@ common.pyc
 *.komodoproject
 /nbproject/private/
 
+# Editor
+.vscode
```
Kenny2github, line 17: Is a `.vscode` file a common editor artifact? Or would it be better to just not commit the vscode file and remove these lines?

.gitignore
```diff
@@ -31,3 +34,4 @@ common.pyc
 /gh-pages/_site
 /*compiler*.jar
 /local_blockly_compressed_vertical.js
+
```
Kenny2github: Just wondering why there's a random extra newline here.

Kenny2github (commented):
Just some questions.

.gitignore
kchadha: Please remove the stray newline here.

core/options.js
```diff
@@ -127,7 +127,19 @@ Blockly.Options = function(options) {
           Blockly.Colours.hasOwnProperty(colourProperty)) {
         // If a property is in both colours option and Blockly.Colours,
         // set the Blockly.Colours value to the override.
-        Blockly.Colours[colourProperty] = colours[colourProperty];
+        // Ensure to override old Blockly category colours by reference instead
+        // of overriding them, as theya re used by reference elsewhere.
```
kchadha: typo: 'theya re' -> 'they are'

I think this comment is a little confusing as is. I would say it's enough to say something along the lines of "Override Blockly category colors with those provided."

kchadha (changes requested):
@vincentbriglia, thank you for making the changes that were previously requested. I've requested some minor changes, and then this is good to go.

vincentbriglia:
thanks for being so stringent, it's good to keep the bar high 👍

.gitignore
vincentbriglia: Hi, vscode is a fairly popular editor, I merely added it here in the gitignore file as I contribute to a few different repositories, each with their own eslint/tslint/prettier/editorconfig setup bound to a potential editor like atom/vscode/intellij. adding the .vscode file in the pulled repository allows me to contribute to this project without automatically reformatting the js codebase according to my default editor settings. 

I don't think there's harm in ensuring that :)

kchadha (approved):
Looks good! Thank you for making these changes.

PR #1551, opened 2018-05-31
Randomized id for svg filters

### LINKED ISSUE #889: blocklyStackGlowFilter does not have a unique id
https://github.com/LLK/scratch-blocks/blob/develop/core/inject.js#L118 states that 

// Each filter/pattern needs a unique ID for the case of multiple Blockly
// instances on a page.  Browser behaviour becomes undefined otherwise.
// https://neil.fraser.name/news/2015/11/01/
// TODO (tmickel): Look into whether block highlighting still works.

but the stack glow filter and replacement glow filter do not follow this rule.

### LINKED ISSUE #1549: Displaying an embedded workspace causes field highlights to stop working
### Expected Behavior

Field highlights should always work.

### Actual Behavior

After displaying an embedded workspace (for example, the custom procedures modal) and subsequently dismissing it, fields will disappear when dragging a compatible block over them instead of glowing.

highlight

### Steps to Reproduce

Prerequisites: Safari 11.1 or iOS browser (Chrome 66 is not affected).
1. Load the Scratch Preview
2. Add a block to the workspace and verify that the field highlights correctly when you drag a block into it.
3. Click "Make a block" under "My Blocks"
4. Dismiss the modal
5. Drag a parameter into the field again. Notice that the field disappears instead of highlighting.

Mac OS 10.13.4, Safari 11.1
iOS 11

howettl:
### Resolves

Resolves #1549 
Resolves #889 

### Proposed Changes

Randomly generate the id for the stack glow filter and replacement glow filter

### Reason for Changes

When `inject` is called for an embedded workspace (ie. when displaying the custom procedures modal), the stack and replacement glow filters are generated again with the same ID as the original instances.  This is an issue in Safari and iOS browsers which then stop displaying the filters when the embedded workspace is destroyed.  By randomly generating the filter IDs, we prevent collisions.

### Test Coverage

Tested by instantiating an embedded workspace and verifying the fix.

### DIFF
--- a/core/block_render_svg_horizontal.js
+++ b/core/block_render_svg_horizontal.js
@@ -335,7 +335,9 @@ Blockly.BlockSvg.prototype.updateColour = function() {
  */
 Blockly.BlockSvg.prototype.highlightForReplacement = function(add) {
   if (add) {
-    this.svgPath_.setAttribute('filter', 'url(#blocklyReplacementGlowFilter)');
+    var replacementGlowFilterId = this.workspace.options.replacementGlowFilterId
+      || 'blocklyReplacementGlowFilter';
+    this.svgPath_.setAttribute('filter', 'url(#' + replacementGlowFilterId + ')');
     Blockly.utils.addClass(/** @type {!Element} */ (this.svgGroup_),
         'blocklyReplaceable');
   } else {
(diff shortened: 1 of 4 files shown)


rachel-fenichel:
Also resolves https://github.com/LLK/scratch-blocks/issues/889, probably.

rachel-fenichel:
LGTM, Paul should double-check in context of scratch-gui.

howettl:
You're right, it would resolve #889. I updated the description.

paulkaplan:
Wonderful! Thanks!

paulkaplan (approved):

paulkaplan:
As a follow up, it looks like we missed passing the stack glow filter id down into the toolbox/flyout workspace, which caused blocks in the toolbox to disappear when clicked! Fixed now, my bad for missing that fact.

PR #1552, opened 2018-05-31
Clear selected block before executing undo/redo

### LINKED ISSUE #773: Undo/Redo UI buttons in vertical playground is broken.
### Expected Behavior

Undo and Redo buttons in the vertical playground should work exactly as the context menu.

### Actual Behavior

When undo and redo buttons are clicked repetitively in the vertical playground, the blocks in the workspace are incorrectly restored.

### Steps to Reproduce

1. place a repeat block in the workspace.
2. place another block inside the repeat block as shown in the first attachment.
3. detach the inner block of the repeat block.
4. press undo (in the left side menu) and notice the block goes back inside the repeat block. (correct behavior)
5. press redo (in the left side menu) and notice the block now goes to a different location then it was before (incorrect behavior)
6. now press undo again (in the left side menu) and notice the inner block now is in a broken state.

screen shot 2017-01-23 at 10 06 46 am
screen shot 2017-01-23 at 10 07 33 am

NOTE: the Undo and Redo works as expected when using the context menu.  

_e.g. Mac OS 10.11.6 chrome 55.0.2883.95

howettl:
### Resolves

Resolves #773 

### Proposed Changes

Clear any selected block before executing an undo/redo.

### Reason for Changes

When attempting to call undo and then redo on a currently selected block, the call to `block.setParent(null)` isn't complete as it does not properly append the block to the workspace and instead it stays attached to its current block. (see https://github.com/LLK/scratch-blocks/blob/f7be0a9ac7742f80c421b1b894f856b31cca47ea/core/block_svg.js#L315). This means when the redo process attempts to reconnect the block, it throws the error `"Failed to execute 'appendChild' on 'Node': The new child element contains the parent."`

### Test Coverage

Tested by following the steps to reproduce in #773.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,5 +31,3 @@ common.pyc
 /gh-pages/_site
 /*compiler*.jar
 /local_blockly_compressed_vertical.js
-
-\.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ common.pyc
 /gh-pages/_site
 /*compiler*.jar
 /local_blockly_compressed_vertical.js
+
+\.idea/
--- a/core/workspace.js
+++ b/core/workspace.js
@@ -488,6 +488,14 @@ Blockly.Workspace.prototype.undo = function(redo) {
   }
   events = Blockly.Events.filter(events, redo);
   Blockly.Events.recordUndo = false;
+  if (Blockly.selected) {
+    Blockly.Events.disable();
+    try {
+      Blockly.selected.unselect();
+    } finally {
+      Blockly.Events.enable();
+    }
+  }
   try {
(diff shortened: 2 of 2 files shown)


rachel-fenichel:
Interesting.  The check for selected came up recently as a weird check, but we didn't catch that it was causing the undo/redo problems as well.

#1553 gets rid of the check for selected.  Previous debugging on Blockly showed that it was definitely wrong.

@paulkaplan does unsetting the selected block without firing an event cause any problems with the VM's understanding of what the blocks are doing?  Does the selected block usually get unset or otherwise changed at some point during an undo anyway?

paulkaplan:
I do not think the selected state of blocks gets processed by the VM at all

PR #1555, opened 2018-06-01
Pass stack glow filter to the toolbox

paulkaplan:
### Resolves

Fixes an issue caused by #1551 that wasn't caught at the time, that the toolbox can stack glow and so needs to be passed the parents stack glow filter id when it is initialized.

### Proposed Changes

Pass down the stack glow filter ID when creating a flyout, or use the workspaces stack glow filter when creating a toolbox.

Note this is required in two places, because you can create a palette of blocks either by creating just a flyout ("simple" mode in the playground) or by creating a toolbox ("categories" mode) which then creates a flyout.


ericrosenbaum (approved):
looks good- thanks @paulkaplan!

PR #1556, opened 2018-06-01
Updated message key that was missed when original key was changed

howettl:
### Resolves

No issue filed that I'm aware of.

### Proposed Changes

The key for the "duplicate" string was recently changed (here: https://github.com/LLK/scratch-blocks/commit/e38969831eb53b19a380ae8faec4ca43e0904404) but one case was not updated.

### Reason for Changes

This was causing the user to be able to duplicate a custom procedure definition block because the key check was failing.

### Test Coverage

Verified that custom procedure definition blocks cannot be duplicated after updating the key reference.


kchadha (approved):
@howettl Thanks for catching this!

Issue #1554, opened 2018-06-01
Comments ToDo


kchadha:
Making one giant Issue for the separate comment tasks that still remain:

- [x] Workspace comments need to be updated to look/work like block comments: Fixed with #1575 
    - [x] Size of comment text
    - [x] Font used for comment text
    - [x] Size of top bar
    - [x] Minimize icon and functionality
    - [x] Delete icon and functionality
- [ ] Some outstanding issues with drag + text selection - Tracking in #1623
- [ ] Auto layout of block comments needs improvement - Tracking in #1624
    - [ ] Block comments render too far away from blocks in wide stacks
- [x] Workspace does not scroll with block comments (e.g. if a block comment gets created or dragged far off, can't scroll the workspace to it) Fixed with #1557
- [ ] Cannot scroll workspace if mouse happens to be over a block comment - Tracking in #1625
- [x] Workspace scroll is wonky if block comments are on the screen. Scroll gets pinned to a corner of the workspace. Fixed with #1587
- [ ] Comment rendering needs improvement - Tracking in #1626
- [ ] Delete block comments by dragging to the toolbox - Tracking in #1627
- [ ] Block comment bubble dragging does not work in RTL - Tracking in google/blockly#1896
- [x] Persist comment state Fixed with #1557
    - [x] Add events to block comments
    - [x] Undo/redo
    - [x] Add event listeners for block and workspace comments in scratch-vm (Fixed with LLK/scratch-vm#1191
    - [x] Modify CommentChange event to track changes in minimized state
    - [x] Track resizing changes through events
- [ ] Comment anchor point is off for custom procedure definition block - Tracking in #1628
- [ ] Allow dragging block comment from middle of comment (like workspace comments) - Tracking in #1629
- [ ] Need minimum resizing dimensions for comments - Tracking in #1630
- [ ] Cannot edit the text of a comment on touch devices - Tracking in #1631
- [x] Comments overlap each other: Fixed with #1589
![overlapping comments](https://user-images.githubusercontent.com/1786240/41428384-f1b3bd34-6fd7-11e8-883d-c2120649fbf9.gif)
![image 2](https://user-images.githubusercontent.com/1786240/41428414-0bbf41da-6fd8-11e8-9642-a6cb77dea34e.png)

rachel-fenichel:
Related to "Cannot scroll workspace if mouse happens to be over a block comment": You probably can't scroll comment contents.  I think that's fine, since you can resize them, but it's worth knowing.

[...]

PR #1557, opened 2018-06-01
Comment Events

kchadha:
### Resolves

All the scratch-vm parts of 'persist comment state' section of #1554.

### Proposed Changes

- add comment events to block comments
- track resize and minimize changes in CommentChange event (track both of these for block comments)
- track resize changes for workspace comments (minimize functionality has not been implemented yet for workspace comments)
- fire a single CommentCreate event when adding a workspace comment (instead of CommentCreate and then CommentMove -- which would make using `Undo` after using `Add Comment` in the workspace context menu result in the comment moving instead of getting deleted.)
- rename comment event file since it now pertains to both block and workspace comments
- add more information to CommentCreate event for use during undo/redo

### Reason for Changes

Persisting state about comments in the scratch-vm. Events are good.

### Test Coverage

Testing in vertical playground as well as scratch-gui (along with the changes in LLK/scratch-vm#1191).


core/events.js
```diff
@@ -132,6 +132,30 @@ Blockly.Events.VAR_DELETE = 'var_delete';
  */
 Blockly.Events.VAR_RENAME = 'var_rename';
 
+/**
+ * Name of event that creates a variable.
```
rachel-fenichel: Fix comments on all of these.

core/comment_events.js
```diff
@@ -96,6 +106,23 @@ Blockly.Events.CommentBase.prototype.toJson = function() {
...
+ * Helper function for finding the comment this event pertains to.
+ * @return {Blockly.WorkspaceComment | Blockly.ScratchBlockComment}
+ *     The comment this event pertains to.
+ * @private
+ */
+Blockly.Events.CommentBase.prototype.getComment_ = function() {
+  var workspace = this.getEventWorkspace_();
+  if (this.blockId) {
+    var block = workspace.getBlockById(this.blockId);
+    if (block) return block.comment;
```
rachel-fenichel: Nit: return on a new line, and braces, even for simple stuff like this.  Here and throughout the PR.

core/comment_events.js
```diff
@@ -130,6 +157,7 @@ Blockly.Events.CommentChange.prototype.type = Blockly.Events.COMMENT_CHANGE;
 Blockly.Events.CommentChange.prototype.toJson = function() {
   var json = Blockly.Events.CommentChange.superClass_.toJson.call(this);
   json['newContents'] = this.newContents_;
+
```
rachel-fenichel: Nit: remove this newline.

core/scratch_block_comment.js
```diff
@@ -39,18 +39,30 @@ goog.require('goog.userAgent');
...
  */
-Blockly.ScratchBlockComment = function(block, x, y, minimized) {
+Blockly.ScratchBlockComment = function(block, id, x, y, minimized) {
   Blockly.ScratchBlockComment.superClass_.constructor.call(this, block);
   this.x_ = x;
   this.y_ = y;
   this.isMinimized_ = minimized || false;
+
+  this.workspace = block.workspace;
+  this.id = goog.isString(id) ? id : Blockly.utils.genUid();
```
rachel-fenichel: The id is essentially unused.  There's no way to look up the comment by ID--you're always looking it up by block ID instead.

You should insert the comment into the workspace's commentDB.  That also ensures that the workspace takes the position/size of the comment into account when getting metrics.

rachel-fenichel (commented):
One major comment that will require some changes, and two nits that I found on my way to the major comment.

core/scratch_block_comment.js
kchadha: Oh cool, I didn't know that re: "That also ensures that the workspace takes the position/size of the comment into account when getting metrics."

That's awesome.

core/comment_events.js
```diff
@@ -96,6 +106,18 @@ Blockly.Events.CommentBase.prototype.toJson = function() {
 Blockly.Events.CommentBase.prototype.fromJson = function(json) {
   this.commentId = json['commentId'];
   this.group = json['group'];
+  this.blockId = json['blockId'];
+};
+
+/**
+ * Helper function for finding the comment this event pertains to.
+ * @return {Blockly.WorkspaceComment | Blockly.ScratchBlockComment}
```
rachel-fenichel: Or null if it no longer exists.

core/xml.js
```diff
@@ -703,7 +705,11 @@ Blockly.Xml.domToBlockHeadless_ = function(xmlBlock, workspace) {
         var bubbleH = parseInt(xmlChild.getAttribute('h'), 10);
         if (!isNaN(bubbleW) && !isNaN(bubbleH) &&
             block.comment && block.comment.setVisible) {
-          block.comment.setBubbleSize(bubbleW, bubbleH);
+          if (block.comment instanceof Blockly.ScratchBlockComment) {
+            block.commet.setSize(bubbleW, bubbleH);
```
rachel-fenichel: commet->comment

core/bubble_dragger.js
```diff
@@ -240,6 +240,11 @@ Blockly.BubbleDragger.prototype.fireMoveEvent_ = function() {
     event.setOldCoordinate(this.startXY_);
```
rachel-fenichel: You can get rid of some of the duplicated code:
```
if (this.draggingBubble_.isComment) {
  var comment = this.draggingBubble_;
} else if (this.draggingBubble_ instanceof Blockly.ScratchBubble) {
  var comment = this.draggingBubble_.comment;
} else {
  return;  // No move events for other bubbles.
} 
var event = new Blockly.Events.CommentMove(comment);
event.setOldCoordinate(this.startXY_);
event.recordNew();
Blockly.Events.fire(event);
```

You can also get rid of the TODO and the return.

core/comment_events.js
```diff
@@ -155,21 +177,40 @@ Blockly.Events.CommentChange.prototype.isNull = function() {
...
-  var comment = workspace.getCommentById(this.commentId);
+  var comment = this.getComment_();
   if (!comment) {
     console.warn('Can\'t change non-existent comment: ' + this.commentId);
     return;
   }
   var contents = forward ? this.newContents_ : this.oldContents_;
 
-  comment.setContent(contents);
+  if (goog.isString(contents)) {
```
rachel-fenichel: Contents used to be only a string, and that's what it's still annotated as in the constructor.  Now it could be either a string or an object, and you have to decide what to do based on that.  If it's an object it may have multiple things that you care about on it.

I prefer having the object optionally have a `contents` property, containing this string.  That would make the API more consistent.  It would require changes in workspace comments as well, but not large ones.

The code below means that you can't simultaneously change (for example) minimized state and height/width.  Is that necessary?

rachel-fenichel: And in either case, update annotations to be accurate.

core/comment_events.js
```diff
@@ -155,21 +177,40 @@ Blockly.Events.CommentChange.prototype.isNull = function() {
...
+      }
+      return;
+    } else if (contents.hasOwnProperty('width') && contents.hasOwnProperty('height')) {
+      comment.setSize(contents.width, contents.height);
+      return;
+    }
+  }
+
+  console.warn('Unrecognized comment change: ' + JSON.stringify(contents) + '; cannot undo/redo');
+  return;
```
rachel-fenichel: Unnecessary return.

core/comment_events.js
```diff
@@ -155,21 +177,40 @@ Blockly.Events.CommentChange.prototype.isNull = function() {
...
 };
 
 /**
  * Class for a comment creation event.
  * @param {Blockly.WorkspaceComment} comment The created comment.
```
rachel-fenichel: Update type.

core/comment_events.js
```diff
@@ -179,6 +220,41 @@ Blockly.Events.CommentCreate = function(comment) {
   }
   Blockly.Events.CommentCreate.superClass_.constructor.call(this, comment);
 
+  /**
+   * The text content of this comment.
+   * @type{string}
```
rachel-fenichel: nit: space between "type" and {

Here and below.

core/comment_events.js
```diff
@@ -179,6 +220,41 @@ Blockly.Events.CommentCreate = function(comment) {
...
+   */
+  this.text = comment.getText();
+
+  /**
+   * The XY position of this comment on the workspace.
+   * @type{goog.math.Coordinate}
+   */
+  this.xy = comment.getXY();
```
rachel-fenichel: Make sure getXY and getHeightWidth return appropriate values in headless mode, especially for workspace comments.

core/comment_events.js
```diff
@@ -399,15 +495,19 @@ Blockly.Events.CommentMove.prototype.isNull = function() {
...
 
   var target = forward ? this.newCoordinate_ : this.oldCoordinate_;
-  // TODO: Check if the comment is being dragged, and give up if so.
-  var current = comment.getXY();
-  comment.moveBy(target.x - current.x, target.y - current.y);
+
+  if (comment instanceof Blockly.ScratchBlockComment) {
+    comment.moveTo(target.x, target.y);
+  } else {
+    // TODO: Check if the comment is being dragged, and give up if so.
```
rachel-fenichel, line 509: Does this TODO apply to both, or only workspace comments?

Do workspace comments need a `moveTo` function as well as a `moveBy` function?

core/contextmenu.js
```diff
@@ -490,6 +499,8 @@ Blockly.ContextMenu.workspaceCommentOption = function(ws, e) {
       comment.render(false);
       comment.select();
     }
+    if (disabled) Blockly.Events.enable();
```
rachel-fenichel: Expand if to multiple lines + braces.

core/events.js
```diff
@@ -132,6 +132,31 @@ Blockly.Events.VAR_DELETE = 'var_delete';
...
+ * (text content, size, or minimized state).
+ * @const
+ */
+Blockly.Events.COMMENT_CHANGE = 'comment_change';
+
+/**
+ * Name of event that deletes a comment.
+ * @const
+ */
+Blockly.Events.COMMENT_DELETE = 'comment_delete';
```
rachel-fenichel, line 158: Also add this code to `fromJson`:
```

    case Blockly.Events.COMMENT_CREATE:
      event = new Blockly.Events.CommentCreate(null);
      break;
    case Blockly.Events.COMMENT_CHANGE:
      event = new Blockly.Events.CommentChange(null);
      break;
    case Blockly.Events.COMMENT_MOVE:
      event = new Blockly.Events.CommentMove(null);
      break;
    case Blockly.Events.COMMENT_DELETE:
      event = new Blockly.Events.CommentDelete(null);
      break;
```

core/scratch_block_comment.js
```diff
@@ -39,18 +39,35 @@ goog.require('goog.userAgent');
...
  * @param {number=} y Initial y position for comment, in workspace coordinates.
  * @param {boolean=} minimized Whether or not this comment is minimized
  *     (only the top bar displays), defaults to false.
  * @extends {Blockly.Comment}
  * @constructor
  */
-Blockly.ScratchBlockComment = function(block, x, y, minimized) {
+Blockly.ScratchBlockComment = function(block, text, id, x, y, minimized) {
   Blockly.ScratchBlockComment.superClass_.constructor.call(this, block);
+  this.text_ = text;
```
rachel-fenichel: Annotate properties.

core/scratch_block_comment.js
```diff
@@ -39,18 +39,35 @@ goog.require('goog.userAgent');
...
   this.x_ = x;
   this.y_ = y;
   this.isMinimized_ = minimized || false;
+
+  this.workspace = block.workspace;
+  this.id = goog.isString(id) && !this.workspace.getCommentById(id) ?
+      id : Blockly.utils.genUid();
+  this.workspace.addTopComment(this);
```
rachel-fenichel: Nit: should be workspace.addComment, since "top comments" no longer means anything.  We can fix that later.

core/scratch_block_comment.js
```diff
@@ -288,6 +303,7 @@ Blockly.ScratchBlockComment.prototype.setVisible = function(visible) {
   // Restore the bubble stats after the visibility switch.
   this.setText(text);
   this.setBubbleSize(size.width, size.height);
+  if (visible) Blockly.ScratchBlockComment.fireCreateEvent(this);
```
rachel-fenichel: Make this if multiline and add braces.

core/scratch_block_comment.js
```diff
@@ -305,6 +321,8 @@ Blockly.ScratchBlockComment.prototype.toggleMinimize_ = function() {
  */
 Blockly.ScratchBlockComment.prototype.setMinimized = function(minimize) {
   if (this.isMinimized_ == minimize) return;
```
rachel-fenichel: Make this if multiline and add braces.

core/scratch_block_comment.js
```diff
@@ -333,12 +351,32 @@ Blockly.ScratchBlockComment.prototype.setBubbleSize = function(width, height) {
...
-      this.height_ = height;
     }
-  } else {
-    this.width_ = width;
-    this.height_ = height;
+  }
+};
+
+/**
+ * Set the un-minimized size of this comment. If the comment has an un-minimized
```
rachel-fenichel, line 398: @package if possible.

core/scratch_block_comment.js
```diff
@@ -415,6 +456,33 @@ Blockly.ScratchBlockComment.prototype.getHeightWidth = function() {
...
+ *    Object with top left and bottom right coordinates of the bounding box.
+ * @package
+ */
+Blockly.ScratchBlockComment.prototype.getBoundingRectangle = function() {
+  var commentXY = this.getXY();
+  var commentBounds = this.getHeightWidth();
+  var topLeft;
+  var bottomRight;
+  if (this.workspace.RTL) {
+    // TODO for some reason this doesn't work with workspace scroll in RTL
```
rachel-fenichel: Make an issue for this and add the number here.

core/scratch_block_comment.js
```diff
@@ -434,3 +502,57 @@ Blockly.ScratchBlockComment.prototype.showContextMenu_ = function(e) {
...
+  element.setAttribute('id', this.id);
+  element.textContent = this.text_;
+  element.setAttribute('x', Math.round(this.x));
+  element.setAttribute('y', Math.round(this.y));
+  element.setAttribute('h', this.height_);
+  element.setAttribute('w', this.width_);
+  return element;
+};
+
+
```
rachel-fenichel: Nit: extra newline.

core/scratch_bubble.js
```diff
@@ -50,16 +52,20 @@ goog.require('goog.userAgent');
  * @extends {Blockly.Bubble}
  * @constructor
  */
-Blockly.ScratchBubble = function(workspace, content, anchorXY,
+Blockly.ScratchBubble = function(comment, workspace, content, anchorXY,
     bubbleWidth, bubbleHeight, bubbleX, bubbleY, minimized) {
 
+  // Needed for Events
+  this.comment = comment;
```
rachel-fenichel: Annotate with type and visibility.

core/workspace_comment_render_svg.js
```diff
@@ -332,6 +335,15 @@ Blockly.WorkspaceCommentSvg.prototype.unbindDragEvents_ = function() {
 Blockly.WorkspaceCommentSvg.prototype.resizeMouseUp_ = function(/*e*/) {
   Blockly.Touch.clearTouchIdentifier();
   this.unbindDragEvents_();
+  var oldHW = this.resizeStartSize_;
+  this.resizeStartSize_ = null;
+  if (this.width_ == oldHW.width &&
+      this.height_ == oldHW.height) return;
```
rachel-fenichel: Make this if multiline and add braces.

core/workspace_comment_render_svg.js
```diff
@@ -342,7 +354,18 @@ Blockly.WorkspaceCommentSvg.prototype.resizeMouseUp_ = function(/*e*/) {
...
+  // emitted while the user is still in the process of resizing
+  // the comment, so disable events here. The event is emitted in
+  // resizeMouseUp_.
+  var disabled = false;
+  if (Blockly.Events.isEnabled()) {
+    Blockly.Events.disable();
+    disabled = true;
+  }
+  this.setSize(this.RTL ? -newXY.x : newXY.x, newXY.y);
+  if (disabled) Blockly.Events.enable();
```
rachel-fenichel: Make this if multiline and add braces

core/bubble_dragger.js
kchadha: I actually even noticed that and was going to 'come back to it' but then forgot haha.

core/comment_events.js
kchadha: Sounds good. I’ll just make it an object instead of a string or object. I was considering doing that anyways but originally thought this might be a less invasive change.

In terms of not being able to have the multiple state items change in a single change event, I made it that way because a single user interaction shouldn’t be able to change both, and we’re never emitting a CommentChange event with multiple state changes. I figure if we ever need that, I’d rather the developer have to explicitly change this piece of code for this new kind of functionality.

core/comment_events.js
kchadha: I wasn't sure about this TODO, it was left over from before I added the instanceof check. I'm not sure what scenario would lend itself to this code getting run while the comment is being dragged...

core/scratch_block_comment.js
kchadha: #1562

core/comment_events.js
kchadha: #1563

core/scratch_block_comment.js
kchadha: #1564

core/comment_events.js
```diff
@@ -96,14 +106,32 @@ Blockly.Events.CommentBase.prototype.toJson = function() {
...
+  return workspace.getCommentById(this.commentId);
 };
 
 /**
  * Class for a comment change event.
  * @param {Blockly.WorkspaceComment} comment The comment that is being changed.
  *     Null for a blank event.
- * @param {string} oldContents Previous contents of the comment.
- * @param {string} newContents New contents of the comment.
+ * @param {!object} oldContents Object containing previous state of a comment's
```
rachel-fenichel: We should validate this, but doesn't have to be now.  I just filed an issue: #1565

core/comment_events.js
```diff
@@ -243,6 +329,13 @@ Blockly.Events.CommentDelete = function(comment) {
     return;  // Blank event to be populated by fromJson.
   }
   Blockly.Events.CommentDelete.superClass_.constructor.call(this, comment);
+  this.xy = comment.getXY();
```
rachel-fenichel, line 329: Update type annotation on the `comment` param to include ScratchBlocksComments.

When I figure out how to make typedefs work, we can typedef those two together for use in all of these places.

core/comment_events.js
rachel-fenichel: Okay, we can leave it and I can come back and try to figure out if that's discussing a valid problem.  If not I'll delete it.

core/scratch_block_comment.js
```diff
@@ -434,3 +535,56 @@ Blockly.ScratchBlockComment.prototype.showContextMenu_ = function(e) {
...
+ * Encode a comment subtree as XML with XY coordinates.
+ * @param {boolean=} opt_noId True if the encoder should skip the comment id.
+ * @return {!Element} Tree of XML elements.
+ * @package
+ */
+Blockly.ScratchBlockComment.prototype.toXmlWithXY = function() {
+  var element = goog.dom.createDom('comment');
+  element.setAttribute('id', this.id);
+  element.textContent = this.text_;
+  element.setAttribute('x', Math.round(this.x));
```
rachel-fenichel: `this.x` and `this.y` -> `this.x_` and `this.y_`

core/scratch_block_comment.js
```diff
@@ -39,18 +39,63 @@ goog.require('goog.userAgent');
...
  *     (only the top bar displays), defaults to false.
  * @extends {Blockly.Comment}
  * @constructor
  */
-Blockly.ScratchBlockComment = function(block, x, y, minimized) {
+Blockly.ScratchBlockComment = function(block, text, id, x, y, minimized) {
   Blockly.ScratchBlockComment.superClass_.constructor.call(this, block);
+  /**
+   * The text content of this comment.
+   * @type {string}
```
rachel-fenichel, line 56: @private for the underscore ones.

core/comment_events.js
kchadha: Done.

core/scratch_block_comment.js
kchadha: Good catch!!

core/scratch_block_comment.js
kchadha: done

core/comment_events.js
kchadha: Thank you for filing the issue!

kchadha:
@rachel-fenichel,

Re: https://github.com/LLK/scratch-blocks/pull/1557#discussion_r193556107

I ended up making the change to allow the `CommentChange.prototype.run` function to handle multiple types of changes at once (e.g. minimize state change + text change).

@paulkaplan and I talked about a similar thing I was doing in scratch-vm and we came to the conclusion that since there's no real reason a CommentChange event has to pertain to a single state change, we can avoid potential future "Is this code like this for a reason?" questions by letting this code handle those cases where a future developer might emit a single CommentChange event with multiple state changes.

core/comment_events.js
kchadha: Re: my previous comment in this thread, I ended up making the change to allow multiple types of comment changes at once (e.g. minimized state and text).

core/comment_events.js
```diff
@@ -155,21 +183,36 @@ Blockly.Events.CommentChange.prototype.isNull = function() {
...
+    }
+  }
+  if (contents.hasOwnProperty('width') && contents.hasOwnProperty('height')) {
+    comment.setSize(contents.width, contents.height);
+  }
+  if (contents.hasOwnProperty('text')) {
+    comment.setText(contents.text);
+  }
+
+  console.warn('Unrecognized comment change: ' + JSON.stringify(contents) + '; cannot undo/redo');
```
rachel-fenichel: This now fires every single time.  Delete this line.

rachel-fenichel (commented):
One more change, then LGTM!

core/comment_events.js
kchadha: Oops! Totally forgot about that. =]

kchadha:
@rachel-fenichel wohooo! We made it! =D

rachel-fenichel:
pikachu_dancing

PR #1558, opened 2018-06-05
python build script fixed

harsh424jan:
### Resolves

Resolves issue [#2119](https://github.com/LLK/scratch-gui/issues/2119) in scratch-gui

### Proposed Changes

While checking for remote compiler in do_compile, compare self.closure_env["closure_compiler"] with REMOTE_COMPILER instead of self.closure_env

### Reason for Changes

Error running build script

### Test Coverage

Run the file to test for it's correctness

### DIFF
--- a/build.py
+++ b/build.py
@@ -322,7 +322,7 @@ class Gen_compressed(threading.Thread):
     self.do_compile(params, target_filename, filenames, remove)
 
   def do_compile(self, params, target_filename, filenames, remove):
-    do_compile = self.do_compile_remote if self.closure_env == REMOTE_COMPILER else self.do_compile_local
+    do_compile = self.do_compile_remote if self.closure_env["closure_compiler"] == REMOTE_COMPILER else self.do_compile_local
     json_data = do_compile(params, target_filename)
 
     if self.report_errors(target_filename, filenames, json_data):
@@ -349,7 +349,7 @@ class Gen_compressed(threading.Thread):
       for group in [["google-closure-compiler"], dash_args]:
         args.extend(filter(lambda item: item, group))
 
-      proc = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+      proc = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE,shell=True)
       (stdout, stderr) = proc.communicate()
 
       # Build the JSON response.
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -322,7 +322,7 @@ def gen_generator(self, language):
     self.do_compile(params, target_filename, filenames, remove)
 
   def do_compile(self, params, target_filename, filenames, remove):
-    do_compile = self.do_compile_remote if self.closure_env == REMOTE_COMPILER else self.do_compile_local
+    do_compile = self.do_compile_remote if self.closure_env["closure_compiler"] == REMOTE_COMPILER else self.do_compile_local
```
rschamp, line 325: I believe this has since been resolved with https://github.com/LLK/scratch-blocks/pull/1512

build.py
```diff
@@ -349,7 +349,7 @@ def do_compile_local(self, params, target_filename):
       for group in [["google-closure-compiler"], dash_args]:
         args.extend(filter(lambda item: item, group))
 
-      proc = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+      proc = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE,shell=True)
```
rschamp, line 352: What is the reason for using the `shell` kwarg?

build.py
```diff
@@ -592,7 +592,7 @@ def exclude_horizontal(item):
 
     # Sanity check the local compiler
     test_args = [closure_compiler, os.path.join("build", "test_input.js")]
-    test_proc = subprocess.Popen(test_args, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    test_proc = subprocess.Popen(test_args, stdin=subprocess.PIPE, stdout=subprocess.PIPE,shell=True)
```
rschamp, line 595: Same question here - why `shell=True`?

rschamp (commented):
I believe the important part of this PR has already been resolved. Since the tests also fail, I think I'm going to close this unless the `shell` argument should be included. I think the common advice is to avoid `shell` if possible.

lishuai199502:
i have tried add shell=true to line352 and line595,but failed finally.
image

harsh424jan:
image

build.py not working when shell = False .
This is the reason for shell=True

PRCBWQ:
> e tried add shell=true to line352 and line595,but failed finally.

can you solve this problem? I have same problem

PR #1559, opened 2018-06-06
Add when touching hat

ericrosenbaum:
### Resolves

Part of https://github.com/LLK/scratch-gui/issues/607

### Proposed Changes

Add a hat block for when a sprite is touching the mouse pointer, edge of the stage, or another sprite.

### Reason for Changes

This is a "low floor" feature, making it easier to make simple projects where you want a sprite to do something when it touches e.g. another sprite.

### DIFF
--- a/blocks_vertical/event.js
+++ b/blocks_vertical/event.js
@@ -29,12 +29,12 @@ goog.require('Blockly.ScratchBlocks.VerticalExtensions');
 
 Blockly.Blocks['event_whentouchingobject'] = {
   /**
-   * Block to when a sprite is touching an object.
+   * Block for when a sprite is touching an object.
    * @this Blockly.Block
    */
   init: function() {
     this.jsonInit({
-      "message0": "when this sprite touches %1",
+      "message0": Blockly.Msg.EVENT_WHENTOUCHINGOBJECT,
       "args0": [
         {
           "type": "input_value",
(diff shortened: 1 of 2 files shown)


blocks_vertical/event.js
```diff
@@ -27,6 +27,48 @@ goog.require('Blockly.Colours');
 goog.require('Blockly.constants');
 goog.require('Blockly.ScratchBlocks.VerticalExtensions');
 
+Blockly.Blocks['event_whentouchingobject'] = {
+  /**
+   * Block to when a sprite is touching an object.
```
kchadha: 'to' -> 'for' (or 'to fire'/'that fires')

blocks_vertical/event.js
```diff
@@ -27,6 +27,48 @@ goog.require('Blockly.Colours');
...
+   * @this Blockly.Block
+   */
+  init: function() {
+    this.jsonInit({
+      "message0": "when this sprite touches %1",
```
kchadha: This string should go in msg/messages.js so that this block can get translated. (E.g. 'Blockly.Msg.EVENT_WHENFLAGCLICKED').

I realize we might have this same problem in other places, so we might need to do an audit of that...

kchadha (changes requested):
Some changes needed.

kchadha (approved):

Mr-Dave2:
For the sake of beginners you might want to add in the GUI a way to disable these new blocks since they are always 'hot'. For example, if a program has {"when this sprite touches mouse-pointer", Hide } then trying to manually position that sprite on the stage will be confusing. This should already be a problem for "when this sprite is clicked" except that event is now a trailing-edge event (which is a separate issue for some current projects).  [Sorry for for the duplicate post. I had carelessly attached the first one to the wrong issue. ]

PR #1560, opened 2018-06-06
Add new menu type field matrix

khanning:
### Proposed Changes

Create a new field type for 5x5 LED matrix displays.
image

### Reason for Changes

For interacting with hardware devices that have programmable LED matrix displays (e.g. micro:bit).

#### OS X (High Sierra 10.13.4)
  - Chrome (67.0.3396.99)
  - Firefox (52.9.0)
  - Safari (11.1)

#### Android (8.1.0)
  - Chrome (67.0.3396.87)


khanning:
Thanks to @rachel-fenichel the colour scheme issue is fixed! The PR can now be reviewed.

core/field_matrix.js
```diff
@@ -0,0 +1,456 @@
...
+  // Build the DOM.
+  this.fieldGroup_ = Blockly.utils.createSvgElement('g', {}, null);
+  this.size_.width = Blockly.FieldMatrix.MATRIX_THUMB_SIZE +
+    Blockly.FieldMatrix.ARROW_SIZE + (Blockly.BlockSvg.DROPDOWN_ARROW_PADDING * 1.5);
+
+  this.sourceBlock_.getSvgRoot().appendChild(this.fieldGroup_);
+
+  this.getSourceColours_();
+
+  if (!this.ledThumbs_) {
```
rachel-fenichel: General comment:
Any properties that the field has should be explicitly declared and annotated, often in the constructor, unless they're declared in the superclass.

This isn't done in most of the existing fields because they got created before I understood how helpful these annotations are.

Benefits:
- When debugging down the line, you won't need to search for the first place that a property is used to know what it is
- Automated type checking can catch bugs (if it's being used)
- The next person editing this code won't come find where you live to understand it
- You catch places where you're mixing types in an unsafe way
- You can explicitly set the default value, rather than checking for undefined

Downsides:
- Constructors become gigantic, and most of it is in comments
- Takes time
- ???

Blockly's [style guide](https://developers.google.com/blockly/guides/modify/web/style-guide#jsdoc) describes what annotations should look like

blocks_vertical/default_toolbox.js
```diff
@@ -541,5 +541,12 @@ Blockly.Blocks.defaultToolbox = '<xml id="toolbox-categories" style="display: no
       '</value>'+
     '</block>'+
     '<block type="extension_music_reporter" id="extension_music_reporter"></block>'+
+    '<block type="extension_microbit_display" id="extension_microbit_display">'+
+      '<value name="MATRIX">'+
+        '<shadow type="matrix">'+
+          '<field name="MATRIX">4540074</field>'+
```
ericrosenbaum: ♥

ericrosenbaum (approved):
This looks ready to go!

Issue #1567, opened 2018-06-07
Travis Build Failure - Error: Unable to access jarfile ../webdriverio-test/selenium-server-standalone-3.0.1.jar


kchadha:
Travis build fails with error:

```
Error: Unable to access jarfile ../webdriverio-test/selenium-server-standalone-3.0.1.jar

> scratch-blocks@0.1.0 test:unit /home/travis/build/LLK/scratch-blocks
> node tests/jsunit/test_runner.js

{ Error: Couldn't connect to selenium server
    at getHTML("#closureTestRunnerLog") - test_runner.js:36:4
  message: 'connect ECONNREFUSED 127.0.0.1:4444',
  type: 'RuntimeError',
  seleniumStack: 
   { status: -1,
     type: 'ECONNREFUSED',
     message: 'Couldn\'t connect to selenium server',
     orgStatusMessage: 'connect ECONNREFUSED 127.0.0.1:4444' } }
```

kchadha:
It should be noted that #1568 was not the thing that actually fixed this issue. Restarting the build that had the described error fixed the build.

PR #1568, opened 2018-06-07
Update the chromedriver version

paulkaplan:
Update the chromedriver version requested by the test setup scripts, version 2.29 does not work with the newest version of chrome released recently and automatically included by Travis, causing the build to fail.

I'll open a follow up issue, but we should really move away from those custom scripts and just use the `chromedriver` and `selenium-webdriver` npm packages, which handle the selenium server and chromedriver path stuff for you.

Fixes https://github.com/LLK/scratch-blocks/issues/1567

If the build passes, this is good to go.


kchadha (approved):
LGTM

Issue #1563, opened 2018-06-06
Test comments in Headless Mode


kchadha:
Headless mode needs a unit test for block and workspace comments (e.g. testing text content, size, coordinates, etc.).

PR #1572, opened 2018-06-07
Add workspace comment tests and other event tests

rachel-fenichel:
### Resolves

Part of #1563 (for ws comments)

### Proposed Changes

- Test getters and setters for xy, height/width, and text content for workspace comments
- Enable events test (which had been removed, and had a util that I needed for these tests)

### Reason for Changes

Test test test
### Test Coverage

Testing 123

### DIFF
--- a/tests/jsunit/event_test.js
+++ b/tests/jsunit/event_test.js
@@ -29,15 +29,6 @@ goog.require('goog.testing.MockControl');
 
 var mockControl_;
 var workspace;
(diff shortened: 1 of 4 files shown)


kchadha (approved):
LGTM. Thanks for adding the tests! It should be noted that once these changes are in, #1575 should be rebased off of develop and the tests in here should be updated in that PR because that PR adds a parameter to the WorkspaceComment constructor to keep track of minimized state.

PR #1575, opened 2018-06-08
Workspace Comment Rendering & Minimize

kchadha:
### Resolves

`Workspace comments need to be updated to look/work like block comments` section of #1554 

### Proposed Changes

Workspace comments now look and work like block comments with respect to sizing, text size/font, minimize/delete icons and functionality.

Screenshots of the new workspace comments:

image

image

### Reason for Changes

More comments work! Compatibility w/2.0 functionality.

### Test Coverage

Tested in the vertical playground and in the scratch-gui. Added unit tests for minimize functionality.

### DIFF
--- a/core/workspace_comment_render_svg.js
+++ b/core/workspace_comment_render_svg.js
@@ -35,7 +35,7 @@ goog.require('Blockly.WorkspaceCommentSvg');
  * @const
  * @private
  */
-Blockly.WorkspaceCommentSvg.RESIZE_SIZE = 8;
+Blockly.WorkspaceCommentSvg.RESIZE_SIZE = 12 * Blockly.WorkspaceCommentSvg.BORDER_WIDTH;
 
 /**
  * Radius of the border around the comment.
@@ -43,7 +43,7 @@ Blockly.WorkspaceCommentSvg.RESIZE_SIZE = 8;
  * @const
  * @private
  */
-Blockly.WorkspaceCommentSvg.BORDER_RADIUS = 3;
+Blockly.WorkspaceCommentSvg.BORDER_WIDTH = 1;
 
 /**
  * Offset from the foreignobject edge to the textarea edge.
@@ -51,15 +51,37 @@ Blockly.WorkspaceCommentSvg.BORDER_RADIUS = 3;
  * @const
  * @private
  */
-Blockly.WorkspaceCommentSvg.TEXTAREA_OFFSET = 2;
+Blockly.WorkspaceCommentSvg.TEXTAREA_OFFSET = 12;
 
 /**
- * Offset from the top to make room for a top bar.
- * @type {number}
- * @const
+ * The height of the comment top bar.
+ * @package
+ */
+Blockly.WorkspaceCommentSvg.TOP_BAR_HEIGHT = 32;
+
+/**
+ * The size of the minimize arrow icon in the comment top bar.
  * @private
  */
-Blockly.WorkspaceCommentSvg.TOP_OFFSET = 10;
+Blockly.WorkspaceCommentSvg.MINIMIZE_ICON_SIZE = 16;
+
+/**
+ * The size of the delete icon in the comment top bar.
+ * @private
+ */
+Blockly.WorkspaceCommentSvg.DELETE_ICON_SIZE = 12;
+
+/**
+ * The inset for the top bar icons.
+ * @private
+ */
+Blockly.WorkspaceCommentSvg.TOP_BAR_ICON_INSET = 6;
+
+/**
+ * Width that a minimized comment should have.
+ * @private
+ */
+Blockly.WorkspaceCommentSvg.MINIMIZE_WIDTH = 200;
 
 /**
  * Returns a bounding box describing the dimensions of this comment.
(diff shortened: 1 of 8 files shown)


core/workspace_comment_render_svg.js
```diff
@@ -124,14 +148,14 @@ Blockly.WorkspaceCommentSvg.prototype.render = function() {
...
-    Blockly.bindEventWithChecks_(
-        this.deleteGroup_, 'mouseout', this, this.deleteMouseOut_);
-    Blockly.bindEventWithChecks_(
-        this.deleteGroup_, 'mouseup', this, this.deleteMouseUp_);
-  }
+  Blockly.bindEventWithChecks_(
+      this.minimizeArrow_, 'mousedown', this, this.minimizeArrowMouseDown_);
+  Blockly.bindEventWithChecks_(
+      this.minimizeArrow_, 'mouseup', this, this.minimizeArrowMouseUp_);
+  Blockly.bindEventWithChecks_(
```
rachel-fenichel: Why get rid of the `deleteMouseOut` event?  And the `isDeletable` check?

rachel-fenichel:
I'm reviewing now, but in the meantime can you post screenshots of this in minimized and full forms?

core/workspace_comment_render_svg.js
```diff
@@ -211,43 +239,107 @@ Blockly.WorkspaceCommentSvg.prototype.addResizeDom_ = function() {
...
+  e.stopPropagation();
+};
+
+/**
+ * Handle a mouse-up on bubble's minimize icon.
+ * @param {!Event} e Mouse up event.
+ * @private
+ */
+Blockly.WorkspaceCommentSvg.prototype.minimizeArrowMouseUp_ = function(e) {
+  this.toggleMinimize_();
```
rachel-fenichel: This function (and the delete one) mean that the action will happen on mouseup, regardless of whether the mouse down occurred on this icon.

You're also not highlighting/otherwise indicating selection on the active icon.  

If you don't want to add that here, make an issue to add that behaviour in later, to avoid unexpected deletion/minimization of comments.

core/workspace_comment_render_svg.js
```diff
@@ -276,40 +368,46 @@ Blockly.WorkspaceCommentSvg.prototype.resizeMouseDown_ = function(e) {
...
+        'xlink:href', Blockly.mainWorkspace.options.pathToMedia + 'comment-arrow-up.svg');
+    // Hide text area
+    this.commentEditor_.setAttribute('display', 'none');
+    // Hide resize handle if it exists
+    if (this.resizeGroup_) {
+      this.resizeGroup_.setAttribute('display', 'none');
+    }
+    if (labelText && this.labelText_ != labelText) {
+      // Update label and display
+      // TODO is there a better way to do this?
```
rachel-fenichel, line 438: https://github.com/LLK/scratch-blocks/commit/02da1b69712356210c7ff3e6a769e60a7880dde1

core/workspace_comment_render_svg.js
```diff
@@ -429,29 +545,63 @@ Blockly.WorkspaceCommentSvg.prototype.setSize = function(width, height) {
...
+  this.setMinimized(!this.isMinimized_);
+};
+
+/**
+ * Set the minimized state for this comment.
+ * @param {boolean} minimize Whether the comment should be minimized
+ * @package
+ */
+Blockly.WorkspaceComment.prototype.setMinimized = function(minimize) {
+  if (this.isMinimized_ == minimize) return;
```
rachel-fenichel: Put the return on a separate line, and add braces.

core/workspace_comment_render_svg.js
```diff
@@ -276,40 +368,46 @@ Blockly.WorkspaceCommentSvg.prototype.resizeMouseDown_ = function(e) {
...
- * Handle a mouse-up on comment's delete icon.
- * @param {!Event} e Mouse up event.
- * @private
- */
-Blockly.WorkspaceCommentSvg.prototype.deleteMouseUp_ = function(e) {
-  // Delete this comment
-  this.dispose(true, true);
-  // This event has been handled.  No need to bubble up to the document.
-  e.stopPropagation();
+Blockly.WorkspaceCommentSvg.prototype.setRenderedMinimizeState_ = function(minimize, labelText) {
```
rachel-fenichel, line 425: You have this function and another one (`setMinimized`) that calls it and does some related things. What's the relationship between them?  When would this one be called without calling the other?

Some of the explanation should go into the comments as well.

core/workspace_comment_render_svg.js
kchadha: Hmmm I thought that the  lack of highlight was by design, but I can check in with @carljbowman  and review what we have for both workspace/block comments after the base of these changes go in (whatever we decide that is).

I'll try to fix the delete/minimize mouse up issue you mentioned.

kchadha: It looks like this problem probably existed before these changes as well -- in the case of deletion for example, the actual dispose was happening on the mouseup event. `deleteMouseDown` and `deleteMouseOut` were just adding/removing a CSS class.

core/workspace_comment_render_svg.js
kchadha: The `deleteMouseOut` was removing a highlight that this icon never gets. This is a different delete icon from before. I will add the isDeletable check back in.

core/workspace_comment_render_svg.js
kchadha: `setRenderedMinimizeState` gets called from `setMinimized` *iff* this comment is rendered. `setMinimized` is responsible for keeping track of the minimize state regardless of whether or not this comment is rendered, and `setRenderedMinimizeState` actually changes the rendering to the minimized form (e.g. just the top bar shows, the truncated comment text in the top bar, etc.).  In block comment land, these happened in two separate classes. The equivalent of `setMinimized` happened in `Blockly.ScratchBlockComment` and the equivalent of `setRenderedMinimizeState` happened in `Blockly.ScratchBubble`.

I will add some of this explanation to the comments! =D

core/workspace_comment_render_svg.js
kchadha: Thanks for the pointer! Done (here and in ScratchBubble).

core/workspace_comment_render_svg.js
kchadha: Done.

kchadha:
@rachel-fenichel Just addressed your initial comments.

rachel-fenichel (approved):

Issue #1577, opened 2018-06-08
Add support for switching language without reloading


chrisgarrity:
- [x] script to generate a combined messages file that includes all supported locales
- [x] add language switching to playgrounds
- [x] load new translations from Transifex
- [x] refresh the workspace without reloading the page

Issue #1570, opened 2018-06-07
Add flag to disable recycling blocks in the flyout


rachel-fenichel:
Necessary changes:
- In flyout_base.js, add the property `Blockly.Flyout.prototype.recyclingEnabled_` and set it to `true` by default.
- In flyout_base.js, add the function `Blockly.Flyout.prototype.setRecyclingEnabled` to toggle the above property.  Mark it as public
- In `Blockly.Flyout.prototype.clearOldBlocks_` check `if (this.recyclingEnabled_ && block.isRecyclable())` on line 693.

To use this in scratch-gui, you'll call `workspace.getFlyout().setRecyclingEnabled(false)` before your language swap, and `workspace.getFlyout().setRecyclingEnabled(true)` after the swap.

Open question: Is it better to have the setRecyclingEnabled call on the flyout or on the whole workspace?
On the workspace:
Pro: Don't need to expose `getFlyout`.  Keep everything about the flyout package where possible.
Con: Adds another function at the workspace level, plus associated state.

On the flyout:
Pro: keep the state where it's actually being used.
Con: `getFlyout` is currently package.  Making it public exposes a large API surface area.

rachel-fenichel:
This came up when swapping the message files to change languages--refreshing the toolbox changes the messages for non-recycled blocks, but the recycled blocks stay the same.

PR #1578, opened 2018-06-08
Switch languages without reloading

chrisgarrity:
### Resolves

#1577 and #1570 I could separate the second into a separate PR, but for now it's here

### Proposed Changes
Creates and loads a combined messages file with all the translations. Switching locales overwrites `Blockly.Msg` table with new values and refreshes the workspace and toolbox. To see it work properly requireed implementing a flag to prevent the toolbox from recycling blocks.

Category labels are still an issue, but this seemed like a good time to get everything so far reviewed. Categories are a bigger can of worms.

### Reason for Changes

Supporting multiple languages

### Test Coverage

Added assertions to the scripts that process message files (js and json) to fail if there are problems. Added `npm run translate` to the test step to make sure messages are ok.


core/blockly.js
```diff
@@ -62,6 +62,7 @@ goog.require('Blockly.inject');
 goog.require('Blockly.utils');
 goog.require('goog.color');
 
+goog.require('Blockly.ScratchMsgs');
```
rachel-fenichel: Move this between `ggog.require('Blockly.Procedures');` and `goog.require('Blockly.Toolbox');`

core/scratch_msgs.js
```diff
@@ -0,0 +1,44 @@
...
+ * @fileoverview Scratch Messages singleton, with function to override Blockly.Msg values.
+ * @author chrisg@media.mit.edu (Chris Garrity)
+ */
+'use strict';
+
+/**
+ * Name space for the Msg singleton.
+ * Msg gets populated in the message files.
+ */
+goog.provide('Blockly.ScratchMsgs');
```
rachel-fenichel, line 31: Nit: swap these two lines.  Imports are ordered alphabetically,

core/scratch_msgs.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Change the Blockly.Msg strings to a new Locale
+ * Does not exist in Blockly, but needed in scratch-blocks
+ * @param {string} locale E.g., 'de', or 'zh-tw'
+ * @package
+ */
+Blockly.ScratchMsgs.setLocale = function(locale) {
+  if (Object.keys(Blockly.ScratchMsgs.locales).includes(locale)) {
+    Blockly.Msg = Blockly.ScratchMsgs.locales[locale];
+  }
```
rachel-fenichel: Consider adding an error if the locale string is not recognized.

msg/messages.js
```diff
@@ -24,6 +24,8 @@
  *
  * After modifying this file, run `npm run translate` from the root directory
  * to regenerate `./msg/json/en.json`.
+ * IMPORTANT:
```
rachel-fenichel, line 27: This file and the message files are now permanent merge conflicts with blockly.  You'll need to add them to `cleanup.sh` to make sure they're resolved correctly.

Currently `cleanup.sh` just removes (almost) all of the files in `msg/js` and `msg/json`.  You'll need to update it to resolve in favor of the scratch-blocks file.  That code is ~[here](https://github.com/LLK/scratch-blocks/blob/develop/cleanup.sh#L21).

You'll also need to add `msg/messages.js` to the explicit list of files that should always be resolved in favor of scratch-blocks.  That's [here](https://github.com/LLK/scratch-blocks/blob/develop/cleanup.sh#L80).

rachel-fenichel (commented):
Looks reasonable.  @AnmAtAnm to confirm that the message tables are being used correctly.

core/scratch_msgs.js
chrisgarrity: @rachel-fenichel really swap? the first is a `provide` not `require`. However, on looking at other files it looks like there's always a blank line between provides and requires, so I'll add that.

rachel-fenichel: You're correct, that's a provide so it's in the right place.  My bad.

msg/messages.js
chrisgarrity: I'm actually debating whether we really want to check in the js and json files. Either way, I'll fix up the cleanup script. Thanks for pointing it out.

i18n/create_scratch_msgs.js
```diff
@@ -0,0 +1,64 @@
...
+// Globals
+const PATH_INPUT = path.resolve(__dirname, '../msg/json/*.json');
+const PATH_OUTPUT = path.resolve(__dirname, '../msg');
+
+let en = fs.readFileSync(path.resolve(__dirname, '../msg/json/en.json'));
+en = JSON.parse(en);
+const enKeys = Object.keys(en).sort().toString();
+
+// Check that translation is valid:
+// elt: array [key, translation] from <locale>.json
```
AnmAtAnm: What does `elt` stand for? Please note in the comment.

i18n/create_scratch_msgs.js
```diff
@@ -0,0 +1,64 @@
...
+      assert.strictEqual(
+          tMatch.sort().toString(),
+          enMatch.sort().toString(),
+          `${key} is missing or has duplicate placeholders: ${translation}`
+      );
+    }
+};
+
+const validate = function (json, name) {
+    // this is a little stricter than we need - it would be harmless if the translation had extra keys
```
AnmAtAnm: Actually extra keys have caused real problems in the past, because later keys references do not get replaced. This is the appropriate amount of testing.

i18n/create_scratch_msgs.js
```diff
@@ -0,0 +1,64 @@
...
+    assert.strictEqual(Object.keys(json).sort().toString(), enKeys, `${name}: Locale json keys do not match en.json`);
+    Object.entries(json).forEach(validatePlaceholders);
+};
+
+let file = '';
```
AnmAtAnm: Good place to use a [multiline template string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Multi-line_strings).

msg/js/es-419.js
```diff
@@ -0,0 +1,271 @@
...
+Blockly.Msg["MOTION_SCROLLRIGHT"] = "desplazar a la derecha %1";
+Blockly.Msg["MOTION_SCROLLUP"] = "desplazar arriba %1";
+Blockly.Msg["MOTION_ALIGNSCENE"] = "alinear escena %1";
+Blockly.Msg["MOTION_ALIGNSCENE_BOTTOMLEFT"] = "abajo-izquierda";
+Blockly.Msg["MOTION_ALIGNSCENE_BOTTOMRIGHT"] = "abajo-derecha";
+Blockly.Msg["MOTION_ALIGNSCENE_MIDDLE"] = "medio";
+Blockly.Msg["MOTION_ALIGNSCENE_TOPLEFT"] = "arriba-izquierda";
+Blockly.Msg["MOTION_ALIGNSCENE_TOPRIGHT"] = "arriba-derecha";
+Blockly.Msg["MOTION_XSCROLL"] = "desplazar x";
+Blockly.Msg["MOTION_YSCROLL"] = "desplazar y
```
AnmAtAnm: newline in string leading to unterminated parse. In addition to fixing this value, double check your string escaping in script that generates these files.

msg/js/qqq.js
```diff
@@ -0,0 +1,483 @@
+// This file was automatically generated.  Do not modify.
+
+'use strict';
+
+goog.provide('Blockly.Msg.qqq');
+goog.require('Blockly.Msg');
+
+Blockly.Msg["VARIABLES_DEFAULT_NAME"] = "default name - A simple, general default name for a variable, preferably short. For more context, see [[Translating:Blockly#infrequent_message_types]].
+{{Identical|Item}}";
+Blockly.Msg["TODAY"] = "button text - Button that sets a calendar to today's date.
```
AnmAtAnm: More incorrect escaping. (Admittedly, this is just the `qqq` file).

msg/json/es-419.json
```diff
@@ -0,0 +1,265 @@
...
+    "MOTION_SCROLLRIGHT": "desplazar a la derecha %1",
+    "MOTION_SCROLLUP": "desplazar arriba %1",
+    "MOTION_ALIGNSCENE": "alinear escena %1",
+    "MOTION_ALIGNSCENE_BOTTOMLEFT": "abajo-izquierda",
+    "MOTION_ALIGNSCENE_BOTTOMRIGHT": "abajo-derecha",
+    "MOTION_ALIGNSCENE_MIDDLE": "medio",
+    "MOTION_ALIGNSCENE_TOPLEFT": "arriba-izquierda",
+    "MOTION_ALIGNSCENE_TOPRIGHT": "arriba-derecha",
+    "MOTION_XSCROLL": "desplazar x",
+    "MOTION_YSCROLL": "desplazar y\n",
```
AnmAtAnm: The source of the error above.

core/scratch_msgs.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Change the Blockly.Msg strings to a new Locale
+ * Does not exist in Blockly, but needed in scratch-blocks
+ * @param {string} locale E.g., 'de', or 'zh-tw'
+ * @package
+ */
+Blockly.ScratchMsgs.setLocale = function(locale) {
+  if (Object.keys(Blockly.ScratchMsgs.locales).includes(locale)) {
+    Blockly.Msg = Blockly.ScratchMsgs.locales[locale];
```
AnmAtAnm, line 43: I think this will work in most places. It will not work in dropdowns.  See https://github.com/google/blockly/issues/1895

msg/js/es-419.js
chrisgarrity: Good catch - I'll add a check as there is no reason for a newline in any of the msg json files.

core/scratch_msgs.js
chrisgarrity: @AnmAtAnm is there a particular type of dropdown that shouldn't work? In my vertical playground the looks effect dropdown on the `change %1 effect by %2` updates when I switch languages as does the dropdown on the `go to %1` block. However, I've noticed that we're missing some dropdown strings entirely (previous backdrop), and maybe they will be a problem.

core/blockly.js
```diff
@@ -62,7 +63,6 @@ goog.require('Blockly.inject');
 goog.require('Blockly.utils');
 goog.require('goog.color');
 
-
```
kchadha: I think the convention in the rest of the files is to have 2 newlines separating the 'goog.require' statements from the code.

core/flyout_base.js
```diff
@@ -264,6 +264,14 @@ Blockly.Flyout.prototype.dragAngleRange_ = 70;
  */
 Blockly.Flyout.prototype.scrollAnimationFraction = 0.3;
 
+/**
+ * Whether to recycle blocks when refreshing the flyout. Nothing can be
```
kchadha: I think this comment needs some clarity, because right now it seems like the second sentence is referring to the initial state of this property, when really it's referring to the fact that this initial value can't be used when the locale changes. I would just take the locale part of it out since it describes a use case for changing this property rather than a description of what this property itself does.

i18n/create_scratch_msgs.js
```diff
@@ -0,0 +1,66 @@
...
+// entry: array [key, translation]  corresponding to a single string from <locale>.json
+// - messages with placeholders have the same number of placeholders
+// - messages must not have newlines embedded
+const validateEntry = function (entry) {
+    const re = /(%\d)/g;
+    const [key, translation] = entry;
+    const placeholdersCount = en[key].match(re) ? en[key].match(re).length : 0;
+    if (placeholdersCount > 0) {
+      const tMatch = translation.match(re);
+      const enMatch = en[key].match(re);
```
kchadha: `en[key].match(re)` is referenced 3 times here. This `enMatch` declaration can move up above the `placeholdersCount` declaration.

i18n/create_scratch_msgs.js
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const validate = function (json, name) {
+    assert.strictEqual(Object.keys(json).sort().toString(), enKeys, `${name}: Locale json keys do not match en.json`);
+    Object.entries(json).forEach(validateEntry);
+};
+
+let file = `// This file was automatically generated.  Do not modify.
+
+'use strict\';
```
kchadha: Should this first `'` also be escaped? Alternatively, should none of these be escaped since we're using the `` ` `` formatted strings?

i18n/js_to_json.js
```diff
@@ -12,6 +13,7 @@ const PATH_OUTPUT = path.resolve(__dirname, '../msg/json/en.json');
 // Match function
 const match = function (str) {
     if (str.indexOf('Blockly.Msg.') !== 0) return false;
+    assert.notStrictEqual(str.indexOf("';"), str.length - 2, `[${str}] uses single quoted string, should use double quotes.`);
     if (str.indexOf('";') !== str.length - 2) return false;
```
kchadha: Should there be a similar assertion for this line? It looks like this will fail silently if " is not the second to last character in the line.

kchadha: Also, a general question, perhaps not directly related to this PR: why not use a regex for this `match` function?

E.g. something like `/^Blockly\.Msg\.*";$/`

msg/js/en.js
```diff
@@ -45,6 +45,7 @@ Blockly.Msg["DATA_INDEX_RANDOM"] = "random";
 Blockly.Msg["EVENT_WHENFLAGCLICKED"] = "when %1 clicked";
 Blockly.Msg["EVENT_WHENTHISSPRITECLICKED"] = "when this sprite clicked";
 Blockly.Msg["EVENT_WHENSTAGECLICKED"] = "when stage clicked";
+Blockly.Msg["EVENT_WHENTOUCHINGOBJECT"] = "when this sprite touches %1";
```
kchadha, line 48: Isn't this file being gitignored now? Should these changes be in this PR?

msg/json/en.json
```diff
@@ -39,6 +39,7 @@
     "EVENT_WHENFLAGCLICKED": "when %1 clicked",
     "EVENT_WHENTHISSPRITECLICKED": "when this sprite clicked",
     "EVENT_WHENSTAGECLICKED": "when stage clicked",
+    "EVENT_WHENTOUCHINGOBJECT": "when this sprite touches %1",
```
kchadha, line 42: Same question here as for `msg/js/en.js`

msg/messages.js
```diff
@@ -261,59 +263,70 @@ Blockly.Msg.SOUND_CHANGEVOLUMEBY = "change volume by %1";
...
-Blockly.Msg.CANNOT_DELETE_VARIABLE_PROCEDURE = 'Can\'t delete the variable "%1" because it\'s part of the definition of the function "%2"';
-Blockly.Msg.DELETE_VARIABLE = 'Delete the "%1" variable';
+Blockly.Msg.CHANGE_VALUE_TITLE = "Change value:";
+Blockly.Msg.RENAME_VARIABLE = "Rename variable";
+Blockly.Msg.RENAME_VARIABLE_TITLE = "Rename all '%1' variables to:";
+Blockly.Msg.RENAME_VARIABLE_MODAL_TITLE = "Rename Variable";
+Blockly.Msg.NEW_VARIABLE = "Make a Variable";
+Blockly.Msg.NEW_VARIABLE_TITLE = "New variable name:";
+Blockly.Msg.VARIABLE_MODAL_TITLE = "New Variable";
+Blockly.Msg.VARIABLE_ALREADY_EXISTS = "A variable named '%1' already exists.";
```
kchadha: Are we okay with this user visible display message quoting the selected variable name changing to use single quotes? E.g. in this message:

image

msg/scratch_msgs.js
```diff
@@ -0,0 +1,8552 @@
+// This file was automatically generated.  Do not modify.
```
kchadha, line 1: If this file is generated, should it be included in this PR?

package.json
```diff
@@ -17,9 +17,9 @@
     "test:setup": "tests/scripts/test_setup.sh",
     "test:unit": "node tests/jsunit/test_runner.js",
     "test:lint": "eslint .",
-    "test": "npm run test:lint && npm run test:setup && npm run test:unit",
+    "test": "npm run test:lint && npm run test:setup && npm run test:unit && npm run translate",
```
kchadha: Maybe the translate step should run before the unit tests (and perhaps before the setup), so that it can fail early for simple failures like using `'` instead of `"`.

tests/vertical_playground.html
```diff
@@ -482,6 +496,40 @@ <h1>Vertical Blocks</h1>
           <option value="top">Top</option>
           <option value="bottom">Bottom</option>
         </select>
+        <select name="locale" onchange="setLocale(this.value)">
```
kchadha, line 499: maybe eventually this list should be generated based on what's in `msg/js` or `msg/json`

core/scratch_msgs.js
AnmAtAnm: I'm presuming you blocks and static dropdowns use the same JSON-based loading mechanism as Blockly.  Dynamic drop downs and `block.init()` initialized fields should not have this error. There is a possibility your team has fixed the underlying bug, but I haven't heard anything like that. It comes down to how the options table is centralized for the specific dropdown of a particular type of block (so the options can be programmatically updated everywhere), and in that process the option message is dereference once, rather than per-block.

i18n/create_scratch_msgs.js
AnmAtAnm: The escape is not necessary inside a template literal (but doesn't hurt). It should be consistent, though.

msg/js/en.js
AnmAtAnm: There is a `!` in the modified `.gitignore` for `en.json` and `en.js.`.

msg/messages.js
AnmAtAnm: Good catch.  Similar quote characters are changed in numerous message string below.

Now I'm curious why we are resolving the quoting consistency problem with `"` strings instead of `'` strings, in this `.js` file.

msg/js/en.js
kchadha: Oh right! I guess I might have been responding to an offline conversation with @chrisgarrity instead of the actual .gitignore file. Oops. =]

msg/scratch_msgs.js
chrisgarrity: Having this file checked in should allow others to view the blocks in other languages. Otherwise you need credentials for Transifex to download the json files and generate it. In the same way that we check in the en.{js,json} files even though they are also generated.

tests/vertical_playground.html
chrisgarrity: I agree that would be nice to do someday.

core/scratch_msgs.js
chrisgarrity: I think all our dropdowns are initialized in the `init()` function, so maybe it's moot for scratch-blocks. There are definitely some issues with dynamic dropdowns, but they're actually getting initialized in the gui and that is what I'm working on next.

core/blockly.js
chrisgarrity: 👍 no need to introduce unnecessary changes.

msg/js/en.js
chrisgarrity: The en.{js, json} files have always been checked in, so I just tried to stay consistent.

msg/json/en.json
chrisgarrity: We agreed to leave the changes in scratch-blocks for the new event block as Eric is still hoping to get it in before the freeze. This way it'll already be in the translations.

i18n/js_to_json.js
chrisgarrity: I think it's for performance. regex is slower and the message strings do have a very predictable format.

msg/messages.js
chrisgarrity: I think it was originally because we're generating json, and that requires double quotes. Given that we mix apostrophe and double quotes in the same string, we're going to be escaping things one way or the other. I'm changing messages.js to require single quotes, but really it could go either way.

msg/json/en.json
kchadha: Ah, I wasn't actually referring to that particular line, just the same question about the en.json file as with the en.js file, which you've already answered.

msg/messages.js
```diff
@@ -24,242 +24,256 @@
  *
  * After modifying this file, run `npm run translate` from the root directory
  * to regenerate `./msg/json/en.json`.
+ * IMPORTANT:
+ * All message strings must use double quotes for the scripts to work properly
```
kchadha: This comment should be updated now that the strings have changed to use single quotes instead of double quotes.

kchadha (approved):
LGTM

PR #1580, opened 2018-06-10
Add "costume name" block

towerofnix:
### Resolves

Towards LLK/scratch-vm#355. Should be merged alongside LLK/scratch-vm#1226.

### Proposed Changes

Adds the "costume name" obsolete block:

The "costume name" looks block

The block doesn't show up in the block palette, since it's an obsolete block.

### Reason for Changes

Compatibility with projects that use the block.

### Test Coverage

No new tests, but tested manually with [this project](https://scratch.mit.edu/projects/228377453/).


apple502j:
Shouldn't it be converted to "costume (name/number)" block?

Kenny2github:
@apple502j No, the "costume name" block is an obsolete block that doesn't show up in the palette.

towerofnix:
@apple502j Yeah, I wanted to do that (map `costumeName` to `costume [name]`), but I couldn't figure out how. There doesn't seem to be a way to make the input automatically be `name`. `costumeIndex` ("costume #" in 2.0) gets mapped to `costume [index]`, but *really* it just gets mapped `costume [first dropdown item]`, and there's no way to make a mapping to `costume [second dropdown item]`. See my struggles [here](https://github.com/LLK/scratch-vm/issues/355#issuecomment-385784867). Since that comment was never replied to, I decided to just go ahead and get the "costume name" block implemented the ordinary way (by adding a new block that doesn't get shown in the block palette).

rachel-fenichel:
Just swinging through with a review...

You shouldn't need to update any of the compressed files for this (dart_compressed, javascript_compressed, lua_compressed, php_compressed, python_compressed).  Just check them out from develop: 

With develop up to date, run

`git checkout develop -- dart_compressed.js javascript_compressed.js lua_compressed.js php_compressed.js python_compressed.js`

towerofnix:
@rachel-fenichel Woops, right. Thanks. Usually I don't add those but forgot them this time around.

(Edit: fixed merge issues...looks like my develop branch was a bit bork'd.)

kchadha:
@towerofnix, re: your comment:
> @apple502j Yeah, I wanted to do that (map costumeName to costume [name]), but I couldn't figure out how. There doesn't seem to be a way to make the input automatically be name. costumeIndex ("costume #" in 2.0) gets mapped to costume [index], but really it just gets mapped costume [first dropdown item], and there's no way to make a mapping to costume [second dropdown item]. See my struggles here. Since that comment was never replied to, I decided to just go ahead and get the "costume name" block implemented the ordinary way (by adding a new block that doesn't get shown in the block palette).

This can be done in the scratch-vm. The same thing is being done for the Scratch 2.0 `backdrop name` block here:

https://github.com/LLK/scratch-vm/blob/develop/src/serialization/sb2.js#L945-L950

You can add a similar case to that switch statement for the costume name block, and then the sb2_specmap translation for the obsolete `costumeName` opcode can translate to the 3.0 `looks_costumenumbername` opcode, the same way `sceneName` gets translated.

I think that change would eliminate the need for this pull request as well.

PR #1586, opened 2018-06-13
Add extension status button stub

ericrosenbaum:
### Proposed Changes

Add a button that scratch extensions can optionally show. This version is not yet styled- it uses the basic flyout button class. In the future I'll make a subclass that can display a status button to spec (a small image button inline with the category label, which can change state to show different images).  

### Reason for Changes

Work in progress on supporting hardware extensions, where the user needs to be able to initiate a bluetooth connection workflow.

### DIFF
--- a/core/flyout_base.js
+++ b/core/flyout_base.js
@@ -536,6 +536,22 @@ Blockly.Flyout.prototype.show = function(xmlList) {
         } else {
           gaps.push(default_gap);
         }
+      } else if ((tagName == 'BUTTON') &&
+        (xml.getAttribute('type') === 'status')) {
+        // Scratch extensions can display a status button
+        var extensionId = xml.getAttribute('extensionId');
+        if (extensionId) {
+          var callbackKey = 'STATUS_PRESSED_' + extensionId.toUpperCase();
+          xml.setAttribute('callbackKey', callbackKey);
+          var callback = function(id) {
+            Blockly.statusButtonCallback(id);
+          }.bind(this, extensionId);
+          this.workspace_.registerButtonCallback(callbackKey, callback);
+          // @todo: make a new image button class to use here
+          var curButton = new Blockly.FlyoutButton(this.workspace_,
+              this.targetWorkspace_, xml, false);
+          contents.push({type: 'button', button: curButton});
+        }
       } else if (tagName == 'BUTTON' || tagName == 'LABEL') {
         // Labels behave the same as buttons, but are styled differently.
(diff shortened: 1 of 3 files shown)


core/flyout_base.js
```diff
@@ -536,6 +536,22 @@ Blockly.Flyout.prototype.show = function(xmlList) {
         } else {
           gaps.push(default_gap);
         }
+      } else if ((tagName == 'BUTTON') &&
+        (xml.getAttribute('type') === 'status')) {
```
kchadha: The scratch-blocks convention is to use `==`

kchadha: Also does this need to be on two separate lines? If so, you need to indent an extra 2 spaces on this second line.

core/flyout_base.js
```diff
@@ -536,6 +536,22 @@ Blockly.Flyout.prototype.show = function(xmlList) {
...
+        // Scratch extensions can display a status button
+        var extensionId = xml.getAttribute('extensionId');
+        if (extensionId) {
+          var callbackKey = 'STATUS_PRESSED_' + extensionId.toUpperCase();
+          xml.setAttribute('callbackKey', callbackKey);
+          var callback = function(id) {
+            Blockly.statusButtonCallback(id);
+          }.bind(this, extensionId);
```
kchadha: Do you need the extra wrapper? Can this line be:

```
var callback = Blockly.statusButtonCallback.bind(this, extensionId);
```

Or does that not work...?

kchadha (changes requested):
Had some comments and questions.

core/flyout_base.js
ericrosenbaum: 👍

core/flyout_base.js
ericrosenbaum: 👍

core/flyout_base.js
```diff
@@ -536,6 +536,22 @@ Blockly.Flyout.prototype.show = function(xmlList) {
         } else {
           gaps.push(default_gap);
         }
+      } else if ((tagName == 'BUTTON') &&
```
rachel-fenichel: Why is this not a nested if inside the tagName == BUTTON || tagName == LABEL check?

ericrosenbaum: I felt this way was clearer, because this part of the code will soon change and create a different type of button.

kchadha (approved):
LGTM

PR #1587, opened 2018-06-14
Workspace scroll bugfix and unit tests!

kchadha:
### Resolves

`Workspace scroll is wonky if block comments are on the screen. Scroll gets pinned to a corner of the workspace.` section of #1554.

### Proposed Changes

ScratchBlockComments should be able to handle `NaN` values for x and y positions, and their x and y positions properties should be updated after auto-positioning the comment (even if they are not rendered).

### Reason for Changes

The workspace metrics get wonky if scratch block comments are not correctly reporting positioning information.

### Test Coverage

Added unit tests! Particularly, `test_blockCommentXYNaNPositionProvided` and `test_blockCommentXYNullPositionProvided` fail prior to this fix and pass post changes.

### DIFF
--- a/core/scratch_block_comment.js
+++ b/core/scratch_block_comment.js
@@ -57,13 +57,18 @@ Blockly.ScratchBlockComment = function(block, text, id, x, y, minimized) {
    * @private
    */
   this.text_ = text;
+
+  var xIsValidNumber = typeof x == 'number' && !isNaN(x);
+  var yIsValidNumber = typeof y == 'number' && !isNaN(y);
+
   /**
    * Whether this comment needs to be auto-positioned (based on provided values
    * for x and y position).
    * @type {boolean}
    * @private
    */
-  this.needsAutoPositioning_ = !x && x != 0 && !y && y != 0;
+  this.needsAutoPositioning_ = !xIsValidNumber && !yIsValidNumber;
+  // If both of the given x and y params are invalid, this comment needs to be auto positioned.
 
   /**
    * The x position of this comment in workspace coordinates. Default to 0 if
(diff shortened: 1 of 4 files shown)


core/scratch_block_comment.js
```diff
@@ -58,17 +58,27 @@ Blockly.ScratchBlockComment = function(block, text, id, x, y, minimized) {
    */
   this.text_ = text;
   /**
-   * The x position of this comment in workspace coordinates.
+   * Whether this comment needs to be auto-positioned (based on provided values
+   * for x and y position).
+   * @type {boolean}
+   * @private
+   */
+  this.needsAutoPositioning_ = !x && x != 0 && !y && y != 0;
```
paulkaplan: Is there a value in particular that causes this? Is it always null or NaN? this is a pretty hard to parse line, could be simplified if you knew it was always null/NaN and just checked for that. But not sure if there is really a better way.

core/scratch_block_comment.js
```diff
@@ -58,17 +58,27 @@ Blockly.ScratchBlockComment = function(block, text, id, x, y, minimized) {
...
+  this.x_ = typeof x == 'number' && !isNaN(x) ? x : 0;
   /**
-   * The y position of this comment in workspace coordinates.
+   * The y position of this comment in workspace coordinates. Default to 0 if
+   * y position is not provided or is not a valid number.
    * @type {number}
    * @private
    */
-  this.y_ = y;
+  this.y_ = typeof y == 'number' && !isNaN(y) ? y : 0;
```
paulkaplan: Could you do this check on x/y above and reuse the result of `typeof y == 'number' && !isNaN(y)` to make the line I commented on above more clear?

paulkaplan (approved):
I think this LG! I think @rachel-fenichel might have some other comments, but this feels good enough to go in.

rachel-fenichel:
Yeah, LG.

core/scratch_block_comment.js
kchadha: I want the flag to be set if both x and y are any of: NaN, null, or undefined

kchadha:
Woo! I'll wait for the tests and then merge. =D

Issue #1592, opened 2018-06-15
Comment Cleanup


kchadha:
There is currently a lot of code duplication between ScratchBlockComment, ScratchBubble and WorkspaceComment*

It would be good to clean all of this up and break out common functionality/constants.

cwillisf:
scratch-blocks v2 is a complete rewrite built on top of upstream [Blockly](https://github.com/RaspberryPiFoundation/blockly/). The internal code referenced in this issue no longer exists. Thank you for reporting it!

If you believe this issue is still relevant to scratch-blocks v2, please let us know and we can reopen it.

PR #1593, opened 2018-06-15
Add a character limit to the comment textareas.

kchadha:
Title says it all.

There is some duplication here, but I would rather address that for all the comment constants in a future PR (tracking that issue in #1592).


rachel-fenichel:
Not over 9000 :(

You can get over the limit when loading from XML.  Do you care?

LGTM.

PR #1597, opened 2018-06-18
Add flyout image button and flyout status button

ericrosenbaum:
### Proposed Changes

Further work on the status button for extensions to display in the flyout. 

- Add a FlyoutImageButton class that inherits from FlyoutButton, but displays an image.
- Add a FlyoutStatusButton class that inherits from FlyoutImageButton, but has a setStatus function to be used by the extension to update the button's image.
- Add status enums for 'ready' and 'not ready' states.
- Add a top-level function to update a status button with a status enum and an id (we'll use the scratch extension id).

One remaining chunk of work for a future PR is to position the status button to spec.

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -381,6 +381,18 @@ Blockly.prompt = function(message, defaultValue, callback, _opt_title,
  */
 Blockly.statusButtonCallback = function(id) {
   window.alert('status button was pressed for ' + id);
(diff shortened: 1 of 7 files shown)


ericrosenbaum:
Replacing this with https://github.com/LLK/scratch-blocks/pull/1610

PR #1598, opened 2018-06-18
Enable space-infix-ops rule

rachel-fenichel:
### Proposed Changes

Enable [this rule](https://eslint.org/docs/rules/space-infix-ops) for eslint.

### Reason for Changes

We're consistent about this rule (only 9 violations in the codebase) so we should formalize + automate checking it.

### Test Coverage

I ran `eslint core` from the base on the blockly directory.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -41,6 +41,7 @@
         "quotes": ["off"], # Blockly mixes single and double quotes
         "semi": ["error", "always"],
         "space-before-function-paren": ["error", "never"], # Blockly doesn't have space before function paren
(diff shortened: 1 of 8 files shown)


kchadha:
@rachel-fenichel Looks like there are conflicts with the default_toolbox file. Other than that, looks good!

rachel-fenichel:
Ah, because Andrew updated it in a recent PR.  I'll update it and merge when that's settled.

kchadha:
@rachel-fenichel there's still a merge conflict, btw

PR #1610, opened 2018-06-25
Add flyout extension category header

ericrosenbaum:
Adds a new class that extends Flyout Button called FlyoutExtensionCategoryHeader. 

If an extension category XML includes `showStatusButton = "true"`, then instead of a label at the top of the category, put this new header that includes the text label and a button that serves both as a status indicator for device connectivity, and a button to launch the Scratch GUI's device connection modal. There's a global function updateStatusButton that the GUI will use to update it.

As part of the work I modified FlyoutButton to use a helper function to add its text SVG.

Notes:

- I tried but was not quickly able to make the new text SVG of the label layout exactly right in RTL, so that will have to wait for future work. 
- Ideally this header would support additional buttons (which we may wish to add in the future), enabled by additional flags in the XML, but that will also have to wait for future work.
- No alt text on the button images yet

### DIFF
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -381,6 +381,23 @@ Blockly.prompt = function(message, defaultValue, callback, _opt_title,
  */
 Blockly.statusButtonCallback = function(id) {
   window.alert('status button was pressed for ' + id);
+  Blockly.updateStatusButton(id, Blockly.StatusButtonState.READY);
+};
+
+/**
+ * Update the visual state of a status button in an extension category header.
+ * @param {string} id An extension id.
+ * @param {Blockly.StatusButtonState} newStatus the new state for the status button.
+ */
+Blockly.updateStatusButton = function(id, newStatus) {
+  var buttons = this.getMainWorkspace().getFlyout().buttons_;
+  for (var i = 0; i < buttons.length; i++) {
+    if (buttons[i] instanceof Blockly.FlyoutExtensionCategoryHeader) {
+      if (buttons[i].extensionId == id) {
+        buttons[i].setStatus(newStatus);
+      }
+    }
+  }
 };
 
 /**
--- a/core/blockly.js
+++ b/core/blockly.js
@@ -381,16 +381,16 @@ Blockly.prompt = function(message, defaultValue, callback, _opt_title,
  */
 Blockly.statusButtonCallback = function(id) {
   window.alert('status button was pressed for ' + id);
-  Blockly.updateStatusButton(id, Blockly.StatusButtonState.READY);
 };
 
 /**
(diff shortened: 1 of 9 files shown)


core/flyout_extension_category_header.js
```diff
@@ -0,0 +1,157 @@
...
+goog.require('Blockly.FlyoutButton');
+
+/**
+ * Class for a category header in the flyout for Scratch extensions which can
+ * display a textual label and a status button.
+ * @param {!Blockly.WorkspaceSvg} workspace The workspace in which to place this
+ *     header.
+ * @param {!Blockly.WorkspaceSvg} targetWorkspace The flyout's target workspace.
+ * @param {!Element} xml The XML specifying the header.
+ * @constructor
```
rachel-fenichel: @extends {Blockly.FlyoutButton}

core/flyout_extension_category_header.js
```diff
@@ -0,0 +1,157 @@
...
+ */
+Blockly.FlyoutExtensionCategoryHeader = function(workspace, targetWorkspace, xml) {
+  /**
```
rachel-fenichel: Much of this constructor code can be shared with FlyoutButton.

core/blockly.js
```diff
@@ -381,6 +381,23 @@ Blockly.prompt = function(message, defaultValue, callback, _opt_title,
  */
 Blockly.statusButtonCallback = function(id) {
   window.alert('status button was pressed for ' + id);
+  Blockly.updateStatusButton(id, Blockly.StatusButtonState.READY);
+};
+
+/**
+ * Update the visual state of a status button in an extension category header.
```
rachel-fenichel: Now that you're not passing in a callback key, registering callbacks, etc., this doesn't need to be in blockly.js.  You can probably move this to your new file, skipping the need to go through the main workspace->flyout->button list, and delete statusButtonCallback entirely.

ericrosenbaum: We're overriding statusButtonCallback in the GUI [here](https://github.com/LLK/scratch-gui/blob/7fc01216d959e63ad70952123162b57420d34daf/src/containers/blocks.jsx#L61
), so I think it makes sense to keep that in blockly.js,

We're not using the updateStatusButton function yet, but that will also need to get called from the GUI.

rachel-fenichel: Okay, that makes sense.

Next question: can you pass a workspace into `updateStatusButton` instead of using `getMainWorkspace`?  It's better form not to use that function in new code--it's a holdover from when you could only have one blockly workspace on a page.

rachel-fenichel (approved):

PR #1621, opened 2018-07-04
Local Variables

kchadha:
### Resolves

Towards allowing creating local variables.
Towards resolving LLK/scratch-vm#1278.

### Proposed Changes

Allow for local variables by passing in extra scope information (provided by the gui). This information is included in the VarCreate event. Pass in a list of additional var names to check against when creating a variable (e.g. local var names from all sprites when checking for name conflicts to create a global var.

### Reason for Changes

Allowing creation of local variables (providing the relevant information to the VM through the `var_create` event).

#### Related PRs
This PR should be merged first.

This PR is also related to the following PRs:
LLK/scratch-vm#1301
LLK/scratch-gui#2548

### DIFF
--- a/core/variable_events.js
+++ b/core/variable_events.js
@@ -88,6 +88,7 @@ Blockly.Events.VarCreate = function(variable) {
   Blockly.Events.VarCreate.superClass_.constructor.call(this, variable);
   this.varType = variable.type;
   this.varName = variable.name;
+  this.isLocal = variable.isLocal;
 };
 goog.inherits(Blockly.Events.VarCreate, Blockly.Events.VarBase);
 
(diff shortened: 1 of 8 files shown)


paulkaplan:
As we talked about offline, this looks good but I think we should include xml serialization in this so that we don't end up "lying" to scratch blocks about global/local variables, so in the future if we do want to use this information (e.g. for rendering global vars differently) it doesn't turn out to be wrong. This will also require an update to the linked VM issue. Even though it isn't strictly needed, we should make sure to keep the data model clean and correct.

kchadha:
@paulkaplan I made the serialization change and another change as well to fix variables in the vertical playground (they weren't working because additionalVars was not getting initialized without the gui connection).

PR #1640, opened 2018-07-11
Add "delete all of [list]" block

thisandagain:
### Proposed Changes
Adds "delete all of [list]" block as per discussion.

### Reason for Changes
Maintaining and further exposing useful functionality from Scratch 2.0

### Reference
load


kyleplo:
Could you add "first", "last" and "random" options?

mrjacobbloom:
+1 for a "last" option, `delete (length of [list v]) of [list v]` is clunky

Kenny2github:
@kyleplo @mrjacobbloom um, this is a static block to delete *all* list items; first, last, and random will be separate blocks AFAIK.

towerofnix:
@Kenny2github Sure, but I think their suggestion here is to make this into a less-static block - that is, a block with a dropdown "all", "first", "last", "random". Like the "go forward/backward 1 layers" looks block.

gnarf:
Hrm - so the current scratch 2 has `delete (#/last/all) of [list]` -- is this just to make the options more available since we don't have the dropdown/shadow for last and all options?

image

https://github.com/LLK/scratch-vm/blob/develop/src/blocks/scratch3_data.js#L125-L137

It seems the current `delete [item] of [list]` block also takes `all`, `last`, `random` and `any` -- would we be changing these blocks on import from scratch 2 to be new "delete all" "delete last" and "delete random" blocks?

https://github.com/LLK/scratch-vm/blob/develop/src/util/cast.js#L168-L190

gnarf:
> +1 for a "last" option, delete (length of [list v]) of [list v] is clunky

@mrjacobbloom - It's hidden but you can currently just `delete "last" of [list v]` (put "last" in as the value and it works in scratch 3

gnarf:
you can also use "random" "any" or "last" in the `item [#] of [list]` blocks

chrisgarrity (commented):
Changes look fine for localization, and I checked switching languages in the vertical_playground.

thisandagain:
> The current scratch 2 has delete (#/last/all) of [list] -- is this just to make the options more available since we don't have the dropdown/shadow for last and all options?

Correct, we are adding this block to resolve the lack of "hybrid" dropdown + input fields in Scratch 3.0. The other options (e.g. #, last, first, random) are all easily achievable with a few additional reporters, but deleting everything from a list is a hassle so we are adding this block to help resolve the issue.

kchadha (approved):
LG!

PR #1641, opened 2018-07-12
Separate updating translations from generating src.

chrisgarrity:
* setLocale uses Object.assign so any new keys in Blockly.Msg are maintained after changing locale. I.e. if a translation doesn’t exist in the translation English default will be used. 
In theory, it could end up with something other than English if the key is in one translation, but not another. However, it’s unlikely as we’re pulling the translations from transifex and either they will all have the key (if the key exists in the `en.json` source on transifex), or none of them will have the key.

* translate script only generates `en` sources.
* translate:update script generates new scratch_msgs.js file. It will fail if the keys in the downloaded files don’t match the ones in messages.


thisandagain (approved):
🎉 Thank you!

PR #1644, opened 2018-07-12
Rename Variable Bugfix

kchadha:
### Proposed Changes

Fix issue with renameVariable which didn't know about additionalVars. Decide whether or not to use additionalVars based on the new variable isLocal flag (additionalVars should only be used to check for conflicts if the variable is global).

### Related PRs
LLK/scratch-gui#2610

### Test Branch
https://llk.github.io/scratch-gui/rename-var-bugfix-travis


paulkaplan (approved):
LG

PR #1646, opened 2018-07-16
Issue/widen category

chrisgarrity:
### Resolves

Widen the category menu to support more languages.

### Proposed Changes

Increases the width of the category menu by 8px. This is a compromise. There are still a few languages with category names that get cut off, but it is wide enough for more of them to fit without shrinking the fontsize further.

### Reason for Changes

Localization.

### Test Coverage

Testing needs to happen in GUI because category labels are not localized in the playground.


Kenny2github:
謝謝！ 😄

PR #1648, opened 2018-07-16
Issue/1618 split nextbackdrop

### LINKED ISSUE #1618: Have two versions of 'next backdrop' for translation
Consider splitting `next backdrop` string into two strings. One for the block, and the other for the menu dropdown. 

They don't necessarily get translated the same way for all languages.

chrisgarrity:
### Resolves

resolves #1618 

### Proposed Changes

Allows for two translations of 'next backdrop' one for the block, and the other for the menu,

### Reason for Changes

In some languages the same text cannot be used in both places.

Note - the second commit is just updated translations, so no need to review that part.


paulkaplan (approved):
LG

PR #1649, opened 2018-07-17
Update motion.js

a49594a:
Blockly.Msg.MOTION_TURNRIGHT&Blockly.Msg.MOTION_TURNLEFT was in the wrong place

### Resolves

### Proposed Changes

### Reason for Changes

### Test Coverage

### DIFF
--- a/blocks_vertical/motion.js
+++ b/blocks_vertical/motion.js
@@ -55,7 +55,7 @@ Blockly.Blocks['motion_turnright'] = {
    */
   init: function() {
     this.jsonInit({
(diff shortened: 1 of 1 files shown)


thisandagain:
@paulkaplan Can you take a look at this?

paulkaplan (approved):
Yup this is right. It never mattered because neither turn left nor turn right strings actually mention left or right, because they use the image (both are = "turn %1 %2 degrees" in english, %1 being the icon). 

@chrisgarrity we might want to mention this change to translators, although I don't think the direction should be in the translation, similar to english.

PR #1651, opened 2018-07-18
Log warning instead of throwing error on variable conflicts

kchadha:
### Resolves

Prevents variable conflicts from causing user-facing possibly fatal errors.

### Proposed Changes

Log a warning when a variable conflict occurs instead of throwing an error. When an error is thrown, it makes blocks in the project disappear from the workspace or eventually makes whole projects crash. Instead we should log a warning so that the situation causing the variable conflict can be noted and resolved in scratch-vm.

### Reason for Changes

Helps to lessen the repercussions of variable conflicts that are not yet resolved e.g. in the following two issues:

LLK/scratch-gui#2180
LLK/scratch-gui#2094

### Test Coverage

Manually tested the two issues mentioned above.

### DIFF
--- a/core/variable_map.js
+++ b/core/variable_map.js
@@ -182,15 +182,22 @@ Blockly.VariableMap.prototype.createVariable = function(name,
   var variable = this.getVariable(name, opt_type);
   if (variable) {
     if (opt_id && variable.getId() != opt_id) {
-      throw Error('Variable "' + name + '" is already in use and its id is "'
+      // There is a variable conflict. Variable conflicts should be eliminated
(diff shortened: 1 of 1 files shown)


thisandagain (approved):
Looks good. Successfully ran test cases from issue https://github.com/LLK/scratch-gui/issues/2180 and https://github.com/LLK/scratch-gui/issues/2094 as well as our "all the blocks" smoke testing project.

rachel-fenichel:
Does this mean that any error that scratch-blocks throws causes block disappearances and fatal crashes, or is it specific to this particular error?

PR #1657, opened 2018-07-21
Separate local and global variables in block palette

### LINKED ISSUE #1636: Local variables should all be sorted after global variables
### Expected Behavior

Local variables should, in the block palette, be sorted so that they all appear (in alphanumeric order) after all the global variables.

### Actual Behavior

Local variables and global variables are sorted together.

### Steps to Reproduce

Create a global variable such as "A". Create a local variable such as "B". Create a global variable such as "C". Note that the sort order is A, B (local), C, not A, C, B (local).

---

This sort of ties into #1379. I believe in 2.0 all variables are sorted together, so I suppose this is a request more than a compatibility issue. It should be noted that Scratch 1.4 has a visible horizontal-rule separator between global variables and local variables:

A, C; separator line; B. Scratch 1.4.

I don't think scratch-blocks has support for `<hr>`-style separators, but it does have ordinary separator spaces (e.g. between "point towards mouse-pointer" and "change x by 10"). Could be useful here.

towerofnix:
Resolves #1636.

The Variables category, showing: "For all sprites: Global A, Global B. For this sprite only: Local A, Local B."

This PR will still need a bit of work, but here is a first draft, to prompt code and design discussion.

### Design question

**Which labels should show up?** Here are the two main options to consider:

1. Given there is a global variable, only show the "For all sprites" label when there is *also* at least one local variable.
2. Make labels show up according to whether there are any variables of the particular scope. That is, if there are any local variables, show the label "For this sprite only", and if there are any global variables, show the label "For all sprites".
3. (Alternatively, always show both labels, alongside the text "None" when there are no variables of a particular scope, or never show either label, as in Scratch 1.4.)

Those (first) two options basically combat each other in a crucial way. The first takes the perspective of, "Don't expose the user to potentially confusing/overwhelming behavior." The second takes, "Do expose the user to potentially interesting/useful/educational behavior." To make these clear, see a case of a beginning user creating a variable. They click on "Make a Variable", type a name (like Fruit), and leave "For all sprites" selected, since it's the default one and they aren't sure what the difference between the options is. Now assume we go with option 2 above. They will, immediately and throughout working on their project, see the label "For all sprites". They might respond in one of three ways:

* Confusion: Why did I make this sprite "for all sprites"? Does that matter? Should I be worried? Do I have to know the difference between "for all sprites" and "for this sprite only"?
* Interest: All the same questions as above, but in an "I want to learn" / "I am curious" perspective!
* Disregard: Just ignoring it.

I'm inclined towards the second option, because I think the vast majority of users will either ignore the label or be interested in the difference between "for all sprites" and "for this sprite only". I suppose a few might get confused or overwhelmed by the new label (maybe users with experience in 2.0 who never even paid attention to the local/global option). But it's not really that much of an extra load in the UI - that is, to the human using the UI - and it has a fairly obvious meaning, seeing as it's identical in language to the choice selected in the "make a variable" prompt.

As I said, I think option 2 is best here, but I'm interested in other opinions, if you have one.

### Definitely TODOs

- [ ] Figure out how to deal with the "For this sprite only" and "For all sprites" messages, regarding localization. These are already defined in scratch-gui, but I'm not sure how to access that same translation from scratch-blocks. Temporarily, I've added some messages to `msg/messages.js`.
- [x] Clean up and separate the code for building the variable blocks, in such a way that it's easy to...
- [x] Also do all this for list blocks.
- [ ] In [the screenshot](https://user-images.githubusercontent.com/9948030/43030874-3d9e67ba-8c6d-11e8-814f-9d960eb19aaa.png), the gap between "Global B" and "For this sprite only" seems to be smaller than the one between "Local B" and "set Global A to". This should probably be the other way around.

### DIFF
--- a/core/data_category.js
+++ b/core/data_category.js
@@ -42,13 +42,39 @@ goog.require('Blockly.Workspace');
  */
 Blockly.DataCategory = function(workspace) {
   var variableModelList = workspace.getVariablesOfType('');
-  variableModelList.sort(Blockly.VariableModel.compareByName);
+
+  var globalVariableList = variableModelList.filter(function(entry) {
+    return !entry.isLocal;
+  });
+  var localVariableList = variableModelList.filter(function(entry) {
+    return entry.isLocal;
+  });
+
+  globalVariableList.sort(Blockly.VariableModel.compareByName);
+  localVariableList.sort(Blockly.VariableModel.compareByName);
+
   var xmlList = [];
 
   Blockly.DataCategory.addCreateButton(xmlList, workspace, 'VARIABLE');
 
-  for (var i = 0; i < variableModelList.length; i++) {
-    Blockly.DataCategory.addDataVariable(xmlList, variableModelList[i]);
+  if (globalVariableList.length > 0) {
+    Blockly.DataCategory.addLabel(xmlList, Blockly.Msg.FOR_ALL_SPRITES + ':');
+  }
+
+  for (var i = 0; i < globalVariableList.length; i++) {
+    Blockly.DataCategory.addDataVariable(xmlList, globalVariableList[i]);
+  }
+
+  if (globalVariableList.length > 0) {
+    xmlList[xmlList.length - 1].setAttribute('gap', 24);
+  }
+
+  if (localVariableList.length > 0) {
+    Blockly.DataCategory.addLabel(xmlList, Blockly.Msg.FOR_THIS_SPRITE_ONLY + ':');
+  }
+
+  for (var i = 0; i < localVariableList.length; i++) {
+    Blockly.DataCategory.addDataVariable(xmlList, localVariableList[i]);
   }
 
   if (variableModelList.length > 0) {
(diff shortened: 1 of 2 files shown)


core/data_category.js
```diff
@@ -439,6 +465,18 @@ Blockly.DataCategory.addBlock = function(xmlList, variable, blockType,
...
 };
 
+/**
+ * Construct a flyout label with the given text tag. Add the label to the given
+ *     xmlList.
+ * @param {!Array.<!Element>} xmlList Array of XML block elements.
+ * @param {string} text Text content of the label.
+ */
+Blockly.DataCategory.addLabel = function(xmlList, text) {
+  var labelText = '<xml><label text="' + text + '"></label></xml>';
```
towerofnix, line 443: This feels vulnerable to nonsense similar to XSS, e.g. if a translator put `"></label>` in the translation string. Is `setAttribute` better here, i.e. `labelText = '<xml></label></label></xml>'` and `label.setAttribute('text', text)`?

paulkaplan:
Thanks so much for putting in all this thought on the issue! We used your demo as a starting point for discussion, just like you mentioned, and here is what we talked about: 
1. We need to do a bit of work in understanding how people are actually using the local variables, and to see how the confusion comes up. Since it is a fairly high-ceiling feature, scratchers often seem to get around it by naming their variables something like "local ___". 
2. We have in the past decided against adding more hierarchy to the toolbox (e.g. trying to avoid having nested categories), and this style would introduce another layer of hierarchy into this generally flat and physically small interface. 
3. Would introducing a distinction require an extra distinction for cloud variables? 
4. Other alternatives could be indicating the difference on the reporter blocks themselves, or together in the text on the block.

With those things in mind, I think I'll close this PR, repost some of this discussion/screenshot to the original image, and then schedule a feature time to look more into this.

PR #1661, opened 2018-07-23
When opt_noId is specified to true, shadow ids should also be removed.

### LINKED ISSUE #1389: Blocks disappear after sharing the love more than once and deleting one of the shared blocks.
### Expected Behavior

A sprite's blocks should not disappear.

### Actual Behavior

Sometimes a sprite's blocks disappear after a very specific combination of steps. This has to do with  block shadows not being copied properly.

### Steps to Reproduce
Below are two different ways to repro this issue: 

1. Create a project with two sprites `A` and `B`.
2. Drag out a block with a droppable input on `A` (e.g. move 10 steps)
3. Plug another block into that input
4. share the love of those two connected blocks from A to B 
5. share the love of those two connected blocks from A to B again
6. delete any of the blocks on B
7. Switch sprites and switch back.
8. Observe errors in the console.

OR

Replace steps 5 and 6 with:
   5. Share the love of the shared blocks back from B to A
   6. Delete any of the blocks on A

kchadha:
### Resolves

Resolves LLK/scratch-vm#1389

### Proposed Changes

`blockToDom` was not correctly removing ids from the entire block tree when `opt_noId` was specified (e.g. dragging a block outside the workspace). In particular, shadows were being serialized to XML with their IDs. This was resulting in a bug where sharing the love of the same block more than once would result in multiple stacks of blocks referencing the same shadow block. Deleting any of these stacks would result in crashing the workspace.

These changes remove the IDs from the shadow dom being serialized if the `opt_noId` flag is specified.

### Test Coverage

Manual testing in the vertical_playground and the scratch-gui to ensure that shadow IDs are no longer being preserved and the bug referenced in LLK/scratch-vm#1389 is resolved.

### DIFF
--- a/core/xml.js
+++ b/core/xml.js
@@ -213,7 +213,13 @@ Blockly.Xml.blockToDom = function(block, opt_noId) {
       }
       var shadow = input.connection.getShadowDom();
       if (shadow && (!childBlock || !childBlock.isShadow())) {
-        container.appendChild(Blockly.Xml.cloneShadow_(shadow));
+        var shadowClone = Blockly.Xml.cloneShadow_(shadow);
+        // Remove the ID from the shadow dom clone if opt_noId
+        // is specified to true.
+        if (opt_noId && shadowClone.getAttribute('id')) {
+          shadowClone.removeAttribute('id');
+        }
+        container.appendChild(shadowClone);
       }
       if (childBlock) {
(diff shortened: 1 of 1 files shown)


kchadha:
\cc @rachel-fenichel

rachel-fenichel:
Makes sense/lgtm.

PR #1664, opened 2018-07-24
Throw Errors instead of using goog.asserts

rachel-fenichel:
Not urgent to review or merge this week.

### Resolves

None

### Proposed Changes

Mostly just pulling in these changes: https://github.com/google/blockly/commit/3909bd420a22fec4cc830b0d7a31e75c5d0ff244

### Reason for Changes

Blockly is moving away from using closure library code where possible.  We're starting with the easy cases/places where modern browsers standardized so there's no need to hide special casing behind function calls.  The eventual goal is to remove the UI widgets and all other dependencies as well.

This change keeps scratch blocks in line with some of the recent blockly work.  There will be other related changes.

### Test Coverage

I built and played with the vertical playground.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -40,11 +40,11 @@ goog.require('Blockly.FieldLabelSerializable');
 goog.require('Blockly.FieldVariableGetter');
 goog.require('Blockly.Input');
 goog.require('Blockly.Mutator');
+goog.require('Blockly.utils');
 goog.require('Blockly.Warning');
 goog.require('Blockly.Workspace');
 goog.require('Blockly.Xml');
-goog.require('goog.array');
-goog.require('goog.asserts');
+
 goog.require('goog.math.Coordinate');
 goog.require('goog.string');
 
@@ -174,15 +174,16 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
     /** @type {string} */
     this.type = prototypeName;
     var prototype = Blockly.Blocks[prototypeName];
-    goog.asserts.assertObject(prototype,
-        'Error: Unknown block type "%s".', prototypeName);
+    if (!prototype || typeof prototype != 'object') {
+      throw TypeError('Unknown block type: ' + prototypeName);
+    }
     goog.mixin(this, prototype);
   }
 
   workspace.addTopBlock(this);
 
   // Call an initialization function, if it exists.
-  if (goog.isFunction(this.init)) {
+  if (typeof this.init == 'function') {
     this.init();
   }
   // Record initial inline state.
@@ -205,7 +206,7 @@ Blockly.Block = function(workspace, prototypeName, opt_id) {
 
   }
   // Bind an onchange function, if it exists.
-  if (goog.isFunction(this.onchange)) {
+  if (typeof this.onchange == 'function') {
     this.setOnChange(this.onchange);
   }
 };
@@ -546,14 +547,14 @@ Blockly.Block.prototype.setParent = function(newParent) {
   }
   if (this.parentBlock_) {
     // Remove this block from the old parent's child list.
-    goog.array.remove(this.parentBlock_.childBlocks_, this);
+    Blockly.utils.arrayRemove(this.parentBlock_.childBlocks_, this);
 
     // Disconnect from superior blocks.
     if (this.previousConnection && this.previousConnection.isConnected()) {
-      throw 'Still connected to previous block.';
+      throw Error('Still connected to previous block.');
     }
     if (this.outputConnection && this.outputConnection.isConnected()) {
-      throw 'Still connected to parent block.';
+      throw Error('Still connected to parent block.');
     }
     this.parentBlock_ = null;
     // This block hasn't actually moved on-screen, so there's no need to update
@@ -841,8 +842,8 @@ Blockly.Block.prototype.setColour = function(colour, colourSecondary, colourTert
  * @throws {Error} if onchangeFn is not falsey or a function.
  */
 Blockly.Block.prototype.setOnChange = function(onchangeFn) {
-  if (onchangeFn && !goog.isFunction(onchangeFn)) {
-    throw new Error("onchange must be a function.");
+  if (onchangeFn && typeof onchangeFn != 'function') {
+    throw new Error('onchange must be a function.');
   }
   if (this.onchangeWrapper_) {
     this.workspace.removeChangeListener(this.onchangeWrapper_);
@@ -964,7 +965,9 @@ Blockly.Block.prototype.getFieldValue = function(name) {
  */
 Blockly.Block.prototype.setFieldValue = function(newValue, name) {
   var field = this.getField(name);
-  goog.asserts.assertObject(field, 'Field "%s" not found.', name);
+  if (!field) {
+    throw Error('Field "' + name + '" not found.');
+  }
   field.setValue(newValue);
 };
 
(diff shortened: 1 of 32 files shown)


core/block.js
```diff
@@ -964,7 +965,9 @@ Blockly.Block.prototype.getFieldValue = function(name) {
  */
 Blockly.Block.prototype.setFieldValue = function(newValue, name) {
   var field = this.getField(name);
-  goog.asserts.assertObject(field, 'Field "%s" not found.', name);
+  if (!field) {
```
kchadha, line 968: Should this be -- `if (!field || typeof field != 'object')`?

core/block.js
```diff
@@ -980,16 +983,20 @@ Blockly.Block.prototype.setPreviousStatement = function(newBoolean, opt_check) {
       opt_check = null;
     }
     if (!this.previousConnection) {
-      goog.asserts.assert(!this.outputConnection,
-          'Remove output connection prior to adding previous connection.');
+      if (this.outputConnection) {
```
kchadha, line 986: Should this be `if (!!this.outputConnection)`? I'm not sure if there would be a case that meets one but not the other...

kchadha: (same question in the cases below as well)

core/block.js
```diff
@@ -1300,8 +1317,8 @@ Blockly.Block.prototype.jsonInit = function(json) {
  * @param {boolean=} opt_disableCheck Option flag to disable overwrite checks.
  */
 Blockly.Block.prototype.mixin = function(mixinObj, opt_disableCheck) {
-  if (goog.isDef(opt_disableCheck) && !goog.isBoolean(opt_disableCheck)) {
-    throw new Error("opt_disableCheck must be a boolean if provided");
+  if (opt_disableCheck !== undefined && typeof opt_disableCheck != 'boolean') {
```
kchadha, line 1320: This expression uses both `!==` and `!=`, is that intentional? Also, why not also use typeof when checking whether something is (or is not) undefined?

core/block_drag_surface.js
```diff
@@ -179,8 +178,9 @@ Blockly.BlockDragSurfaceSvg.prototype.createDropShadowDom_ = function(defs) {
  * surface.
  */
 Blockly.BlockDragSurfaceSvg.prototype.setBlocksAndShow = function(blocks) {
-  goog.asserts.assert(
-      this.dragGroup_.childNodes.length == 0, 'Already dragging a block.');
+  if (this.dragGroup_.childNodes.length) {
```
kchadha, line 181: This would be clearer to read/understand if it were: `if (this.dragGroup_.childNodes.length > 0)`

core/block_drag_surface.js
```diff
@@ -286,8 +286,9 @@ Blockly.BlockDragSurfaceSvg.prototype.clearAndHide = function(opt_newSurface) {
     this.dragGroup_.removeChild(this.getCurrentBlock());
   }
   this.SVG_.style.display = 'none';
-  goog.asserts.assert(
-      this.dragGroup_.childNodes.length == 0, 'Drag group was not cleared.');
+  if (this.dragGroup_.childNodes.length) {
```
kchadha, line 289: Same as above.

core/extensions.js
```diff
@@ -76,7 +73,7 @@ Blockly.Extensions.register = function(name, initFn) {
  *     registered.
  */
 Blockly.Extensions.registerMixin = function(name, mixinObj) {
-  if (!goog.isObject(mixinObj)){
+  if (!mixinObj || typeof mixinObj != 'object'){
```
kchadha, line 76: Can't this be replaced with just the RHS?

core/field.js
```diff
@@ -81,10 +80,10 @@ Blockly.Field.TYPE_MAP_ = {};
  *     object containing a fromJson function.
  */
 Blockly.Field.register = function(type, fieldClass) {
-  if (!goog.isString(type) || goog.string.isEmptyOrWhitespace(type)) {
+  if ((typeof type != 'string') || (type.trim() == '')) {
     throw new Error('Invalid field type "' + type + '"');
   }
-  if (!goog.isObject(fieldClass) || !goog.isFunction(fieldClass.fromJson)) {
+  if (!fieldClass || (typeof fieldClass.fromJson != 'function')) {
```
kchadha, line 86: Should the LHS be `typeof fieldClass != 'object`

core/utils.js
```diff
@@ -935,3 +935,50 @@ Blockly.utils.getViewportBBox = function() {
...
+/**
+ * Fast prefix-checker.
+ * Copied from Closure's goog.string.startsWith.
+ * @param {string} str The string to check.
+ * @param {string} prefix A string to look for at the start of `str`.
+ * @return {boolean} True if `str` begins with `prefix`.
+ */
+Blockly.utils.startsWith = function(str, prefix) {
+  return str.lastIndexOf(prefix, 0) == 0;
+};
```
kchadha, line 948: There should probably be unit tests for these newly added functions.

core/workspace_drag_surface_svg.js
```diff
@@ -160,8 +161,9 @@ Blockly.WorkspaceDragSurfaceSvg.prototype.clearAndHide = function(newSurface) {
   Blockly.utils.insertAfter_(bubbleCanvas, blockCanvas);
   // Hide the drag surface.
   this.SVG_.style.display = 'none';
-  goog.asserts.assert(
-      this.SVG_.childNodes.length == 0, 'Drag surface was not cleared.');
+  if (this.SVG_.childNodes.length) {
```
kchadha, line 164: Same comment as above about preferring: `this.SVG_.childNodes.length > 0`...

core/workspace_drag_surface_svg.js
```diff
@@ -180,8 +182,9 @@ Blockly.WorkspaceDragSurfaceSvg.prototype.clearAndHide = function(newSurface) {
  */
 Blockly.WorkspaceDragSurfaceSvg.prototype.setContentsAndShow = function(
     blockCanvas, bubbleCanvas, previousSibling, width, height, scale) {
-  goog.asserts.assert(
-      this.SVG_.childNodes.length == 0, 'Already dragging a block.');
+  if (this.SVG_.childNodes.length) {
```
kchadha, line 185: Same as above. It's okay if we don't end up changing these, but I want to track them all together via comments in case they do end up being changed.

rachel-fenichel:
@NeilFraser can you comment on some of the RHS/LHS questions?  This PR is all code that you changed in Blockly.

core/block.js
NeilFraser: this.getField returns a Field object, or null.  No other return values are possible.  There's no need to check for anything else.

core/block.js
NeilFraser: If statements have an implicit Boolean() in the argument.

core/block.js
NeilFraser: != isn't useful for checking against undefined, since it will also catch null, '', and other falsy values.  Whereas comparing against the string 'boolean' does not require !==.

`typeof opt_disableCheck != 'undefined'` is more verbose and less clear and slower than just `opt_disableCheck !== undefined`

core/block_drag_surface.js
NeilFraser: Just following the Google JS Style Guide which refers to this as "shorter and nicer" (under "Tricks and Tips).  It is the same way that the Closure Library is written:
https://github.com/google/closure-library/blob/master/closure/goog/color/color.js#L116

core/extensions.js
NeilFraser: Nope:  typeof null == 'object'

core/field.js
NeilFraser: Nope: typeof null == 'object'

kchadha:
@NeilFraser Thanks for answering all the questions! 

@rachel-fenichel This change looks good to go, but I think we are holding off on making big changes while the new 3.0 banner is up on the scratch website.

cc/ @thisandagain

rachel-fenichel:
Friendly ping--can we merge this yet?

kchadha:
@rachel-fenichel, I just talked to Ray, and he said it should be okay to merge.

kchadha (approved):
LG

PR #1668, opened 2018-08-02
RTL positioning for category labels and buttons in the flyout.

### LINKED ISSUE #1584: Category labels and buttons should right align if the layout is RTL
### Expected Behavior

If the layout is RTL, category labels and flyout buttons should be right aligned.

### Actual Behavior

Category labels and buttons are left aligned - for example the Data category in the playground:
screen shot 2018-06-12 at 2 28 54 pm

### Steps to Reproduce

In the vertical playground, switch the direction to 'RTL', scroll to the Data category.

kchadha:
### Resolves

Resolves #1584

### Proposed Changes

Re-position category labels and buttons (including extension status button) in the toolbox when in RTL mode.

### Reason for Changes

Localization.

### Test Coverage

Manual testing in the vertical_playground as well as scratch-gui.

/cc @rachel-fenichel @AnmAtAnm

### DIFF
--- a/core/flyout_extension_category_header.js
+++ b/core/flyout_extension_category_header.js
@@ -81,7 +81,8 @@ Blockly.FlyoutExtensionCategoryHeader.prototype.createDom = function() {
   var marginX = 15;
   var marginY = 10;
 
-  var statusButtonX = (this.flyoutWidth_ - statusButtonWidth - marginX) / this.workspace_.scale;
+  var statusButtonX = this.workspace_.RTL ? (marginX - this.flyoutWidth_ + statusButtonWidth) :
+      (this.flyoutWidth_ - statusButtonWidth - marginX) / this.workspace_.scale;
 
   if (this.imageSrc_) {
     /** @type {SVGElement} */
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -490,7 +490,11 @@ Blockly.VerticalFlyout.prototype.layout_ = function(contents, gaps) {
     } else if (item.type == 'button') {
       var button = item.button;
       var buttonSvg = button.createDom();
-      button.moveTo(cursorX, cursorY);
+      if (this.RTL) {
+        button.moveTo(flyoutWidth - this.MARGIN - button.width, cursorY);
+      } else {
+        button.moveTo(cursorX, cursorY);
+      }
       button.show();
       // Clicking on a flyout button or label is a lot like clicking on the
       // flyout background.


core/flyout_extension_category_header.js
```diff
@@ -81,7 +81,8 @@ Blockly.FlyoutExtensionCategoryHeader.prototype.createDom = function() {
   var marginX = 15;
   var marginY = 10;
 
-  var statusButtonX = (this.flyoutWidth_ - statusButtonWidth - marginX) / this.workspace_.scale;
+  var statusButtonX = this.workspace_.RTL ? (marginX - this.flyoutWidth_ + statusButtonWidth) :
+      (this.flyoutWidth_ - statusButtonWidth - marginX) / this.workspace_.scale;
```
rachel-fenichel, line 85: Why are you dividing by scale in the LTR case but not the RTL case?

kchadha: Yeah, that's what @chrisgarrity and I were trying to figure out too. For some reason, not dividing by the scale in the LTR case displays the extension status button part of the way through the width of the flyout instead of at the right edge of the flyout. 

In the RTL case, dividing by the workspace scale does the analogous thing.

Our guess was that somewhere before this point, the flyout width is already being divided by the workspace scale but only in the RTL case... but we couldn't find it.

chrisgarrity: I think it's related to how we calculate the moveTo x in flyout_vertical below. It's using flyoutWidth. That flyoutWidth has already had the scale factor applied: https://github.com/LLK/scratch-blocks/blob/c21757220dbda20098fd7ee25a3232671bcb316d/core/flyout_vertical.js#L448

chrisgarrity (approved):
This is working as is, and all the other variations didn't get both the labels and the status into the correct positions.

core/flyout_extension_category_header.js
rachel-fenichel: Ah, and that value of flyoutWidth is only being used for RTL.

But then it looks like most (all?) parts of statusButtonX are being negated in that moveTo below, so I'm still very confused about what this code is doing.
Is this value of statusButtonX the

kchadha: Basically, I think the statusButtonX calculation is just placing the status button to the left of the `Extensions` label (instead of to the right, like it is in LTR), and then the moveTo is moving the *entire* label (actual status button + the word `Extensions`) into the right place on the flyout

PR #1670, opened 2018-08-03
Make holding shift to scroll work on Firefox

### LINKED ISSUE #1662: Holding shift while scrolling doesn't scroll horizontally
### Expected Behavior

Holding shift while using the scroll wheel should scroll left or right based on deltaY.

### Actual Behavior

Holding shift while scrolling doesn't change anything.

(Other browsers set deltaX instead of deltaY while shift is held, but Firefox doesn't.)

### Steps to Reproduce

Hold shift and use the scroll wheel while running Firefox. Notice that the block workspace scrolls up/down instead of left/right.

Firefox 62.0b10, Debian (testing).

---

Sort of branching off of #1427 here. This is exactly the same issue as LLK/scratch-paint#156, which I fixed with in LLK/scratch-paint#263. Shouldn't be too big of a deal to port the same code to scratch-blocks.

towerofnix:
### Resolves

Fixes #1662.

### Proposed Changes

Copied from LLK/scratch-paint#263:

>This is a "workaround" for browsers which do not automatically set deltaX when the shift key is held (seen on Firefox-Linux and Chrome-Windows, but not Chrome-Mac). This PR makes it so that when the shift key is held while scrolling, deltaX is ignored, and the distance scrolled horizontally is set to deltaY.

### Reason for Changes

Support for horizontal scrolling with the scroll wheel by holding shift on Firefox.

### Test Coverage

Tested locally on Firefox/Linux. Functionally-identical code reviewed and tweaked in discussion with @fsih in the above scratch-paint PR.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1320,6 +1320,14 @@ Blockly.WorkspaceSvg.prototype.onMouseWheel_ = function(e) {
     var x = this.scrollX - e.deltaX * multiplier;
     var y = this.scrollY - e.deltaY * multiplier;
 
+    if (e.shiftKey && e.deltaX === 0) {
+      // Scroll horizontally (based on vertical scroll delta)
+      // This is needed as for some browser/system combinations which do not
+      // set deltaX. See #1662.
+      x = this.scrollX - e.deltaY * multiplier;
+      y = this.scrollY; // Don't scroll vertically
+    }
+
     this.startDragMetrics = this.getMetrics();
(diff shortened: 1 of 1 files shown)


kchadha (commented):
These changes look good, @fsih, do you mind testing the before/after of this on Windows firefox?

fsih:
Tested Chrome, Firefox, and Edge on Windows; none were working before, but all are horizontal scrolling with shift key pressed now!

PR #1677, opened 2018-08-10
Place newly-created custom blocks in the top left of the visible workspace

### LINKED ISSUE #1334: New define block should be added to upper left of view, not absolute workspace
If you are scrolled far over in a large project, the new define block gets placed in the absolute top left

towerofnix:
### Resolves

Resolves #1334.

### Proposed Changes

Places newly-created custom blocks in the top left of the visible workspace.

AFAIK a good follow-up (which I really wouldn't know where to begin with testing) is to make this go to the top-*right* corner in RTL mode, if that's something that is wanted.

### Reason for Changes

Convenience: without this PR, new custom blocks are placed somewhere else in the workspace, which is both somewhat unpredictable (it's presumably the top left of the "absolute workspace"), and annoying, because you have to scroll all the way there to grab it and pull it where you want it to go.

### Test Coverage

Manually tested at several different zoom levels and scroll positions.

### DIFF
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -417,7 +417,8 @@ Blockly.Procedures.createProcedureCallbackFactory_ = function(workspace) {
       var blockDom = Blockly.Xml.textToDom(blockText).firstChild;
       Blockly.Events.setGroup(true);
       var block = Blockly.Xml.domToBlock(blockDom, workspace);
-      block.moveBy(30, 30);
+      var scale = workspace.scale; // To convert from pixel units to workspace units
+      block.moveBy((-workspace.scrollX + 30) / scale, (-workspace.scrollY + 30) / scale);
       block.scheduleSnapAndBump();
       Blockly.Events.setGroup(false);
     }
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -418,8 +418,15 @@ Blockly.Procedures.createProcedureCallbackFactory_ = function(workspace) {
       Blockly.Events.setGroup(true);
       var block = Blockly.Xml.domToBlock(blockDom, workspace);
       var scale = workspace.scale; // To convert from pixel units to workspace units
-      // Position the block so that it is at the top left of the visible workspace.
-      block.moveBy((-workspace.scrollX + 30) / scale, (-workspace.scrollY + 30) / scale);
+      // Position the block so that it is at the top left of the visible workspace,
+      // padded from the edge by 30 units. Position in the top right if RTL.
+      var posX = -workspace.scrollX;
+      if (workspace.RTL) {
+        posX += workspace.getMetrics().contentWidth - 30;
+      } else {
+        posX += 30;
+      }
+      block.moveBy(posX / scale, (-workspace.scrollY + 30) / scale);
       block.scheduleSnapAndBump();
       Blockly.Events.setGroup(false);
     }
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -418,6 +418,7 @@ Blockly.Procedures.createProcedureCallbackFactory_ = function(workspace) {
       Blockly.Events.setGroup(true);
       var block = Blockly.Xml.domToBlock(blockDom, workspace);
       var scale = workspace.scale; // To convert from pixel units to workspace units
+      // Position the block so that it is at the top left of the visible workspace.
       block.moveBy((-workspace.scrollX + 30) / scale, (-workspace.scrollY + 30) / scale);
       block.scheduleSnapAndBump();
       Blockly.Events.setGroup(false);


core/procedures.js
```diff
@@ -417,7 +417,8 @@ Blockly.Procedures.createProcedureCallbackFactory_ = function(workspace) {
       var blockDom = Blockly.Xml.textToDom(blockText).firstChild;
       Blockly.Events.setGroup(true);
       var block = Blockly.Xml.domToBlock(blockDom, workspace);
-      block.moveBy(30, 30);
+      var scale = workspace.scale; // To convert from pixel units to workspace units
```
rachel-fenichel, line 420: Add a comment saying that this is putting it at the top left of the visible workspace.

Also, I agree that the right behaviour for RTL is to put it at the top right.  You can test that in the playgrounds by switching to RTL mode and manually calling a function like this to create your block.

If you don't want to do that right now, create an issue describing it and add a comment here that says "TODO(#issue number): RTL)

towerofnix: Got it, will add that comment. I'll have a go but might make an issue for the RTL part.

towerofnix:
@rachel-fenichel Okay, I made the block get placed at the top right in RTL mode. Here are some screenshots of where the block is positioned in LTR and RTL:

LTR: block in top left corner
RTL: block in top right corner

There is a bit less of a margin in RTL mode, for some reason. Should the margin there be `- 60` instead of `- 30`, or something similar, to account for that?

rachel-fenichel:
Leave it at 30--the margin problem is likely an issue with the measurement of the block itself, not with your math, and it's easier to undo one bit of wonkiness than two.  Thanks.

towerofnix:
@rachel-fenichel Got it.

kchadha:
I tried testing these changes in the custom procedure playground (`test/custom_procedure_playground.html`) and it doesn't seem to work quite right.

In LTR, the first new block gets placed to the top left of the block that is already at the top left of the current view. When adding more custom procedure blocks, it seems to move the definition block it has already placed on top of the pre-existing ones. See images below:

**Workspace when the custom procedure playground is first loaded:**

image

**After adding a new block:**
image

**After adding a new block a second time "(notice that `New block#1` got moved...):**
image


I also tested this in RTL by manually modifiying the workspace options in inject ([here](https://github.com/LLK/scratch-blocks/blob/develop/tests/custom_procedure_playground.html#L154)) to add `rtl: true` in the definition workspace and added a custom procedure using the declaration workspace tool. 

When RTL is true, the block is definition hat is placed outside of the current view, and every subsequent time I try to add a new custom procedure, the definition hat gets placed further and further away out of view.

towerofnix:
@kchadha You're correct that there are issues with blocks being spawned on top of each other, but that's not an issue that's addressed in this PR nor is it one caused by this.

For comparison, here are screenshots of the same actions without my PR (i.e. under the develop branch)..

**Workspace when the custom procedure playground is first loaded:**

Identical to respective screenshot in kchadha's post

**After adding a new block:**

Again, identical to respective screenshot in kchadha's post (but the custom block is called "New Block #1 (in develop branch)"

**After adding a new block a second time:**

Result of having created a second new block; this one differs from the respective screenshot according to the description below

This third image is the only case where there is any difference under the develop branch ([compare with your screenshot in my PR](https://user-images.githubusercontent.com/1786240/44235937-53921f00-a179-11e8-8307-01ae23729cb4.png)). And.. yeah, the behavior under develop certainly makes more sense than under this PR. It's something to look into.

---

However, the focus of this PR is the behavior of where blocks are placed *after* you scroll the workspace. In the following screenshot, which is under the develop branch, I scrolled up and left as far as the workspace would let me, then created a new block (and then took the screenshot):

Result of having created a new block after scrolling as far up and left as possible; description follows

As can be seen, the new block is positioned immediately adjacent to the existing custom block. That is, it is positioned at the center/origin of the workspace, *not* the top left of where I had scrolled to.

This PR changes that behavior to create the custom block at the top left of the visible workspace; that is, the top left of where you've scrolled to. Here is a screenshot where I took the same steps under the branch of this PR:

Result of the same steps taken under this PR; description follows

The new custom block is in the top left of the visible workspace.

This is useful for when the user has scrolled significantly far away from the center (or rather, origin) of the workspace. Under develop, if the user creates a custom block, then they won't know where the custom block appeared; they'd see it in the block palette ("my new block") but would have to search the workspace for the "define my new block" hat block. This PR fixes that by predictably creating the new custom block in the top left of the visible workspace, meaning the "define my new block" will always be visible.

---

It's worth discussing why the behavior of the third screenshots in this post and yours are different, but I don't see why that would block this PR, seeing as A) it doesn't focus on that part of the code in the first place and B) it's a very rare case where the user will create two custom blocks one after the other without scrolling the workspace between. In that case, I think semi-undefined/unpredictable behavior is somewhat acceptable (and of course, it's not like the "define" block isn't created at all; it's just positioned on top of an existing one... which is not altogether surprising since you created two new blocks while scrolled to the same position).

(PS, thanks for mentioning `custom_procedure_playground.html` to me - while developing, I was rebuilding and testing in scratch-gui each time!)

kchadha:
@towerofnix Thank you for investigating the issue some more. Unfortunately, even though this PR is intended to only fix an issue when the user has scrolled the workspace, the changes here still affect the issue where the user does not scroll because moving the newly placed block actually does scroll the workspace a bit. It looks like the changes in this PR are somehow affecting the logic of `block.scheduleSnapAndBump()` call, specifically the call to `block.bumpNeighbors_()`.

@rachel-fenichel I'm wondering if you have any idea of what might be going wrong here?

rachel-fenichel:
I think the bumping issues already existed.  If I go to the custom procedures playground and create a few blocks (a through d), here's what it looks like:
Create 'a':
image
Create 'b':
image
Create 'c':
image
Create 'd':
image

Note that starting from the third block created, the blocks overlap--as opposed to in the images above for this branch, when they do it on the second block.

I agree that there's an interaction with `bumpNeighbours_`, probably just based on the block being placed a few pixels differently in the two versions--the search for neighbours looks for connections within `Blockly.SNAP_RADIUS` pixels.  The pseudocode for that function looks something like this:

```
for each connection on this block and all of its descendants
  find all connections of the opposite type nearby
  for each of these connections
    if either connection is disconnected, bump the inferior block
```

A block that has been bumped does *not* bump other blocks that it lands near, which is why blocks can end up piled on top of each other at the end of a bump (blockly has this too).

One solution may be to add some randomness into the distance that the block is bumped.  That would mean changing `Blockly.RenderedConnection.prototype.bumpAwayFrom_`. 

My suggestion is to accept this change and treat the bumping problem as a separate one, and try adding some randomness (+/- 5 or 10 pixels in each direction) to see if that makes the problem less bad.  Bumping is trying to fix a perception problem, where a used could think that two blocks are connected even when they aren't, so any fix that tests well with users is probably fine.

kchadha:
@rachel-fenichel, thanks for investigating and writing up the issue more clearly. I think I'll file a separate issue for this and mark it needs-discussion.

@towerofnix, @rachel-fenichel, any idea about why the blocks are getting placed so far off screen and further and further away each time in RTL? If we want to resolve that in a separate issue that's fine with me, but it should either be fixed here or turned into an issue so we can keep track of it.

rachel-fenichel:
Issue I just wrote up for it in Blockly: https://github.com/google/blockly/issues/2011
Scratch-blocks should probably have a separate one for this, since you may make different decisions (especially about the custom procedures).

I'm not sure about the RTL problem.

kchadha:
Added an issue for the bump neighbors problem (#1694) and the RTL placement issue (#1695), so I think this PR is good to go!

Thanks @towerofnix!

PR #1682, opened 2018-08-16
Remove uses of goog.dom.removeNode from the codebase

rachel-fenichel:
### Resolves

Reducing use of the closure library in scratch-blocks and blockly, with an eventual goal of removing the library entirely.

### Proposed Changes

Replace calls to `goog.dom.removeNode(thing)` with `this.parentNode.removeChild(thing)`.


rachel-fenichel:
Actually not ready to merge or review.

PR #1687, opened 2018-08-16
Utility function cleanup for merges

rachel-fenichel:
Piecewise merges :/

### DIFF
--- a/core/comment.js
+++ b/core/comment.js
@@ -199,6 +199,8 @@ Blockly.Comment.prototype.setVisible = function(visible) {
         /** @type {!Blockly.WorkspaceSvg} */ (this.block_.workspace),
         this.createEditor_(), this.block_.svgPath_,
         this.iconXY_, this.width_, this.height_);
+    // Expose this comment's block's ID on its top-level SVG group.
+    this.bubble_.setSvgId(this.block_.id);
     this.bubble_.registerResizeEvent(this.resizeBubble_.bind(this));
     this.updateColour();
   } else {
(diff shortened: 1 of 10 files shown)


core/comment.js
```diff
@@ -199,6 +199,8 @@ Blockly.Comment.prototype.setVisible = function(visible) {
         /** @type {!Blockly.WorkspaceSvg} */ (this.block_.workspace),
         this.createEditor_(), this.block_.svgPath_,
         this.iconXY_, this.width_, this.height_);
+    // Expose this comment's block's ID on its top-level SVG group.
+    this.bubble_.setSvgId(this.block_.id);
```
kchadha, line 203: Should there be an analogous line in ScratchBlockComment which sets the ID to the comment's actual ID (instead of the block ID...)?

kchadha (commented):
Minor question/comment in the code comments.

core/comment.js
rachel-fenichel: Up to you.  These changes come from https://github.com/google/blockly/pull/1700 and https://github.com/google/blockly/pull/1698.  

The change was to make it possible to get to the owner block from the bubble's DOM, for debugging.  Adding another attribute with the comment ID seems reasonable to me.

`setSvgId` has been in the scratch blocks codebase for a few months already

kchadha: Well I was thinking that setSvgId would take the comment ID as a parameter in the case of scratch comments. 

Regardless, do you think that change should be made in a separate PR? I think it makes sense for all the changes to go together here since this is the PR that's adding the IDs to scratch-blocks.

rachel-fenichel: I'd prefer to just merge this and handle that separately.  This change has plenty of other small stuff that I'd like to get in.  Alternatively, I can just remove this file from the PR and merge everything else.

kchadha: Filed an issue for it in #1696

rachel-fenichel: Okay.  Do you want me to remove this change from this PR?

PR #1693, opened 2018-08-17
Move code for pasing colour options into colours.js

rachel-fenichel:
### Resolves

None

### Proposed Changes

Moves a chunk of scratch-specific colour code out of options.js

### Reason for Changes

Separate scratch-specific code from general code, to improve merges.

### Additional Information
Scratch and Blockly do colours slightly differently (primary, secondary, tertiary vs. single colour).  We'd eventually like to move to a better way of doing colours in Blockly, but it's not high priority.  In the meantime, this is code that will continue to be different between the two repos, so I'm going to consider colours.js to be a file where we always use the scratch version during a merge, and try to move colour-related code there in both repositories.  (The file doesn't exist yet in Blockly, but I will make it as I move things there).

If we reach a point where our colour handling is essentially the same, I will be able to later handle merges between the two different versions of colours.js.  If we stay divergent, this will at least be cleaner.

### DIFF
--- a/core/colours.js
+++ b/core/colours.js
@@ -112,3 +112,36 @@ Blockly.Colours = {
   "valueReportBackground": "#FFFFFF",
   "valueReportBorder": "#AAAAAA"
 };
+
+/**
+ * Override the colours in Blockly.Colours with new values basded on the
+ * given dictionary.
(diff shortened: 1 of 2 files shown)


kchadha (approved):
LG

PR #1698, opened 2018-08-22
Revert "Throw Errors instead of using goog.asserts"

thisandagain:
Reverts LLK/scratch-blocks#1664
Resolves https://github.com/LLK/scratch-vm/issues/1496

Unfortunately, it looks like the changes introduced in #1664 caused some critical issues when combined with the VM / GUI in how exceptions are handled. We absolutely want to make this change and help with the removal of the Closure Library from Scratch Blocks, but we'll need to make sure we get the entire error handling flow sorted out across repos. 

As this is a critical bug, we are reverting for now and then will follow-up with a more permanent / long-term solution.

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -40,11 +40,11 @@ goog.require('Blockly.FieldLabelSerializable');
 goog.require('Blockly.FieldVariableGetter');
 goog.require('Blockly.Input');
 goog.require('Blockly.Mutator');
-goog.require('Blockly.utils');
 goog.require('Blockly.Warning');
 goog.require('Blockly.Workspace');
 goog.require('Blockly.Xml');
-
+goog.require('goog.array');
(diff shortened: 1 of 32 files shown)


rachel-fenichel:
Okay.

Unless something else changed recently, this means that the assert was previously being hit, and ignored.

kchadha (approved):
Looks good! I tested the changes in GUI and compared with the changes in the original PR #1664.

I [commented](https://github.com/LLK/scratch-vm/issues/1496#issuecomment-415185400) on the issue this is resolving noting that the VM is definitely doing something wrong with shadow blocks which is what causes scratch-blocks to throw an exception, but I agree that scratch-blocks errors should be handled by GUI/VM so that they don't cause user-facing crashes.

Issue #1683, opened 2018-08-16
Text with mixed english and Hebrew is strangely formatted in RTL.


chrisgarrity:
Text with placeholders is by definition a mixture of RTL and LTR characters as the placeholder must be '%1' etc.
The translation for glide is:
![screen shot 2018-08-16 at 5 36 44 pm](https://user-images.githubusercontent.com/399209/44236477-221a5300-a17b-11e8-90af-a4c292424de2.png)
`גלוש %1 שניות ל x: %2 y: %3`

But this appears as:
![screen shot 2018-08-16 at 5 37 45 pm](https://user-images.githubusercontent.com/399209/44236502-365e5000-a17b-11e8-8c99-efd4379aa7b4.png)

Clearly the 'x:' is in the wrong place.

A note from a translator suggests adding this style could help as it would override all the text.
```css
text.blocklyText { direction: rtl; }
```

guyhed:
As Chris wrote, the Hebrew version is not usable as it is now. Some blocks are unreadable, while others, like operators, work wrong. I give examples below.
It seems that the style command
`.blocklyText { direction: rtl; }`
does fix it all (at least, all the mistakes I have noticed).

For example, at present  "slide to" block looks like 
![screenshot from 2018-08-25 14-47-19](https://user-images.githubusercontent.com/1233716/44617979-30045e00-a876-11e8-9cd0-faf6765b88b6.png)
and after the implementation of the style above it looks as it should: 
![screenshot from 2018-08-25 14-46-30](https://user-images.githubusercontent.com/1233716/44617987-52967700-a876-11e8-93f5-2b4094a7f643.png)
The above style also fixes the order of the elements in the operator blocks. For example:
at present in the Hebrew version, due to inverse order of the fields, you get: 
![screenshot from 2018-08-25 14-53-05](https://user-images.githubusercontent.com/1233716/44618027-01d34e00-a877-11e8-8ab7-a15ec81af100.png)
after implementing the above style: 
![screenshot from 2018-08-25 14-53-59](https://user-images.githubusercontent.com/1233716/44618034-321aec80-a877-11e8-8b62-0dbd95b25e73.png)
(BTW, "true" and "false" should be translated, but I have noticed there is another thread regarding that.)

chrisgarrity:
Fixed in https://github.com/LLK/scratch-blocks/pull/1702

PR #1702, opened 2018-08-24
Handle RTL text

### LINKED ISSUE #1675: Text in block dropdown menus should be right-aligned in RTL
### Expected Behavior

Block dropdown menus in RTL should be a mirror image of how they look in LTR.

### Actual Behavior

LTR dropdown menus:

image

RTL dropdown menus:

image

screen shot 2018-08-24 at 12 14 18 pm

chrisgarrity:
### Resolves

resolves #1675 and #1683 (maybe also the rtl problems in edge, but I haven't checked)

Fixes issues related to dropdown menus (right align instead of left align), and bi-directional blocks (e.g., go to x: %1 y:%2, the ‘x: %1 y:%2’ is LTR, but should not be treated as LTR, when rendered in Hebrew it should be ‘%2 :y %1 :x’)

### Proposed Changes

Sets direction `rtl` for blocklyText when injection is RTL. And redefines the rtl dropdown menuitem styles to look better in RTL

### Reason for Changes

With these changes the dropdown menu and blocks text displays better:
screen shot 2018-08-24 at 4 59 57 pm
screen shot 2018-08-24 at 5 01 15 pm

**Without these changes:**
The text is left aligned even though the padding has been reversed:
screen shot 2018-08-24 at 5 00 37 pm

The way mixed latin and hebrew characters are rendered makes nonsense of the block (note the 'x:' at the start of the block):
screen shot 2018-08-24 at 5 00 57 pm

I did try `text-align: right` instead of setting direction for the dropdown, but it still does strange things with mixed bidirectional text:
screen shot 2018-08-24 at 5 50 40 pm

While the one with direction looks correct:
screen shot 2018-08-24 at 5 52 31 pm

/cc @rachel-fenichel while this makes scratch blocks look better, are there horrible side-effects in blockly I'm not aware of?

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -603,8 +603,12 @@ Blockly.Field.prototype.getDisplayText_ = function() {
   // Replace whitespace with non-breaking spaces so the text doesn't collapse.
   text = text.replace(/\s/g, Blockly.Field.NBSP);
   if (this.sourceBlock_.RTL) {
-    // The SVG is LTR, force text to be RTL.
-    text += '\u200F';
+    // The SVG is LTR, force text to be RTL unless a number.
+    if (this.sourceBlock_.editable_ && this.sourceBlock_.type === 'math_number') {
+      text += '\u200E';
+    } else {
+      text = '\u202B' + text + '\u202C';
+    }
   }
   return text;
 };
@@ -676,7 +680,11 @@ Blockly.Field.prototype.updateTextNode_ = function() {
   text = text.replace(/\s/g, Blockly.Field.NBSP);
   if (this.sourceBlock_.RTL && text) {
     // The SVG is LTR, force text to be RTL.
-    text += '\u200F';
+    if (this.sourceBlock_.editable_ && this.sourceBlock_.type === 'math_number') {
+      text += '\u200E';
+    } else {
+      text = '\u202B' + text + '\u202C';
+    }
   }
   if (!text) {
     // Prevent the field from disappearing if empty.
--- a/core/field.js
+++ b/core/field.js
@@ -605,7 +605,7 @@ Blockly.Field.prototype.getDisplayText_ = function() {
   if (this.sourceBlock_.RTL) {
     // The SVG is LTR, force text to be RTL unless a number.
     if (this.sourceBlock_.editable_ && this.sourceBlock_.type === 'math_number') {
-      text += '\u200E';
+      text = '\u202A' + text + '\u202C';
     } else {
       text = '\u202B' + text + '\u202C';
     }
@@ -681,7 +681,7 @@ Blockly.Field.prototype.updateTextNode_ = function() {
   if (this.sourceBlock_.RTL && text) {
     // The SVG is LTR, force text to be RTL.
     if (this.sourceBlock_.editable_ && this.sourceBlock_.type === 'math_number') {
-      text += '\u200E';
+      text = '\u202A' + text + '\u202C';
     } else {
       text = '\u202B' + text + '\u202C';
     }
--- a/core/css.js
+++ b/core/css.js
@@ -75,7 +75,7 @@ Blockly.Css.mediaPath_ = '';
  *     (providing CSS becomes the document's responsibility).
  * @param {string} pathToMedia Path from page to the Blockly media directory.
  */
-Blockly.Css.inject = function(hasCss, pathToMedia, isRtl) {
+Blockly.Css.inject = function(hasCss, pathToMedia) {
   // Only inject the CSS once.
   if (Blockly.Css.styleSheet_) {
     return;
--- a/core/css.js
+++ b/core/css.js
@@ -75,7 +75,7 @@ Blockly.Css.mediaPath_ = '';
  *     (providing CSS becomes the document's responsibility).
  * @param {string} pathToMedia Path from page to the Blockly media directory.
  */
-Blockly.Css.inject = function(hasCss, pathToMedia) {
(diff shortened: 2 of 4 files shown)


ericrosenbaum (approved):
LGTM!

chrisgarrity:
@AnmAtAnm @rachel-fenichel any comments on these changes for RTL blocklyText?

rachel-fenichel:
I'm still digging.  This note may be relevant: https://github.com/LLK/scratch-blocks/blob/develop/core/css.js#L1053

chrisgarrity:
I wondered about splitting out the blocklyDropDownDiv and just overriding the direction on that class, but wasn't sure if we were using the WidgetDiv somewhere that would matter. 

I was just looking at the source in the blockly playground, and somewhere the `blocklyWidgetDiv` got a `style="direction: rtl` added. That would make all the difference!
```
<div class="blocklyWidgetDiv" style="direction: rtl; display: block; height: 69px; top: 127px; left: 295.831px;">
    <div class="goog-menu goog-menu-vertical blocklyDropdownMenu" role="menu" aria-haspopup="true" tabindex="0" style="user-select: none;">
...
    </div>
</div>
```

Also, the text in the blocks has invisible right-to-left markers: `&rlm;` in front of the hebrew text. The question is how did they get there - automatically, or as part of the translation?

chrisgarrity:
Yup, widgetDiv has an rtl parameter to set the direction: 
https://github.com/LLK/scratch-blocks/blob/b537fc4b99b42fadf60fe7cf94f3156719c92bf6/core/widgetdiv.js#L110

but dropdownDiv doesn't:
https://github.com/LLK/scratch-blocks/blob/b537fc4b99b42fadf60fe7cf94f3156719c92bf6/core/dropdowndiv.js#L222

I'm not sure where the right-left-marker is coming from though.

rachel-fenichel:
Is this the right-to-left marker you're seeing? https://github.com/LLK/scratch-blocks/blob/develop/core/field.js#L607

chrisgarrity:
Yes, that's the marker, and it's definitely getting added to the scratch blocks, but it's not handling mixed hebrew with english. I looked through the blockly demo, and I was only able to find one place where English and Hebrew is mixed in the same blocklyText :
screen shot 2018-08-29 at 2 09 19 pm

Everywhere else that I saw they are in separate blocklyText items, for example:
screen shot 2018-08-29 at 2 13 08 pm

Here's what's in the json for 'go to x, y'
```
"MOTION_GOTOXY": "קפוץ אל x: %1 y: %2",
```
That ends up as two pieces of blockly text (and I can't even copy paste them as examples without things ending up out of order)

Even though a `&rlm` gets added to the string for the 'goto x:', the 'x:' ends up on the right of the hebrew (i.e. at the 'end' in LTR)

I did find a problem with just overriding the direction... negative number inputs appear as `10-`

chrisgarrity:
Did you know that when you enter a negative number in blockly in hebrew the minus shows up on the right:
screen shot 2018-08-29 at 6 53 46 pm

but then moves to the left when you stop editing.
screen shot 2018-08-29 at 6 54 00 pm

I do have a new version of scratch-blocks that uses the right-left-embed unicode character instead of the rlm, and everything looks better, except it has the same 'minus sign on the right when editing' as blockly:
screen shot 2018-08-29 at 7 11 35 pm

So it's getting closer!

rachel-fenichel:
Yes, and i think I filed a bug against blockly for that a while back, but I don't remember where the conversation went.

chrisgarrity:
Ah, the new code is actually forcing numbers to be LTR. I can make it put the negative on the right all the time if that's the desired outcome: https://github.com/google/blockly/issues/1802

core/field.js
```diff
@@ -603,8 +603,12 @@ Blockly.Field.prototype.getDisplayText_ = function() {
   // Replace whitespace with non-breaking spaces so the text doesn't collapse.
   text = text.replace(/\s/g, Blockly.Field.NBSP);
   if (this.sourceBlock_.RTL) {
-    // The SVG is LTR, force text to be RTL.
-    text += '\u200F';
+    // The SVG is LTR, force text to be RTL unless a number.
+    if (this.sourceBlock_.editable_ && this.sourceBlock_.type === 'math_number') {
+      text += '\u200E';
+    } else {
+      text = '\u202B' + text + '\u202C';
```
chrisgarrity: Can have negative numbers with the '-' on the right (`10-`) by just using line 610.

guyhed: We have a big problem with the blocks!
It seems that {direction:rtl} sould be use only on some blocks. On other blocks it causes things like:
capture

And it also causes stuff like the "10-" instead of "-10".

Note that all math notation and expression should be purely LTR. 
Maybe the solution is not to use RTL at all in the client, and revise the translation/json so that it will be rendered correctly is the client. 
Another option is to mark expression in the translation that should be rendered RTL (maybe by some prefix), and the rest will be LTR  by default.

chrisgarrity: @guyhed in this version most things are better - for example, < and > give the expected results. Subtraction and division can be handled in the translation by switching the order of the arguments in the translation knowing that they will be rendered RTL.

chrisgarrity:
@rachel-fenichel I think I'm going to land this PR as it is now. There's still the problem while editing numbers in RTL, but that's an existing issue in blockly that we can deal with separately. This fixes the bidi text and dropdown menus so it's worth getting in.

rachel-fenichel:
Sounds good.

ericrosenbaum (approved):
LG!

PR #1425, opened 2018-03-31
Add "item # of (thing) in (list)" block

towerofnix:
### Resolves

Towards LLK/scratch-gui#600. This should be merged alongside LLK/scratch-vm#1008, which contains the functionality of this block.

### Proposed Changes

Adds the "item # of (thing) in (list)" block, where "(thing)" is a text input and "(list)" is a list selection. This block is added to the palette alongside the other list blocks, and is placed below the "item (1) of (list)" block (per the spec).

### Reason for Changes

To implement the help-wanted feature.

### Test Coverage

Tested manually. Lint says this looks good!

PR #1714, opened 2018-09-12
Localize "item # of value in list" block

### LINKED ISSUE #1713: Localization for "item # of () in ()"
### Expected Behavior
All blocks localized

### Actual Behavior
item # of () in () block name is hard-coded.
(See blocks_vertical/data.js L394)

### Steps to Reproduce
See the code.

any

towerofnix:
### Resolves

Fixes #1713.

### Proposed Changes

Adds a translation/localization message for the "item # of () in ()" list block.

### Reason for Changes

Support for more translation.

I think my original PR for this block, #1425, must have been created before `Blockly.Msg` was used for blocks, but merged after - so this is a lone block left untranslated.

### Test Coverage

Tested manually - the block still shows up correctly in `tests/vertical_playground.html`.

### DIFF
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -391,7 +391,7 @@ Blockly.Blocks['data_itemnumoflist'] = {
    */
   init: function() {
     this.jsonInit({
-      "message0": "item # of %1 in %2",
+      "message0": Blockly.Msg.DATA_ITEMNUMOFLIST,
       "args0": [
         {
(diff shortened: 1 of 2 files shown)


chrisgarrity (approved):
Looks good, thanks for noticing the missing string.
The merge build may fail, as the new key should be in scratch_msgs, and it won't be there. I'm going to do a translations sync immediately after merging.

Issue #1223, opened 2017-11-10
List getter block does not have variable context menu


paulkaplan:
### Expected Behavior

_Please describe what should happen_

Right clicking a list getter block should give the same rename/delete options as with variables, and should populate a list of other lists to switch the block to. 

### Actual Behavior

_Describe what actually happens_

Because of https://github.com/LLK/scratch-blocks/issues/1222, the list getter block is actually using the wrong block, but once that is changed, there will be no right click menu. The current mixin for it is hard-coded to using the default variable type and assumes a field called `VARIABLE` for retrieving the variable name. https://github.com/LLK/scratch-blocks/blob/develop/blocks_vertical/data.js#L462-L500

PR #1715, opened 2018-09-13
scratch-gui #3007 - Added context menu to list variables

alexmotoc:
### Resolves

Resolves  https://github.com/LLK/scratch-gui/issues/3007 - New lists cannot be deleted or renamed by right clicking.

### Proposed Changes

Implemented context menu for lists similar to the one present for regular variables.

### Reason for Changes

This is an enhancement with the purpose of mirroring functionality that was present in the Scratch 2.0 editor.

### Test Coverage

No unit tests were implemented. A series of manual tests were executed after creating multiple variables & lists:

- Checked whether context menu was present for both.
- Checked whether helping text used the correct variable names.
- Checked whether renaming & deleting a variable or list produced the expected results.

### DIFF
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -166,7 +166,7 @@ Blockly.Blocks['data_listcontents'] = {
         }
       ],
       "category": Blockly.Categories.dataLists,
-      "extensions": ["colours_data_lists", "output_string"],
+      "extensions": ["contextMenu_getListBlock", "colours_data_lists", "output_string"],
       "checkboxInFlyout": true
     });
   }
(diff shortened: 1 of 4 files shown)


blocks_vertical/data.js
```diff
@@ -526,20 +526,68 @@ Blockly.Constants.Data.CUSTOM_CONTEXT_MENU_GET_VARIABLE_MIXIN = {
...
+        option.text = variablesList[i].name;
+
+        option.callback =
+            Blockly.Constants.Data.VARIABLE_OPTION_CALLBACK_FACTORY(this,
+                option.text, 'LIST');
+        options.push(option);
+      }
+    } else {
+      var renameOption = {
+        text: Blockly.Msg.RENAME_VARIABLE,
```
towerofnix: This should be changed to `Blockly.Msg.RENAME_LIST`, and the appropriate message should be added in `/msg/messages.js`. Right now it displays as "rename variable" instead of "rename list"!

blocks_vertical/data.js
```diff
@@ -526,20 +526,68 @@ Blockly.Constants.Data.CUSTOM_CONTEXT_MENU_GET_VARIABLE_MIXIN = {
...
+        enabled: true,
+        callback: Blockly.Constants.Data.RENAME_OPTION_CALLBACK_FACTORY(this, 'LIST')
+      };
+      var name = this.getField('LIST').text_;
+      var deleteOption = {
+        text: Blockly.Msg.DELETE_VARIABLE.replace('%1', name),
```
towerofnix: Same deal here - should be `DELETE_LIST`.

towerofnix (changes requested):
Right now the options show "variable" instead of "list" - I marked the spots to be changed to fix that in this review.

blocks_vertical/data.js
```diff
@@ -587,12 +639,13 @@ Blockly.Constants.Data.RENAME_OPTION_CALLBACK_FACTORY = function(block) {
...
+ * @param {string} type The type of the field (VARIABLE or LIST).
  * @return {!function()} A function that deletes the variable.
  */
-Blockly.Constants.Data.DELETE_OPTION_CALLBACK_FACTORY = function(block) {
+Blockly.Constants.Data.DELETE_OPTION_CALLBACK_FACTORY = function(block, type) {
   return function() {
     var workspace = block.workspace;
-    var variable = block.getField('VARIABLE').getVariable();
+    var variable = block.getField(type).getVariable();
     workspace.deleteVariableById(variable.getId());
```
towerofnix, line 649: It should be noted that this shows a confirmation bubble which displays "Delete (N) uses of this variable?". Changing that to show "list" would require changing `Blockly.VariableMap.prototype.deleteVariableById` (`core/variable_map.js`) - you'd need to check the type of the variable, then use the appropriate `Blockly.Msg` under `var confirmText`.

blocks_vertical/data.js
alexmotoc: See I wasn't sure about this one - that was my intuition at first too but checked the dropdown for the other list blocks and noticed they had the same naming convention so I kept it unchanged for consistency. What do you think?

screen shot 2018-09-13 at 17 46 28

towerofnix: Oh, interesting. I think that that's a bug and should be changed too. I don't think most users think of lists as another type of variable; that is, both variables and lists are things which store data (which is why they're in the Data category), but they're distinct.

towerofnix: Oh also, I'm 99.5% sure (although I can't check at the moment) that Scratch 2.0 has a "delete list" option - not "delete variable".

alexmotoc: You are correct. In Scratch 2.0 you can find "rename variable" and "delete variable" for variables whereas for lists the menu only contains "delete list". I can make the changes and also look into changing the text in the dropdown.

kchadha:
Hi @alexmotoc, thank you for submitting this pull request and I apologize for the delay in review.

As @towerofnix pointed out, it looks like the original bug this PR is addressing has been fixed elsewhere #1750.

It looks like there was some confusion because we had two different issues for the same bug filed in two different places.

I'm closing this PR as the changes have already been addressed.

alexmotoc:
@kchadha No worries, I saw that the fixed naming conventions have been addressed for the dropdown, however the initial purpose of this PR was to fix another issue - the lack of context menu for lists that is present for variables (right click to rename or delete) which PR #1750 is not adding. Could you double check please?

kchadha:
@alexmotoc, you're absolutely right! Sorry for the confusion! I'll reopen this PR for review. =]

rachel-fenichel:
We've got that now!  https://github.com/LLK/scratch-blocks/pull/1749 fixes #1223.

Cheers,
Rachel

kchadha:
Looks like the context menu addition was added in a different PR.

PR #1729, opened 2018-10-04
Add field sound dropdown

ericrosenbaum:
### Resolves

Progress toward https://github.com/LLK/scratch-gui/issues/1368

### Proposed Changes

Add a new dropdown field, specifically for the menu of sounds in the "play sound" and "start sound" blocks. The only change for now is that this field checks if a special item has been selected, "record...", which triggers a callback that can be overridden by the GUI.

### Reason for Changes

We'd like to make it more obvious that you can record your own sound, and make the process accessible directly from the play sound blocks themselves (as in Scratch 2.0). 

We imagine in the future that we might want to add additional options to this menu (such as "add sound..." to add a sound from the library). 

We also may want to add options like this to other blocks (such as options in the "switch costume" block for drawing a new costume and adding one from the library), which would mean making a similar subclass for the costume menu (or maybe abstract these into a more abstract menu-with-special-options class).

We also should consider using buttons inside the menus to better distinguish this type of option.

I encountered some questions- interested in feedback on these: 
- The special 'record...' option could be added by the field itself (I couldn't get that to work right because of the way the GUI updates dynamic menus). For now I am adding it from the GUI. 
- The value of the record option in the menu is currently a string constant, but it could instead be set to the callback itself (and then the field would check if the value is a function, and if so call it directly). This would prevent the problem that exists in this version, where if you name a sound "SOUND_RECORD_ID", trying to select it triggers the sound recorder.   

### Test Coverage

The vertical playground's "play sound" and "start sound" blocks have a "record..." option that should trigger the callback and show an alert.

### DIFF
--- a/blocks_vertical/sound.js
+++ b/blocks_vertical/sound.js
@@ -37,7 +37,7 @@ Blockly.Blocks['sound_sounds_menu'] = {
       "message0": "%1",
       "args0": [
         {
-          "type": "field_dropdown",
+          "type": "field_sounddropdown",
           "name": "SOUND_MENU",
           "options": [
             ['1', '0'],
--- a/blocks_vertical/sound.js
+++ b/blocks_vertical/sound.js
@@ -49,7 +49,8 @@ Blockly.Blocks['sound_sounds_menu'] = {
             ['7', '6'],
             ['8', '7'],
             ['9', '8'],
-            ['10', '9']
+            ['10', '9'],
+            ['record...', Blockly.SOUND_RECORD_ID]
           ]
         }
       ],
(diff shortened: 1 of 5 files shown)


picklesrus:
I'm curious to learn more about the problem with dynamic menus and the gui.  Adding this to the top level Blockly namespace doesn't feel quite right.  I was expecting this to do something similar to what the eyedropper button does in the color slider field, though now that I look at it,  I don't really understand how that interacts outside the blocks. Do you?

ericrosenbaum:
Closing this, replaced by https://github.com/LLK/scratch-blocks/pull/1731

PR #1731, opened 2018-10-09
Menu items can be functions, to support the 'record...' option

ericrosenbaum:
### Resolves

Progress toward https://github.com/LLK/scratch-gui/issues/1368

### Proposed Changes

Field dropdown, when you select something, checks if its value is a function, and if so, calls it (and does not select it).

Also, add a new message to localize the string 'record...'.

### Reason for Changes

We'd like to make it more obvious that you can record your own sound, and make the process accessible directly from the play sound blocks themselves (as in Scratch 2.0).

Generally, we want to add additional options, such as "add sound..." to add a sound from the library, and the corresponding options for costumes and backdrops like "paint costume..." and "add costume...".

### Test Coverage

The vertical playground's "play sound" and "start sound" blocks have a "record..." option that calls a function.

### DIFF
--- a/blocks_vertical/sound.js
+++ b/blocks_vertical/sound.js
@@ -49,7 +49,10 @@ Blockly.Blocks['sound_sounds_menu'] = {
             ['7', '6'],
             ['8', '7'],
             ['9', '8'],
-            ['10', '9']
+            ['10', '9'],
+            ['call a function', function() {
+              window.alert('function called!');}
+            ]
           ]
         }
       ],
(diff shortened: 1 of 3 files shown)


paulkaplan:
@ericrosenbaum it looks like this wont set the value of the block once the action is taken, which i would think we might want (i.e. set the value of the block to the recorded sound). Is this something we want to happen later?

paulkaplan (approved):
LGTM, ~although because this can only call functions, i'm not sure how useful this will be long term... see my other comment.~ Just reviewed w/ eric, scratch2 doesn't actually do this, so ¯\\_(ツ)_/¯

PR #1741, opened 2018-10-23
Fix number picker positioning above block

### LINKED ISSUE #1506: Numpad for number fields shows up in the wrong place
### Expected Behavior

Soft number pad (touch devices) should show up near the number field that is focused.

### Actual Behavior

Number pad shows up at the top of the workspace for a block close to the bottom of the page.
img_0005

### Steps to Reproduce

1. On a touch device, drag out a ( () + () ) block to the bottom of the workspace.
2. Focus on a field of the block to type in it.

iOS 11.2.2 Safari

rachel-fenichel:
### Resolves

#1506 

### Proposed Changes

Get rid of a few extra terms in y positioning math when the dropdown div should be above the block.

### Reason for Changes

Fix #1506 

### Test Coverage

Tested in the vertical playground with a "a / b" block.

### Additional Information
I also removed an extra 8 px of padding that was accidentally getting added to the field_number dropdown position when it's below the block, and removed that constant.

### DIFF
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -82,13 +82,6 @@ Blockly.FieldNumber.fromJson = function(options) {
 Blockly.FieldNumber.DROPDOWN_WIDTH = 168;
 
 /**
- * Extra padding to add between the block and the num-pad drop-down, in px.
- * @type {number}
- * @const
- */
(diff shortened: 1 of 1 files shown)


picklesrus:
LGTM

PR #1742, opened 2018-10-23
When duplicating on touch, don't start a drag with the new blocks

### LINKED ISSUE #1647: Duplicating a stack causes the blocks workspace to be unresponsive
### Expected Behavior

After duplicating a stack of blocks they should appear on the workspace and you should be able to continue to work.

### Actual Behavior

After long tapping and duplicating the blocks nothing happens on the stage.  The new blocks appear exactly above the old blocks and they can not be moved.  No other blocks can be moved in the workspace and no blocks can be dragged from the blocks pallet to the workspace.

If you select the stage and then go back to the sprite where the code was duplicated you can now edit blocks again.

### Steps to Reproduce

Build a stack of blocks
Press and hold on the stack of blocks
Select Duplicate from the menu that appears
Note: the menu closes and it might look like no new blocks appear or that they appear right on top of the old blocks.
Try to drag blocks on the stage and add new blocks to the workspace
Note: You can't

There was another issue with duplication on android that this may be related to: https://github.com/LLK/scratch-gui/issues/2376

Chrome on Android

rachel-fenichel:
### Resolves

Fixes #1647 

### Proposed Changes

Check whether the context menu on a block was opened from a mouse event or other event.

If it was a mouse event, keep the current behaviour of creating new blocks and immediately starting a drag.

If it was not a mouse event, place the new blocks slightly offset from the old blocks (down and to the right in LTR, down and to the left in RTL) and don't start a new drag.

### Reason for Changes

Starting a new drag on touchend with no mouse pointer is a bad experience--no pointer to track, and no way to tell that the current state is "dragging".

### Test Coverage

Tested in LTR and RTL in the playground.

### Additional Information
I'm just offsetting by 100, 100.  If you want other offsets, that's easy to change as well.

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -669,8 +669,10 @@ Blockly.BlockSvg.prototype.showContextMenu_ = function(e) {
   var block = this;
   var menuOptions = [];
 
+  var isMouseEvent = Blockly.Touch.getTouchIdentifierFromEvent(e) == 'mouse';
   if (this.isDeletable() && this.isMovable() && !block.isInFlyout) {
-    menuOptions.push(Blockly.ContextMenu.blockDuplicateOption(block));
+    menuOptions.push(
+        Blockly.ContextMenu.blockDuplicateOption(block, isMouseEvent));
     if (this.isEditable() && this.workspace.options.comments) {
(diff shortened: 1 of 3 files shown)


picklesrus:
LGTM

PR #1744, opened 2018-10-23
Use grandparent block colours for number picker dropdowns

rachel-fenichel:
### Resolves

No issue, just came up in passing

### Proposed Changes

Set dropdown colours based on grandparent block colours for the number picker dropdown, making it match the behaviour of other dropdowns.

### Reason for Changes

Looks nicer

### Test Coverage

Old:
image

New:
image


picklesrus:
LGTM

PR #1748, opened 2018-10-24
Remove unneeded logs

### LINKED ISSUE #1220: "tried to start the same gesture twice" and "Trying to cancel a gesture recursively."
I'm guessing these are logged because they're not supposed to happen? Here is a way to get these to happen.

Trying to cancel a gesture recursively: Drag a block from the workspace to the flyout to delete it.
tried to start the same gesture twice: Click on a block, and while holding down left click, also right click the same spot

rachel-fenichel:
### Resolves

Fixes #1220 
### Proposed Changes

Get rid of some warnings that are no longer useful

### Reason for Changes

See my comment on #1220 for full context.

### DIFF
--- a/core/gesture.js
+++ b/core/gesture.js
@@ -570,7 +570,6 @@ Blockly.Gesture.prototype.handleUp = function(e) {
   Blockly.longStop_();
 
   if (this.isEnding_) {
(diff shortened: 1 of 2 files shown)


picklesrus (commented):
LGTM

PR #1749, opened 2018-10-24
Add a context menu to the list getter block

### LINKED ISSUE #1223: List getter block does not have variable context menu

rachel-fenichel:
### Resolves

Fixes #1223
### Proposed Changes

Create a mixin to add context menu options to list getter blocks.  It's parallel to the one that we have for variable getter blocks.  

Modify the referenced callback to take a `fieldName` parameter, since the variable blocks have a field named `VARIABLE` and the list blocks have a field named `LIST`.

### Reason for Changes

Match Scratch 2.0 behaviour.

### Test Coverage

Tested by creating multiple variables and multiple lists in the vertical playground, then checking that I can right-click on them.  In the flyout, right-click shows rename and delete options.  These work for lists, and still work for variables.

In the main workspace right-click shows all of the names of other variables of the same type.  This works for lists, and still works for variables.

### Additional Information
This adds two new messages, and the vm probably needs to listen for events and do something in response.

### DIFF
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -166,7 +166,7 @@ Blockly.Blocks['data_listcontents'] = {
         }
       ],
       "category": Blockly.Categories.dataLists,
-      "extensions": ["colours_data_lists", "output_string"],
+      "extensions": ["contextMenu_getListBlock", "colours_data_lists", "output_string"],
       "checkboxInFlyout": true
     });
   }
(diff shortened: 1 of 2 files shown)


picklesrus (commented):
LGTM

PR #1750, opened 2018-10-24
Use different text for variable and list deletion/renaming

### LINKED ISSUE #1434: Rename and delete options in list dropdowns refer to variables
### Expected Behavior

Dropdown text in the list picker refers to a list, not a variable.

### Actual Behavior

image

rachel-fenichel:
### Resolves

#1434 
### Proposed Changes

Check the default type of the variable field, and use different rename and delete messages accordingly.
### Reason for Changes

Not mixing up variables and lists in text that we display to users.

### Test Coverage

Tested in the vertical playground by creating variables and lists, opening the dropdowns, and renaming and deleting them.  Rename and delete both work on both variables and lists, and the text looks correct for all four options.

### DIFF
--- a/core/field_variable.js
+++ b/core/field_variable.js
@@ -313,11 +313,20 @@ Blockly.FieldVariable.dropdownCreate = function() {
     options.unshift(
         [Blockly.Msg.NEW_BROADCAST_MESSAGE, Blockly.NEW_BROADCAST_MESSAGE_ID]);
   } else {
-    options.push([Blockly.Msg.RENAME_VARIABLE, Blockly.RENAME_VARIABLE_ID]);
-    if (Blockly.Msg.DELETE_VARIABLE) {
+    // Scalar variables and lists have the same backing action, but the option
+    // text is different.
+    if (this.defaultType_ == Blockly.LIST_VARIABLE_TYPE) {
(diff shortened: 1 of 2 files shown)


towerofnix:
~~Note the (needs review, September milestone!) PR #1715 also fixes this. Although this PR is probably cleaner code to the same effect (compare with [that PR](https://github.com/LLK/scratch-blocks/pull/1715/files#diff-fb72a00279b943f530ea656a48336806R316)). Nonetheless might be interesting to take a look at it, since it's got other somewhat related changes. (ping @alexmotoc)~~

Edit: Oh. Apparently you also made another PR which includes the rest of the equal behavior in #1715, so.. welp. :P

---

The problem is that LLK/scratch-gui#3007, effectively a *duplicate* of #1223, was made, and was marked as "help wanted", but never as a duplicate (nobody noticed). So @alexmotoc made a PR to fix that, #1715. It was marked to be reviewed in September, but nobody got around to it. Then later, you, @rachel-fenichel, decided to fix #1223. Now there was already Alex's PR, but you didn't know about it, because it was fixing the *other* issue! So it wasn't mentioned in #1223. So now we have @rachel-fenichel's PRs and @alexmotoc's PR, which make the same changes and fix the same issues.

picklesrus (commented):
LGTM

PR #1751, opened 2018-10-24
Don't allow illegal connections to procedure prototype blocks

### LINKED ISSUE #1534: Control blocks can be wrapped around procedure prototype blocks
### Expected Behavior

Procedure prototype blocks should be non-interactive.

### Actual Behavior

If I create a custom procedure, then drag a control block (C-shape) onto the workspace and near to the procedure prototype block inside of the procedure definition block, the grey area appears around the prototype block indicating that I can wrap the control block around it.  Releasing the block at this point causes the prototype block to be duplicated within the control block and draggable around the workspace.
A third, non-draggable copy of the prototype block is also added to the workspace.

### Steps to Reproduce

1. Create a custom procedure
2. Drag a C-shaped control block close to the prototype block within the procedure definition
screen shot 2018-05-16 at 1 32 50 pm

3. Release the control block when the highlight appears
screen shot 2018-05-16 at 1 32 54 pm

Mac OS 10.13.4 Chrome 66

rachel-fenichel:
### Resolves

Fixes #1534

### Proposed Changes

Add another reason that connections can be disallowed because of custom procedures: when the superior block is not a procedure definition block but the inferior block is a procedure prototype block, it's not allowed.

### Reason for Changes

See #1534.  Behaviour was amusing but bad.

### Test Coverage

Tested by adding a few blocks to the custom procedure playground and checking there.

### DIFF
--- a/core/connection.js
+++ b/core/connection.js
@@ -312,11 +312,15 @@ Blockly.Connection.prototype.canConnectWithReason_ = function(target) {
     return Blockly.Connection.REASON_CHECKS_FAILED;
   } else if (blockA.isShadow() && !blockB.isShadow()) {
     return Blockly.Connection.REASON_SHADOW_PARENT;
-  } else if (blockA.type == Blockly.PROCEDURES_DEFINITION_BLOCK_TYPE &&
-      blockB.type != 'procedures_prototype' &&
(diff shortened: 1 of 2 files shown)


picklesrus:
LGTM!

PR #1752, opened 2018-10-24
Automate translation update

### LINKED ISSUE #1635: Automate syncing with Transifex and add test
Add a task that works for all Scratch Team developers to ensure that everything gets sync'd when a new message is added to the messages file:
* Regenerate the en.json source
* Upload the source file `en/json` to Transifex
* Translation files are downloaded from Transifex (to ensure new key is included for all languages)
* Regenerate the full scratch_msgs file with the new key.

Finally:
* Add a test to ensure that the task to regenerate everything has been run if there's a new message.

### Background from Paul
I encountered an issue in this pr https://github.com/LLK/scratch-blocks/pull/1622 where I added a key to `msg/messages.js` but did not run `npm run translate`, which was able to merged to develop but then couldn't be merged into GUI because it caused the integration tests to fail. 

I'm gonna write out all the things that happened from a developer trying to add a new message:
- I created a PR that added a new msg/messages.js: https://github.com/LLK/scratch-blocks/pull/1622 
- I didn't read the note @thisandagain added to the top of that file about running `npm translate` after adding keys.
- There were no tests to fail on the CI that would have prevented merging without having run that script. 
- When I realized something was wrong, I tried to run `npm run translate`, and it modified two files (en.js/en.json) in a way that looked reasonable, but it seemed to blow up most of the content of scratch_messages, so @kchadha and I didn't know how to proceed.
- Only @chrisgarrity knew what to do to fix it! Thanks for the help fixing it (https://github.com/LLK/scratch-blocks/pull/1634)

chrisgarrity:
### Resolves
Resolves #1635 

### Proposed Changes
Adds scripts to sync the source and translations with transifex. They're separate so that downloading the translations can be done without updating the source **file.** 

Adds a test that checks that the English source JSON and the translations in scratch_msgs all have the same keys. 

New strings (blocks) should only be added by the Scratch Team. When adding a new string, `npm run translate:update` must be run before committing. Scratch Team members should have their own api token set in the environment variable TX_TOKEN, or use the transifex scratchbot token. If new strings are added without syncing with transifex `npm run test:messages` will fail.

Part two of this will be creating a process where the `sync_tx_translations.js` can run automatically on a schedule and either submit a PR (like greenkeeper), or just merge (like scratchr2_translations)

### Reason for Changes
Want to support automatic updates of translations

### Test Coverage

Manually tested:
- [ ] removed a key from en.json - failed correctly
- [ ] added a key to en.json - failed correctly
- [ ] renamed a key in en.json - failed correctly
- [ ] removed a key from a locale in scratch-msgs.js - failed correctly
- [ ] added a key to a locale in scratch-msgs.js - failed correctly
- [ ] renamed a key in a locale in scratch-msgs.js - failed correctly


i18n/sync_tx_translations.js
```diff
@@ -0,0 +1,157 @@
...
+file += '\n';
+file += `Blockly.ScratchMsgs.locales["en"] =\n`;
+file += JSON.stringify(en, null, 4);
+file += ';\n';
+
+if (!process.env.TX_TOKEN) {
+  assert.fail('ERROR: You must be a Scratch Team member and set your TX_TOKEN to sync with Transifex');
+};
+
+let locales = [
```
rschamp: Is there a way for this list to be share-able somehow? Can it come from `scratch-l10n` somehow?

i18n/sync_tx_src.js
```diff
@@ -0,0 +1,28 @@
...
+const RESOURCE = 'blocks';
+const TX = new transifex({
+  project_slug: PROJECT,
+  credential: 'api:' + process.env.TX_TOKEN
+});
+
+let en = fs.readFileSync(path.resolve(__dirname, '../msg/json/en.json'));
+en = JSON.parse(en);
+
+if (!process.env.TX_TOKEN) {
```
rschamp: This check could happen before we do anything, even importing stuff!

i18n/sync_tx_src.js
```diff
@@ -0,0 +1,28 @@
...
+  console.log('ERROR: You must be a Scratch Team member and set your TX_TOKEN to sync with Transifex');
```
rschamp: I think `.log` goes to `stdout`, but I think we should probably correctly write to `stderr` with `console.error` to ensure the messages are output correctly when we have this running in an automated system.
```suggestion
  console.error(new Error('You must be a Scratch Team member and set your TX_TOKEN to sync with Transifex'));
```

i18n/sync_tx_src.js
```diff
@@ -0,0 +1,28 @@
...
+  process.exit(1);
+};
+
+// update Transifex with English source
+TX.uploadSourceLanguageMethod(PROJECT, RESOURCE, {content: JSON.stringify(en)}, (err, status) => {
+  if (err) {
+    console.log('ERROR: ', err.message);
```
rschamp: Same here with `console.error`. We might get better debugging info if we did `console.error(err)`, which I think should print a stack trace as well (if they are passing back actual Error objects).
```suggestion
    console.error(err);
```

i18n/sync_tx_translations.js
```diff
@@ -0,0 +1,157 @@
...
+`;
+
+// add English strings
+file += '\n';
+file += `Blockly.ScratchMsgs.locales["en"] =\n`;
+file += JSON.stringify(en, null, 4);
+file += ';\n';
+
+if (!process.env.TX_TOKEN) {
+  assert.fail('ERROR: You must be a Scratch Team member and set your TX_TOKEN to sync with Transifex');
```
rschamp: If you're going to use `assert` here then maybe go all the way and don't check first (you wouldn't need the above `if`, but GitHub only lets you suggest single line change :)
```suggestion
assert.ok(process.env.TX_TOKEN, 'You must be a Scratch Team member and set your TX_TOKEN to sync with Transifex');
```

i18n/sync_tx_translations.js
```diff
@@ -0,0 +1,157 @@
...
+    validate(translation.translations, translation.locale);
+    file += '\n';
+    file += `Blockly.ScratchMsgs.locales["${translation.locale}"] =\n`;
+    file += JSON.stringify(translation.translations, null, 4);
+    file += ';\n';
+  });
+  // write combined file
+  fs.writeFileSync(`${PATH_OUTPUT}/scratch_msgs.js`, file);
+}).catch((err) => {
+  console.log('Someing went wrong:', err);
```
rschamp: Should use `console.error`, and I'd argue for just outputting the error itself rather than stringifying it so we get more context.

```suggestion
  console.error(err);
```

i18n/sync_tx_translations.js
```diff
@@ -0,0 +1,157 @@
+const fs = require('fs');
+const path = require('path');
+const assert = require('assert');
```
rschamp, line 25: Apparently this is deprecated in favor of `const assert = require('assert').strict`, which causes every assertion to use its `strict_` counterpart, so you can just use `assert.equal` rather than `assert.strictEqual`.

i18n/test_scratch_msgs.js
```diff
@@ -0,0 +1,77 @@
...
+// current English keys
+let en = fs.readFileSync(path.resolve(__dirname, '../msg/json/en.json'));
+en = JSON.parse(en);
+const enKeys = Object.keys(en);
+
+// File paths
+const PATH_INPUT = path.resolve(__dirname, '../msg/scratch_msgs.js');
+
+// Match function
+const match = function (str) {
```
rschamp, line 22: What is this doing? It could have a more descriptive name/comment.

i18n/test_scratch_msgs.js
```diff
@@ -0,0 +1,77 @@
...
+        if (!match(str)) return;
+        const result = extract(str);
+        if (!result) return;
+        if (typeof result === 'string') {
+          // locale changed, validate the current collection of keys
+          try {
+              validateKeys();
+          }
+          catch (err) {
+            console.log('Key validation FAILED:', err.message);
```
rschamp: Should use console.error, and we could include a stack trace for more information if we output the full error.
```suggestion
            console.error('Key validation FAILED: %O', err);
```

i18n/test_scratch_msgs.js
```diff
@@ -0,0 +1,77 @@
...
+          }
+          // change locale, and reset keys array
+          locale = result;
+          keys = [];
+        } else {
+          keys.push(result.key);
+        }
+    }))
+    .pipe(es.wait(function (err) {
+        if (err) throw new Error(err);
```
rschamp: I think we could explicitly output this and exit ourselves here?
```suggestion
        if (err) {
            console.error(err);
            process.exit(1);
        }
```

rschamp (changes requested):
Looks like tests are failing.

I've got some suggestions about error handling/assertions (as usual? 😛). But don't know enough about this system to know about what the script is actually trying to do — could you demo/walk me through soon?

i18n/sync_tx_src.js
```diff
@@ -0,0 +1,28 @@
+const fs = require('fs');
+const path = require('path');
+const transifex = require('transifex');
+
+// Globals
+const PROJECT = 'scratch-editor';
+const RESOURCE = 'blocks';
```
rschamp: Eventually it would be nice if this lived in scratch-l10n, so maybe for now we should generalize and pass these in as arguments, along with the path to the JSON file, and then call this with those arguments in package.json?

i18n/test_scratch_msgs.js
chrisgarrity: check last set of keys

i18n/sync_tx_translations.js
chrisgarrity: `const assert = require('assert').strict;` is not supported in node 8.x, it starts in node 9. Since we currently build with node 8 this will fail in travis. I'll add a tech debt issue that we should update when we're no longer supporting node 8.

i18n/sync_tx_src.js
```diff
@@ -0,0 +1,50 @@
...
+ * @fileoverview
+ * Script to upload a source en.json file to a particular transifex project resource.
+ * Expects that the project and resource have already been defined in Transifex, and that
+ * the person running the script has the the TX_TOKEN environment variable set to an api
+ * token that has developer access.
+ */
+
+// Fail immediately if the TX_TOKEN is not defined
+if (!process.env.TX_TOKEN) {
+  console.error(new Error('You must be a Scratch Team member and set your TX_TOKEN to sync with Transifex'));
```
rschamp: Hmmm... looking at this again, this is pretty unfriendly because it doesn't allow us to even get to the "Usage" instructions. Maybe we should incorporate this into that message. It's also not clear in this message how you should supply `TX_TOKEN` — if we include the TX_TOKEN check in the "Usage" check below, we should specify that `TX_TOKEN` needs to be set as an environment variable.

i18n/sync_tx_translations.js
```diff
@@ -0,0 +1,114 @@
...
+ * @fileoverview
+ * Script to pull translations for blocks from transifex and generate the scratch_msgs file.
+ * Expects that the project and resource have already been defined in Transifex, and that
+ * the person running the script has the the TX_TOKEN environment variable set to an api
+ * token that has developer access.
+ */
+
+// Fail immediately if the TX_TOKEN is not defined
+if (!process.env.TX_TOKEN) {
+  console.error(new Error('You must be a Scratch Team member and set your TX_TOKEN to sync with Transifex'));
```
rschamp: Same goes here about how to supply `TX_TOKEN`, and maybe we should explain this error via a "Usage" message as well?

rschamp (approved):
Looks good — I have one more suggestion about the `TX_TOKEN` messages that you can take or leave :)

PR #1754, opened 2018-10-25
Add some missing utility functions

rachel-fenichel:
### Resolves

#1687 references Blockly.utils.startsWith and Blockly.utils.toRadians, neither of which exist in scratch-blocks.

### Proposed Changes

Create the missing functions (should have happened as part of #1687)

### Reason for Changes

Resolve errors on every mouse move.

### Test Coverage

Confirmed that the errors are gone in the vertical playground.

### DIFF
--- a/core/utils.js
+++ b/core/utils.js
@@ -935,3 +935,26 @@ Blockly.utils.getViewportBBox = function() {
     left: scrollOffset.x
   };
 };
+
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
Thanks @rachel-fenichel ! LGTM

PR #1755, opened 2018-10-25
Fix procedure definition blocks with trailing text

### LINKED ISSUE #1658: Custom block deformation with Magyar language
Custom block definitions become deformed when used with the Magyar language. This deformation causes a second of text to appear below the first. This also causes the bottom right corner to no-longer be rounded. This only affects custom block definitions, not any other block type. (as I am presently aware)

Deformation illustrated below:
deformed blocks

I am not currently aware of any other languages in which this occurs, but I will continue to check.

rachel-fenichel:
### Resolves

Fixes #1658 
### Proposed Changes

Adds a bunch of special casing to the rendering code:
- If a dummy input follows a statement input on a procedure definition block, let them stay in the same row
- If a dummy input is in a row after a statement input, for reasons given above, don't add the input's field widths to the running tally.
- When rendering a define block, check if there's a following dummy input and render its fields inline, then adjust size accordingly.

### Reason for Changes

The old code broke the block into multiple lines.  This was surfaced in Magyar and Korean.

### Test Coverage

New:
image

Old:
image

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -797,10 +797,16 @@ Blockly.BlockSvg.prototype.renderCompute_ = function(iconWidth) {
     if (!input.isVisible()) {
       continue;
     }
+    var isSecondInputOnProcedure = this.type == 'procedures_definition' &&
+        lastType && lastType == Blockly.NEXT_STATEMENT;
     var row;
-    if (!lastType ||
-        lastType == Blockly.NEXT_STATEMENT ||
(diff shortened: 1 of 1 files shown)


rachel-fenichel:
Making this code work:

picklesrus (commented):
LGTM. Thanks for leaving good notes for whoever finds this next...!

ZenithRogue:
Sweet! Glad to see my bug fixed!

Issue #1338, opened 2018-01-10
End of dropdown is not visible


kyleplo:
### Expected Behavior

You should be able to see the whole dropdown menu

### Actual Behavior

Part of the menu is hidden on small screens
![screenshot](https://user-images.githubusercontent.com/31634240/34793856-c4ffb34c-f61a-11e7-8bd5-be219c4030df.png)

### Steps to Reproduce

Drag a block with a menu low on the screen and open the menu

### Operating System and Browser

Chrome OS 63

kerrtravers:
I can't seem to repro this using Chrome 63.0.3239.13 on macOS. Does it not let you scroll up to view that other characters above?

thisandagain:
@kyleplo I can't repro this either. What operating system are you on @kyleplo ?

kyleplo:
Windows 10 Desktop

picklesrus:
@rachel-fenichel this is still a problem on scratch-blocks.  I don't think scratch-blocks has https://github.com/google/blockly/pull/1483/files (at least there is no reference to one of the variables defined there). Is this the fix you had in mind?

rachel-fenichel:
Looks like there were a couple related PRs there.  Here's the root issue: https://github.com/google/blockly/issues/1454

PR #1756, opened 2018-10-26
Remove custom touch handling for dropdowns

rachel-fenichel:
### Resolves

None that I found
### Proposed Changes
Get rid of custom touch handling code for dropdown fields.

### Reason for Changes

Scratch-blocks version of [this change](https://github.com/google/blockly/pull/1904).

The touch handler was selecting on touch up, which meant that you couldn't scroll a long dropdown--releasing the scroll would select an item.

Closure implemented touch handling since this code was put in place.
### Test Coverage

Tested in chrome pretending to be mobile, in scratch-blocks playground and integrated with scratch-gui


rachel-fenichel:
#1338 may be related (later elements in the list are unreachable if scrolling selects an item).

picklesrus (approved):
LGTM

PR #1763, opened 2018-11-01
Musical note picker field

### LINKED ISSUE #459: Specialized Menu — Musical Note Picker
Breaking out from [Design - Specialized field editors - Issue #149](https://github.com/LLK/scratch-blocks/issues/149)

For example, this will be used in places like 'play note (   ) for (   ) beats'
screen shot 2016-06-29 at 9 24 55 am

ericrosenbaum:
### Resolves

resolves https://github.com/LLK/scratch-blocks/issues/459

### Proposed Changes

Adds a custom field for selecting a musical note on a piano:
screen shot 2018-11-01 at 3 28 32 pm

Features include:
- Click on piano keys or edit numeric input to update the selection
- Override-able function to play the note when clicked
- Octave buttons with animated transition up and down the piano
- Click and drag to play many notes (not currently working on touch)

### Reason for Changes

The note picker is an existing 2.0 feature, and this is an updated version. We'll use this for the music extension and probably other extensions (including the EV3, which has a "beep" block).


core/field_note.js
```diff
@@ -0,0 +1,852 @@
...
+ */
+Blockly.FieldNote.KEY_LABEL_PADDING = 8;
+
+/**
+ * An array of objects with data describing the keys on the piano.
+ * @type {Array.<{name: String, pitch: Number, isBlack: boolean}>}
+ * @const
+ */
+Blockly.FieldNote.KEY_INFO = [
+  {name: 'C', pitch: 0},
```
apple502j, line 285: We should translate them.

apple502j (changes requested):

core/field_note.js
Kenny2github: @apple502j do musical notes have translations in other languages? At least in Chinese, a C is a C.

Kenny2github:
IIRC Scratch 2.0 has three octaves to pick from, and more if you flip pages. One octave is a bit annoying (when music covers multiple octaves, that means you do a lot of flipping back and forth). Could there be at least 2 octaves per page?

core/field_note.js
ericrosenbaum: Thanks for your thoughts @apple502j and @Kenny2github - we quickly discussed it with the team here and decided to keep the note names untranslated.

ErikMejerHansen:
Will the new field_type be added to the ArgumentTypeMap in https://github.com/LLK/scratch-vm/blob/develop/src/engine/runtime.js  to allow it to be used from extensions?

EDIT: Sorry had not seen: https://github.com/LLK/scratch-vm/pull/1720

blocks_common/note.js
```diff
@@ -0,0 +1,57 @@
+/**
```
picklesrus, line 1: Just curious if this is intentionally in the common folder (i.e. you intend it to be used in the horizontal grammar).  There are 2 inputs on the vertical one it is used in... but maybe for horizontal you would just give it a default length to play?

blocks_common/note.js
```diff
@@ -0,0 +1,57 @@
...
+
+goog.require('Blockly.Blocks');
+
+goog.require('Blockly.Colours');
+
+goog.require('Blockly.constants');
+
+Blockly.Blocks['note'] = {
+  /**
+   * Block for matrix value.
```
picklesrus: Update comment

media/icons/arrow_button.svg
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="utf-8"?>
```
picklesrus: Did you run svgo (or some other optimizer on this?) Seems like there's a lot of extra stuff (spaces/newlines but also sketch info and comments) in there.

core/field_note.js
```diff
@@ -0,0 +1,852 @@
...
+  this.octaveDownMouseDownWrapper_ = null;
+
+  /**
+   * A wrapper for the mouse down event on the octave up button.
+   * @type {!Array.<!Array>}
+   * @private
+   */
+  this.octaveUpMouseDownWrapper_ = null;
+};
+goog.inherits(Blockly.FieldNote, Blockly.FieldTextInput);
```
picklesrus, line 165: I should probably know the answer to this... but why FieldTextInput instead of FieldNumber? Is it just because we don't want the num pad thingy to pop up on mobile?

core/field_note.js
```diff
@@ -0,0 +1,852 @@
...
+  Blockly.FieldNote.superClass_.constructor.call(
+      this, opt_value, opt_validator);
+  this.addArgType('note');
+
+  /**
+   * Width of the field. Computed when drawing it, and used for animation.
+   * @type {number}
+   * @private
+   */
+  this.fieldWidth_ = 0;
```
picklesrus: What do you think about calling this fieldEditorWidth_ instead? For a while I thought it paralleled  the size_ object that fields have.

core/field_note.js
```diff
@@ -0,0 +1,852 @@
...
+ * @const
+ */
+Blockly.FieldNote.ANIMATION_FRACTION = 0.2;
+
+/**
+ * Path to the arrow svg icon, used on the octave buttons.
+ * @type {string}
+ * @const
+ */
+Blockly.FieldNote.ARROW_SVG_PATH = 'icons/arrow_button.svg';
```
picklesrus, line 320: what does this end up being relative to? Should it have the pathToMedia in it?

core/field_note.js
```diff
@@ -0,0 +1,852 @@
...
+
+/**
+ * Clean up this FieldNote, as well as the inherited FieldTextInput.
+ * @return {!Function} Closure to call on destruction of the WidgetDiv.
+ * @private
+ */
+Blockly.FieldNote.prototype.dispose_ = function() {
+  var thisField = this;
+  return function() {
+    Blockly.FieldNote.superClass_.dispose_.call(thisField)();
```
picklesrus, line 348: should this dispose of things like keySvgs and the other svgElement members?

core/field_note.js
```diff
@@ -0,0 +1,852 @@
...
+};
+
+/**
+ * Handle the mouse down event on a piano key.
+ * @param {!Event} e Mouse down event.
+ * @private
+ */
+Blockly.FieldNote.prototype.onMouseDownOnKey_ = function(e) {
+  this.mouseIsDown_ = true;
+  this.mouseUpWrapper_ = Blockly.bindEvent_(document.body, 'mouseup', this, this.onMouseUp_);
```
picklesrus, line 644: do things work okay enough if you mouse down on a key, then drag outside the browser window, and then let the mouse up?

core/field_note.js
```diff
@@ -0,0 +1,852 @@
...
+
+/**
+ * Play a note, by calling the externally overriden play note function.
+ * @private
+ */
+Blockly.FieldNote.prototype.playNoteInternal_ = function() {
+  if (Blockly.FieldNote.playNote_) {
+    Blockly.FieldNote.playNote_(
+        this.getValue(),
+        this.sourceBlock_.parentBlock_.getCategory()
```
picklesrus, line 687: why does playNote need the category?

core/field_note.js
```diff
@@ -0,0 +1,852 @@
...
+ */
+Blockly.FieldNote.playNote_ = function(/* noteNum, id*/) {
+  return;
+};
+
+/**
+ * Handle the event when the octave down button is pressed.
+ * @private
+ */
+Blockly.FieldNote.prototype.onOctaveDown_ = function() {
```
picklesrus: No real objection to having onOctaveDown and onOctaveUp, but I think you could just call changeOctaveBy_(-1) and changeOctaveBy_(1) directly up where you setup the mouse handlers and remove these.

picklesrus (commented):
Just a few questions and mostly minor comments!

blocks_common/note.js
ericrosenbaum: Hm, good point! I didn't think about this, but I think it's okay to keep it there. I could imagine we might want to use it in horizontal in the way you described.

blocks_common/note.js
ericrosenbaum: 👍

media/icons/arrow_button.svg
ericrosenbaum: good catch- optimized it!

core/field_note.js
ericrosenbaum: I think that's exactly the reason (I was just imitating the approach in field_angle, which probably did this same thing to solve that problem).

core/field_note.js
ericrosenbaum: changed it (and the height variable as well)

core/field_note.js
ericrosenbaum: It uses `Blockly.mainWorkspace.options.pathToMedia` down below- I think it's okay as is

core/field_note.js
ericrosenbaum: Good catch

core/field_note.js
ericrosenbaum: Ah! good point. By doing that, I was able to get into a state where it played the notes as I moved the mouse over them, now without the mouse down. But any click fixes it, so I think it's okay.

core/field_note.js
ericrosenbaum: This is because the field will be used by multiple extensions (for now, music and EV3), so playNote needs to pass in a way to identify the extension.

core/field_note.js
ericrosenbaum: Good idea!

fsih:
If @picklesrus picked a PR to pick at, how many pickers would @pickesrus pick?

ericrosenbaum:
> IIRC Scratch 2.0 has three octaves to pick from, and more if you flip pages. One octave is a bit annoying (when music covers multiple octaves, that means you do a lot of flipping back and forth). Could there be at least 2 octaves per page?

@Kenny2github actually the Scratch 2.0 note picker has two octaves, and no buttons to change octaves. This new note picker only has one octave, because the keys need to be wider to be usable on touch devices. That's why we added the new octave switcher buttons.

core/field_note.js
```diff
@@ -361,6 +361,11 @@ Blockly.FieldNote.prototype.dispose_ = function() {
     if (thisField.octaveUpMouseDownWrapper_) {
       Blockly.unbindEvent_(thisField.octaveUpMouseDownWrapper_);
     }
+    this.pianoSVG_ = null;
+    this.keySVGs_ = null;
```
picklesrus: We tend to clear arrays by doing .length = 0. Do we need to do that here?
Differences I can see are that doing something like 
a1 = [1,2,3,4]
a2 = a1
a1 = null;

leaves a2 alone but  replacing a1=null with a1.length = 0 leaves a2 empty as well. Thoughts?

PR #1765, opened 2018-11-02
Trying to resolve #1764

### LINKED ISSUE #1764: field_matrix does not use colors of the block containing it
### Expected Behavior
field_matrix should follow the colors of the block containing it.

### Actual Behavior

It keeps its own greenish color:
image

I've tried to solve the issue in the PR that references this issue, but my solution feels like a hack. Because the `parentBlock` property isn't set on field_matix's `sourceBlock_` when the affected functions are initially called I've added guards to protect against 'undefined'. This solves the issue, but feels wrong. 

In contrast field_angle also references `sourceBlock_.parentBlock_` but does not need the guards against `parentBlock` being undefined. I'm guessing that is because it only references `sourceBlock_.parentBlock_` in `showEditor_` which is called at a later stage than `init` and `setValue` where field_matrix sets its colors. 

So perhaps the issue is that setParent should be called earlier? But that change seems like it would be above my paygrade :)

### Steps to Reproduce

Add a block containing a field_matrix to an extension and the the extensions colors. 
Something along the lines of: 
```
getInfo: function () {
        return {
            blocks: [
                {
                    arguments: {
                        MATRIX: {
                            defaultValue: '0101011111111110111000100', 
                            type: 'matrix'
                        }
                    },
                    blockType: 'command',
                    opcode: 'show',
                    text: 'Show [MATRIX]'
                }
            ],
            colour: '#3D7DD9',
            colourSecondary: '#0D4E8E',
            colourTertiary: '#033D6C',
            id: 'myMatixTest',
            name: 'Matrix test',
            targetTypes: ['hub']
        };
    }
```

Mac OS 10.13.6 (17G65) Chrome Version 70.0.3538.77 (Official Build) (64-bit)

ErikMejerHansen:
### Resolves

#1764 

### Proposed Changes

Makes field_matrix look to its parent block to get its colors

### Reason for Changes

See details in #1764 

### Test Coverage

No tests aded

### DIFF
--- a/core/field_matrix.js
+++ b/core/field_matrix.js
@@ -213,6 +213,10 @@ Blockly.FieldMatrix.prototype.init = function() {
 
   this.sourceBlock_.getSvgRoot().appendChild(this.fieldGroup_);
 
+  // Make sure that the output shape gets the correct color
+  this.sourceBlock_.setShadowColour(this.getColourTertiary_());
+
+
   var thumbX = Blockly.BlockSvg.DROPDOWN_ARROW_PADDING / 2;
   var thumbY = (this.size_.height - Blockly.FieldMatrix.THUMBNAIL_SIZE) / 2;
   var thumbnail = Blockly.utils.createSvgElement('g', {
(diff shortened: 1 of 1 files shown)


thisandagain:
@picklesrus Bump

PR #1769, opened 2018-11-05
Disable pointer events on field angle arrow to fix firefox dragging

paulkaplan:
### Resolves

Resolves an issue where it was hard to use the angle picker on firefox because the image inside the handle would start dragging instead, messing with the events. Demo:

drag-angle--broken

### Proposed Changes

Add pointer-events: none to the image using CSS to prevent starting a drag on the image itself. This does not impact grabbing the handle because those events are bound on the circle around the image. 

### Reason for Changes

Fix firefox angle picker

### Test Coverage

drag-angle--fixed

This can be tested in the vertical playground

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -794,6 +794,9 @@ Blockly.Css.CONTENT = [
     'cursor: pointer;',
   '}',
 
+  '.blocklyAngleDragArrow {',
+    'pointer-events: none',
+  '}',
 
   '.blocklyAngleMarks {',
     'stroke: #fff;',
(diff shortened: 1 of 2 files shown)


picklesrus (approved):
LGTM

rachel-fenichel:
This probably also needs to be pushed upstream.

paulkaplan:
@rachel-fenichel the angle picker is a custom scratch field, but there may be other places where draggable images come into play?

rachel-fenichel:
The angle picker has its base in Blockly, though it doesn't look like the Blockly one has an image specifically.

They have had tightly coupled weirdness around dragging in the past.
image

PR #1772, opened 2018-11-07
Fix local compile on Windows

### LINKED ISSUE #1523: Can't run build.py from the command line
### Expected Behavior

Running `build.py` from the command line should build the various uncompressed and compressed versions of scratch blocks.

It should use the local compiler if possible, and otherwise fall back to the remote compiler.

### Actual Behavior

Before running `npm install` (so no npm package for google-closure-compiler), or after deleting the contents of `node_modules`:
There are two bugs, fixed by #1512 and #1513

After running `npm install`:
```
./build.py
  File "./build.py", line 595, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

Note that it works as part of the `npm prepublish` command.

I will add more information as I have it.

cwillisf:
### Resolves

Resolves #1523

### Proposed Changes

This change sets the `shell=True` flag in the two places we use `subprocess.Popen` to run the Closure Compiler locally.

### Reason for Changes

We call the local compiler with `subprocess.Popen`, passing `google-closure-compiler` as the command to run. On Windows, the actual command is `google-closure-compiler.cmd` -- the extension is automatically added by the shell. Unfortunately `subprocess.Popen` by default bypasses the shell, leading to an error on Windows. Setting `shell=True` makes `Popen` use the Windows shell (`cmd`) and correctly find `google-closure-compiler.cmd`.

### DIFF
--- a/build.py
+++ b/build.py
@@ -323,7 +323,7 @@ class Gen_compressed(threading.Thread):
       for group in [["google-closure-compiler"], dash_args]:
         args.extend(filter(lambda item: item, group))
 
-      proc = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+      proc = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, shell=True)
       (stdout, stderr) = proc.communicate()
 
       # Build the JSON response.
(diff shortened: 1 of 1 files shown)


rachel-fenichel (approved):

cwillisf:
It appears that while this resolves #1523 it's not quite enough to get the locally-compiled build working on Windows. To be continued....

PR #1775, opened 2018-11-08
Use parent block to set colors in the matrix field.

### LINKED ISSUE #1764: field_matrix does not use colors of the block containing it

picklesrus:
Also, call updateMatrix_ explicitly rather than through setValue.
setValue is supposed to just manipulate the model (though it often does both).
In this case doing both doesn't work because the of the initialization order of things.

Partially resolves https://github.com/LLK/scratch-blocks/issues/1764, though not the part where it is used through extensions.

See discussion on the issue for why it was done this way.

### DIFF
--- a/core/field_matrix.js
+++ b/core/field_matrix.js
@@ -272,7 +272,6 @@ Blockly.FieldMatrix.prototype.setValue = function(matrix) {
   }
   matrix = matrix + Blockly.FieldMatrix.ZEROS.substr(0, 25 - matrix.length);
   this.matrix_ = matrix;
-  this.updateMatrix_();
 };
 
 /**
@@ -328,7 +327,7 @@ Blockly.FieldMatrix.prototype.showEditor_ = function() {
   // Button to clear matrix
   var clearButtonDiv = document.createElement('div');
   clearButtonDiv.className = 'scratchMatrixButtonDiv';
-  var clearButton = this.createButton_(this.sourceBlock_.colourSecondary_);
+  var clearButton = this.createButton_(this.sourceBlock_.parentBlock_.colourSecondary_);
   clearButtonDiv.appendChild(clearButton);
   // Button to fill matrix
   var fillButtonDiv = document.createElement('div');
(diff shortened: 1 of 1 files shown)


picklesrus:
FYI @ErikMejerHansen

ericrosenbaum (changes requested):
I was testing this in the playground and noticed an issue: when you drop a reporter onto the input, and then pull it out, the field becomes blank, and I see a console error: `Cannot read property 'colourSecondary_' of null`

picklesrus:
Aww, that's when we respawn shadows. Good catch! I'll have to think about this a little more.

PR #1777, opened 2018-11-10
Fix non-deterministic build error #1409

### LINKED ISSUE #1409: non-deterministic build caused error
We are using CI to build the package, we find that packages built by two of our runners are not working, while other runners produces expected results. Here is the error with bad ones: 

```
Uncaught TypeError: Cannot set property 'ProcedureUtils' of undefined
    at Object.module.exports.$jscomp (vertical.js:779)
    at __webpack_require__ (vertical.js:21)
    at Object.<anonymous> (vertical.js:759)
    at __webpack_require__ (vertical.js:21)
    at Object.<anonymous> (vertical.js:103)
    at __webpack_require__ (vertical.js:21)
    at Object.<anonymous> (vertical.js:95)
    at __webpack_require__ (vertical.js:21)
    at vertical.js:64
    at Object.defineProperty.value (vertical.js:67)
```

After digg into `build.py`, we find that [`gen_blocks()`](https://github.com/LLK/scratch-blocks/blob/develop/build.py#L224-L232) didn't sort filenames outputted by `glob.glob()`. And the two problematic runners have the following order of filenames:

```
['blocks_vertical/procedures.js', 'blocks_vertical/looks.js', 'blocks_vertical/sound.js', 'blocks_vertical/extensions.js', 'blocks_vertical/control.js', 'blocks_vertical/operators.js', 'blocks_vertical/vertical_extensions.js', 'blocks_vertical/event.js', 'blocks_vertical/pen.js', 'blocks_vertical/motion.js', 'blocks_vertical/default_toolbox.js', 'blocks_vertical/sensing.js', 'blocks_vertical/data.js']
```

`blocks_vertical/procedures.js` comes first. When it [creates namespace `Blockly.ScratchBlocks.ProcedureUtils`](https://github.com/LLK/scratch-blocks/blob/develop/blocks_vertical/procedures.js#L33), `Blockly.ScratchBlocks` was still undefined, which causes the error.

Suggesting:
* Sort `filenames` in `build.py:gen_blocks()` , to ensure deterministic build.
* Create namespace properly in `blocks_vertical/procedures.js`

tansly:
### Resolves

This fixes #1409, at least on my machine.

### Proposed Changes

By the time `Blockly.ScratchBlocks.ProcedureUtils` was defined, `Blockly.ScratchBlocks` had not been defined. This makes sure `Blockly.ScratchBlocks` is defined.

### Reason for Changes

This is the first line of JavaScript code that I've ever written in my life, so I'm not even sure if this is the right way to define the namespace. #1409 was blocking me from working on scratch-gui and scratch-blocks. Seems like this change fixed the error.

### Test Coverage

No tests added.

### DIFF
--- a/blocks_vertical/procedures.js
+++ b/blocks_vertical/procedures.js
@@ -30,6 +30,7 @@ goog.require('Blockly.constants');
 
 
 // TODO: Create a namespace properly.
+Blockly.ScratchBlocks = Blockly.ScratchBlocks || {};
 Blockly.ScratchBlocks.ProcedureUtils = {};
 
 // Serialization and deserialization.


picklesrus:
Closing since I think #1896 fixed this as well.

PR #1780, opened 2018-11-10
Support scientific notation in numerical inputs

### LINKED ISSUE #1301: Support scientific notation in numeric inputs
### Expected Behavior

In Scratch 2.0, number-only inputs (the rounded ones) allow you to type the letter 'e', in addition to numbers and decimal points. You can for example type 1e6 (to represent one million) into any input, including number-only inputs.  

### Actual Behavior

It appears that Scratch 3.0 generally supports the conversion from scientific notation, and you can type it into fields that are number-or-string, but you cannot type it into number-only fields. This is a bit confusing, because we have removed the visual distinction between these field types, so there is no way to tell by looking which fields have this constraint. So for example, in Scratch 3.0 you can type 1e6 into the "set variable to" block (a number-or-string input) but not into the "change variable by" block (a number-only input). You can, though, drop a reporter with the value 1e6 into the "change variable by" block and it will work as expected. 

It's not obvious to me if we need to support this pattern or not, but as of now this is a difference from Scratch 2.0. So, should we add support for typing the 'e' into number-only inputs?

towerofnix:
### Resolves

Resolves #1301.

### Proposed Changes

Allows the letter "e" to be inputted into number inputs.

### Reason for Changes

To support scientific notation inside number inputs, which is a feature in Scratch 2.0.

### Test Coverage

Tested manually in blocks and gui with the following script:

The script "when green flag clicked, go to x: -1e2 y: 0, glide 1 secs to x: 1e2 y: 0, say 2.5e3 / 2"

~~It looks like I might be able to add tests in `tests/jsunit/field_number_test.js`, but I'm not 100% sure.~~ I think I would need to figure out how to get Selenium running locally, which seems to be sort of a big task - so I can't add unit tests myself.

Aside: It's possible to input "e123e7e1eeeee", but this is no different from Scratch 2.0 behavior. You can also do the same thing with the negative sign ("-") already. (Interestingly, Scratch 1.4 will do some verification in number inputs; no such in later versions. Also, 2.0 is the first version with support for scientific notation in number inputs.)

### DIFF
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -135,6 +135,9 @@ Blockly.FieldNumber.prototype.getNumRestrictor = function(opt_min, opt_max,
   if (this.negativeAllowed_) {
     pattern += "|[-]";
   }
+  if (this.exponentialAllowed_) {
+    pattern += "|e";
+  }
   return new RegExp(pattern);
 };
 
@@ -151,6 +154,7 @@ Blockly.FieldNumber.prototype.setConstraints_ = function(opt_min, opt_max,
       (Math.floor(opt_precision) != opt_precision);
   this.negativeAllowed_ = (typeof opt_min == 'undefined') || isNaN(opt_min) ||
       opt_min < 0;
+  this.exponentialAllowed_ = this.decimalAllowed_;
 };
 
 /**
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -136,7 +136,7 @@ Blockly.FieldNumber.prototype.getNumRestrictor = function(opt_min, opt_max,
     pattern += "|[-]";
   }
   if (this.exponentialAllowed_) {
-    pattern += "|e";
+    pattern += "|[eE]";
   }
   return new RegExp(pattern);
 };


core/field_number.js
```diff
@@ -135,6 +135,9 @@ Blockly.FieldNumber.prototype.getNumRestrictor = function(opt_min, opt_max,
   if (this.negativeAllowed_) {
     pattern += "|[-]";
   }
+  if (this.exponentialAllowed_) {
+    pattern += "|e";
+  }
   return new RegExp(pattern);
```
joker314, line 141: This looks good! One thing to note is that Scratch 2.0/3.0 handle 'e' in a case-insensitive way when evaluating inputs. This is because the ECMAScript `Number()` function is case-insensitive when dealing with 'e'.

It seems a little inconvenient to need to remember whether caps-lock is on when inputting a number... especially since Scratch tries hard to hide case from the user.

Since it's not just 'e', but all of Scratch, that's case sensitive, I think it's best to change the RegEx flag rather than just a component of the regex, to make changes involving letters faster in future.

I tested both capital and lowercase 'e's using the join block and they are equivelant

Warning: the suggestion below is untested
```suggestion
   return new RegExp(pattern, 'i');
```

core/field_number.js
```diff
@@ -151,6 +154,7 @@ Blockly.FieldNumber.prototype.setConstraints_ = function(opt_min, opt_max,
       (Math.floor(opt_precision) != opt_precision);
   this.negativeAllowed_ = (typeof opt_min == 'undefined') || isNaN(opt_min) ||
       opt_min < 0;
+  this.exponentialAllowed_ = this.decimalAllowed_;
```
towerofnix, line 157: Not 110% sure it makes sense to use a flag for this (`exponentialsAllowed_`) like with the other values, but it's here for code consistency. I made it equal to `decimalAllowed_` because if you *can't* input decimals, you shouldn't be able to input `e` either, so that you can't write (for example) `7e-1`, which is effectively a decimal value (0.7).

joker314 (approved):
Nice PR! Just a question about case-sensitivity.

core/field_number.js
towerofnix: Yes, I thought of this. While testing in 2.0 I noticed that if you try to type `3E4`, it will automatically replace the `E` as you press it with `e`. That's a bit wonky to me; I think just allowing capital E makes more sense. I'll leave it up to a (scratch team :P) PR reviewer to decide, though (but yeah, your change should work; otherwise, appending `|[eE]` to the regex will work too).

joker314: Just a note that it's probably not too hard to match 2.0's behaviour if that's what's decided, according to [Blockly documentation](https://developers.google.com/blockly/reference/js/Blockly.FieldNumber) (my emphasis):

> An optional function that is called to validate any constraints on what the user entered. Takes the new text as an argument and returns either the accepted text, ***a replacement text***, or null to abort the change.

The advantage of 2.0's way is that a person who is not familiar with scientific notation will not be confused by the (lack of a) distinction in "5E3" and "5e3".

paulkaplan:
@towerofnix I think allowing `|[eE]` is a good idea for this. Other than that, looks good!

towerofnix:
@paulkaplan Got it - commit made!

PR #1781, opened 2018-11-11
Make insertion markers for define blocks have the correct width

### LINKED ISSUE #1186: Define block insertion marker is the wrong width
### Expected Behavior

Define block insertion marker should be the same width as the fully rendered define block, including the width of the internal block.

### Actual Behavior

image

### Steps to Reproduce

Disconnect the define block from everything else, then drag it near a stack of blocks.

All

ErikMejerHansen:
### Resolves

#1186 

### Proposed Changes

Make insertion markers for define blocks have to correct width. 

### Reason for Changes

Trying to fix #1186

### Test Coverage

No tests added

### DIFF
--- a/core/block_render_svg_vertical.js
+++ b/core/block_render_svg_vertical.js
@@ -1606,6 +1606,7 @@ Blockly.BlockSvg.prototype.renderDefineBlock_ = function(steps, inputRows,
         input.connection.targetBlock().getHeightWidth().width +
         Blockly.BlockSvg.DEFINE_BLOCK_PADDING_RIGHT;
   } else {
+    // Handles the case where block is being rendered as an insertion marker
     rightSide = Math.max(Blockly.BlockSvg.MIN_BLOCK_X_WITH_STATEMENT, rightSide)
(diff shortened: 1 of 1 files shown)


core/block_render_svg_vertical.js
```diff
@@ -1606,7 +1606,8 @@ Blockly.BlockSvg.prototype.renderDefineBlock_ = function(steps, inputRows,
         input.connection.targetBlock().getHeightWidth().width +
         Blockly.BlockSvg.DEFINE_BLOCK_PADDING_RIGHT;
   } else {
-    rightSide = Blockly.BlockSvg.MIN_BLOCK_X_WITH_STATEMENT;
+    rightSide = Math.max(Blockly.BlockSvg.MIN_BLOCK_X_WITH_STATEMENT, rightSide)
```
rachel-fenichel, line 1610: Add a comment that this case handles the block being rendered as an insertion marker (in all other cases there should be a target block connected on the statement input).

picklesrus:
Thanks @ErikMejerHansen and @rachel-fenichel!

PR #1782, opened 2018-11-13
Track cloud variable info

kchadha:
### Proposed Changes

Track information about cloud variables and handle adding the cloud unicode character to the variable.

### Reason for Changes

Allows for the creation of cloud variables from the variable modal.

### Test Coverage

Testing instructions included in the related scratch-gui PR.

### Related PRs
- LLK/scratch-gui#3749
- LLK/scratch-vm#1755

### DIFF
--- a/core/variable_events.js
+++ b/core/variable_events.js
@@ -89,6 +89,7 @@ Blockly.Events.VarCreate = function(variable) {
   this.varType = variable.type;
   this.varName = variable.name;
   this.isLocal = variable.isLocal;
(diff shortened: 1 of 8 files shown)


paulkaplan (changes requested):
Changes look good, but you need to fix the unit tests for varCreate in tests/jsunit/event_tests, lines 378 and 437 need `isCloud: false` in the expected json.

Once that is fixed, LG

paulkaplan (approved):

PR #1786, opened 2018-11-15
Add "random direction" to "point towards" dropdown

joker314:
### Resolves

Resolves LLK/scratch-gui#1702
Sibling pull request: LLK/scratch-vm#1748
/cc @paulkaplan 

### Proposed Changes

Add "random direction" to dropdown menu for "point towards" (The commit message subject is a typo)

### Reason for Changes

(This is **not** a compatibility change, for clarity)

See the original issue for discussion about wording. "Point towards random direction" is not 'good' English, but it apparently made more intuitive sense than "Point towards random position". Notice that the former and the latter have different meanings. If you're in the top-right quadrant, and you pick a random position, most likely, that position will be to your bottom left. This means it's not a 'random direction', but rather a 'weighted random direction based on position'.

This VM will have an equal chance of pointing in all directions, perhaps the wording to reflect that.

Anyway, it seems the wording was already decided upon earlier.

### Test Coverage

I use Windows. I tried to set up the WSL and all, because it just wouldn't build otherwise. I used the static playground to make sure the dropdown appears. I was unable to see if this change works with the rest of the components of the editor, I'm afraid. (Manual testing only)


joker314:
Sorry, but running `npm run test:messages` causes `msg/json/en.js` and `msg/js/en.js` to both empty themselves into empty objects. I doubt that's the desired behaviour, I'm likely doing something wrong, or my set-up is wrong.

chrisgarrity:
@joker314 it's a limitation of the translation system for scratch-blocks that changes to `messages.js` need intervention from the scratch team. So you're not doing anything wrong. I've just pushed a commit to your branch that updates the translations with the new string. See if you can test now.

PR #1787, opened 2018-11-15
Update scratch-l10n to the latest version

chrisgarrity:
Need to update the version of scratch-l10n because a language (Croatian) was added.


benjiwheeler:
LGTM!

PR #1803, opened 2018-12-04
Fix switching variables via context menu 

### LINKED ISSUE #2944: Switching variable reference in context menu only changes appearance, doesn't actually update the variable reference
### Expected Behavior

Switching a variable reference using the variable reporter's context menu option should actually update the variable reference. 

### Actual Behavior

The variable reporter changes in appearance but still references the old variable it was referring to.
variable_refs

Found by a scratcher in the forums.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/2944

### Proposed Changes

Looks like this code was not updated when we moved from variable names to IDs. So just move the callback code to use IDs and `setValue` instead of names and `setText`

I was able to test this (after fixing the playground) by turning on the workspace event logging and noting that before this change, no "change" event was being emitted when the context menu was used. But after this change, the correct "change" event is emitted when using the context menu.

I also tested this by embedding it in the GUI and making sure the reported value from stack clicks was correct after switching variables.

### DIFF
--- a/core/variables.js
+++ b/core/variables.js
@@ -299,6 +299,7 @@ Blockly.Variables.createVariable = function(workspace, opt_callback, opt_type) {
   // Prompt the user to enter a name for the variable
   Blockly.prompt(newMsg, '',
       function(text, additionalVars, variableOptions) {
+        variableOptions = variableOptions || {};
         var scope = variableOptions.scope;
         var isLocal = (scope === 'local') || false;
         var isCloud = variableOptions.isCloud || false;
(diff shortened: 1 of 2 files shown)


core/variables.js
```diff
@@ -299,6 +299,7 @@ Blockly.Variables.createVariable = function(workspace, opt_callback, opt_type) {
   // Prompt the user to enter a name for the variable
   Blockly.prompt(newMsg, '',
       function(text, additionalVars, variableOptions) {
+        variableOptions = variableOptions || {};
```
kchadha, line 302: Thanks for fixing this! I probably didn't test the playground when making changes here for cloud variables.

kchadha (approved):
LGTM, thanks for fixing this!

PR #1807, opened 2018-12-07
Use variable name as default value for rename prompts

### LINKED ISSUE #3983: "Rename variable" prompt should have current name as default value
### Expected Behavior

When a user opens a prompt to rename a variable, the current name of the variable should be the default value in the text box. This would be useful for, for example, variables with long names where the user only wants to edit one part, or variable names with emojis where the user does not want to copy and paste every emoji they used.

### Actual Behavior

There is no default value.

### Steps to Reproduce

1. Make a variable
2. Rename the variable

Windows NT 10.0, Chrome 70

towerofnix:
### Resolves

Together with LLK/scratch-gui#3984, resolves LLK/scratch-gui#3983. Should be merged **after** that PR.

### Proposed Changes

Passes the variable name as the default value for the rename prompt.

### Reason for Changes

So that it's easier to make a quick modification to a variable name (e.g. if you made a typo while creating the variable).

### Test Coverage

Tested manually.

### DIFF
--- a/core/variables.js
+++ b/core/variables.js
@@ -496,8 +496,8 @@ Blockly.Variables.renameVariable = function(workspace, variable,
   Blockly.prompt(promptText, promptDefaultText,
       function(newName, additionalVars) {
         if (variable.isCloud &&
-            newName.length > 0 && newName.indexOf(Blockly.Variables.CLOUD_PREFIX) == 0 ) {
-          newName = newName.substring(2); // The name validator will add the prefix back
+            newName.length > 0 && newName.indexOf(Blockly.Variables.CLOUD_PREFIX) == 0) {
+          newName = newName.substring(Blockly.Variables.CLOUD_PREFIX.length); // The name validator will add the prefix back
         }
         additionalVars = additionalVars || [];
         var additionalVarNames = variable.isLocal ? [] : additionalVars;
--- a/core/variables.js
+++ b/core/variables.js
@@ -497,7 +497,8 @@ Blockly.Variables.renameVariable = function(workspace, variable,
       function(newName, additionalVars) {
         if (variable.isCloud &&
             newName.length > 0 && newName.indexOf(Blockly.Variables.CLOUD_PREFIX) == 0) {
-          newName = newName.substring(Blockly.Variables.CLOUD_PREFIX.length); // The name validator will add the prefix back
+          newName = newName.substring(Blockly.Variables.CLOUD_PREFIX.length);
+          // The name validator will add the prefix back
         }
         additionalVars = additionalVars || [];
         var additionalVarNames = variable.isLocal ? [] : additionalVars;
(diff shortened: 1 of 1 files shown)


core/variables.js
```diff
@@ -488,7 +488,7 @@ Blockly.Variables.renameVariable = function(workspace, variable,
   var validate = Blockly.Variables.nameValidator_.bind(null, varType);
 
   var promptText = promptMsg.replace('%1', variable.name);
-  Blockly.prompt(promptText, '',
+  Blockly.prompt(promptText, variable.name,
```
kchadha: I think that for cloud variables, the default value in the prompt should be the variable name without the cloud prefix; this is slightly different from the scratch 2.0 behavior, but I think would ultimately lead to less confusion. Proposing code changes below:

kchadha:
```suggestion
  var promptDefaultText = variable.name;
  if (variable.isCloud && variable.name.indexOf(Blockly.Variables.CLOUD_PREFIX) == 0) {
    promptDefaultText = promptDefaultText.substring(Blockly.Variables.CLOUD_PREFIX.length);
  }

  Blockly.prompt(promptText, promptDefaultText,
```

kchadha: And even though this isn't directly related to this PR, maybe [below,](https://github.com/LLK/scratch-blocks/pull/1807/files#diff-5566718ada8dbcce77f03a181262090dR495) you can make a similar optimization to existing code where we use `newName.substring(2)`, that should probably also change to `newName.substring(Blockly.Variables.CLOUD_PREFIX.length)` so that the `2` isn't hardcoded.

kchadha (changes requested):
Minor change requested

core/variables.js
towerofnix: Sure, sounds good! I didn't think of cloud variables - good point. I'll make the constant fix (`2`) below.

towerofnix:
@kchadha Accepted your suggested changes and fixed the code that was assuming the value of `CLOUD_PREFIX.length`, should be ready for re-review.

kchadha:
@towerofnix, thanks for making those changes! It looks like that last commit introduced a lint error for the line of code being too long. Maybe move the comment onto a separate line?

core/variables.js
```diff
@@ -488,11 +488,16 @@ Blockly.Variables.renameVariable = function(workspace, variable,
...
+    promptDefaultText = promptDefaultText.substring(Blockly.Variables.CLOUD_PREFIX.length);
+  }
+
+  Blockly.prompt(promptText, promptDefaultText,
       function(newName, additionalVars) {
         if (variable.isCloud &&
-            newName.length > 0 && newName.indexOf(Blockly.Variables.CLOUD_PREFIX) == 0 ) {
-          newName = newName.substring(2); // The name validator will add the prefix back
+            newName.length > 0 && newName.indexOf(Blockly.Variables.CLOUD_PREFIX) == 0) {
+          newName = newName.substring(Blockly.Variables.CLOUD_PREFIX.length); // The name validator will add the prefix back
```
kchadha:
```suggestion
          newName = newName.substring(Blockly.Variables.CLOUD_PREFIX.length);
          // The name validator will add the prefix back
```

towerofnix: Oops! Thanks.

kchadha (approved):
LGTM -- thank you!

Issue #1784, opened 2018-11-14
On Mac Safari, note picker drop shadow is black


ericrosenbaum:
### Expected Behavior

The drop shadow at the top of the piano keys in the note picker should be a pleasing translucent gray.

### Actual Behavior

On Mac Safari, the drop shadow is black:

![screen shot 2018-11-14 at 11 30 30 am](https://user-images.githubusercontent.com/567844/48496919-12a1b900-e801-11e8-90ff-dd0791f4e78a.png)

### Operating System and Browser

Mac OS 10.13.6 Safari 12.0.1

PR #1811, opened 2018-12-10
Use fill-opacity for notepicker shadow

epicfaace:
### Resolves

On Mac Safari, note picker drop shadow is black #1784

### Proposed Changes

Use fill-opacity instead of rgba to create the drop shadow.

### Reason for Changes

It looks like Safari doesn't support putting an rgba color in the "fill" attribute for svg elements. See https://stackoverflow.com/questions/6042550/svg-fill-color-transparency-alpha/6047978

### Test Coverage

None


picklesrus:
Thanks @epicfaace!

Issue #1783, opened 2018-11-14
On Android, number picker makes the note picker appear half-width


ericrosenbaum:
### Expected Behavior

The note picker field should always be the same size.

### Actual Behavior
 
On Android, sometimes it appears cut off, at only half the correct width. It seems to happen after you've used the number picker, i.e. to set the number of beats on the same block.

![image16](https://user-images.githubusercontent.com/1088291/48427278-672c3200-e736-11e8-82bd-0d3053d7492e.png)

Reported by @benjiwheeler 

### Steps to Reproduce

Use Android Tablet Chrome
Start with new default project
add music extension
drag out play note block
tap on “60” (note field) -- width of picker is good!
tap on “0.25” (beats field) and bring up number picker
tap away from the number picker to dismiss
tap on “60” (note field) -- width of picker is too narrow!

### Operating System and Browser

 Android Tablet Chrome

samelhusseini:
The issue here is that some field editors set the width on the dropdown's contentDiv but never clear it. 
I've fixed this on our side recently by clearing the dropdown div's content width when the dropdown is hidden. 

See: https://github.com/Microsoft/pxt-blockly/pull/171

PR #1812, opened 2018-12-10
Clear dropdown width to fix half-width bug

epicfaace:
### Resolves

On Android, number picker makes the note picker appear half-width #1783

### Proposed Changes

Clear dropdown width when it is hidden

### Reason for Changes

See https://github.com/LLK/scratch-blocks/issues/1783#issuecomment-438519140

### Test Coverage

None


picklesrus:
Thanks @epicfaace!

PR #1814, opened 2018-12-10
Deletion of single hat blocks connected to a stack

### LINKED ISSUE #245: Deletion behavior on right-clicking the first block
Currently, the context menu indicates that only one block will be deleted. In fact, the entire stack is deleted! :( Either we should fix the menu or fix the deletion behavior.

### LINKED ISSUE #1776: Deleting a hat block incorrectly implies deleting a single block
### Expected Behavior
When you right click on a block and you select the option `Delete block`, it should delete one block.

### Actual Behavior
delete

Right clicking on a hat block uses the message `Delete block` but it deletes the whole stack. Either the message should indicate that more blocks will be deleted, or only delete the hat block.

Note clicking on the forever when it contains a block the context menu says `Delete 2 blocks`.
MacOS Chrome

epicfaace:
### Resolves

Deletion behavior on right-clicking the first block - fixes #245, fixes #1776

### Proposed Changes

When deleting a block with no previousConnection (i.e. a hat block), this change allows for disconnection of the hat block from the rest of the stack. Thus, only the hat block is deleted.

### Reason for Changes

"Delete this block" should delete only a single block, even if it's a hat block.

### Test Coverage

None

### DIFF
--- a/core/block.js
+++ b/core/block.js
@@ -339,8 +339,7 @@ Blockly.Block.prototype.unplug = function(opt_healStack) {
       // Disconnect from any superior block.
       this.outputConnection.disconnect();
     }
-  }
-  else {
+  } else {
     if (this.previousConnection) {
       var previousTarget = null;
       if (this.previousConnection.isConnected()) {
--- a/core/block.js
+++ b/core/block.js
@@ -339,13 +339,16 @@ Blockly.Block.prototype.unplug = function(opt_healStack) {
       // Disconnect from any superior block.
       this.outputConnection.disconnect();
     }
-  } else if (this.previousConnection) {
-    var previousTarget = null;
-    if (this.previousConnection.isConnected()) {
-      // Remember the connection that any next statements need to connect to.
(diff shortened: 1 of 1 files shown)


towerofnix:
I believe this also fixes #1776 (although I haven't tested it). (Aside, I recommend editing the "resolves" section in the PR description to include phrases like "fixes #1776" and "fixes #245", so that GitHub automatically closes those issues once this is merged.)

picklesrus:
Hi again @epicfaace.  Thanks for all of these recent PRs! They're great!  I played with this a little yesterday, and while I think it is right, I'm going to hold off reviewing further for now.  We have quite a bit landing right now and are attempting to only land critical or minor things between now and the 3.0 launch.  I'll come back to this in January!

epicfaace:
Makes sense @picklesrus , sounds good!

picklesrus:
LGTM. I made a tiny change to put the else back up a line but I'll merge once travis finishes running its checks.  Thank you!!

PR #1815, opened 2018-12-10
Fix extra events being emitted from the color slider

paulkaplan:
### Resolves

Related to LLK/scratch-vm#1695, we were getting echo events from setting the initial slider values. 

Move the call to set those values above the callbacks being bound to avoid this.


kchadha (changes requested):
Hmmm for some reason this change makes it so that the slider thumbs are not initialized to the correct positions when the field editor is opened...

paulkaplan:
Fixed, thanks for catching that.

kchadha (approved):
LGTM

kchadha:
Merging since I need these changes for the PR I am about to submit.

PR #1816, opened 2018-12-11
Conditionally turn off slider handle callbacks.

kchadha:
### Resolves

Resolves LLK/scratch-vm#1695

### Proposed Changes

Turn off slider callbacks from triggering when getting called as part of `updateSliderHandles_` function (which only gets called when we want to make the slider handles reflect the already set value of the field).

### Reason for Changes

The slider callbacks were getting called, resulting in another translation of the original color value (looks purple chosen by the eye dropper), distorting the output color value (original: #9966ff, distorted translation: #9666ff) to the point where scratch-vm was reporting isTouchingColor as false when it should have been true. 

### Test Coverage

Tested with the above issue when connected to scratch-gui.


paulkaplan (approved):
LG. @BryceLTaylor lets make sure to get some 👀 on this for testing

PR #1823, opened 2018-12-17
Make all number inputs LTR

### LINKED ISSUE #1707: In RTL, while editing a number field, the negative sign is on the incorrect side
### Expected Behavior

The negative sign should always appear on the correct side while typing in a number.

### Actual Behavior

In RTL, while editing a number, the negative sign incorrectly appears on the right side. Once the edit is accepted, the negative sign moves to the correct side (after https://github.com/LLK/scratch-blocks/pull/1702):

rtl-negative-sign

In a report from an educator for the Perisian language, this bug results in young people thinking they need to type 10- in order for it to display correctly. Then the value is wrong and doesn't do what they expect.

### LINKED ISSUE #1708: Editing number inputs have the minus sign in the wrong place (right) in RTL mode
### Expected Behavior
Numbers and math in general are still LTR in RTL languages (e.g. Hebrew)

Number inputs should still have the minus sign on the left.

related to https://github.com/google/blockly/issues/1802

### Actual Behavior

When the input is selected, the minus sign switches to the right. 
If you type a new negative value the minus sign is on the right until you hit enter or click outside. 

### Steps to Reproduce

Switch into Hebrew and then type a negative number in an input

### LINKED ISSUE #1709: Angle picker number input has the minus sign on the wrong side (right) in RTL
### Expected Behavior

Numbers are still LTR in RTL languages (e.g. Hebrew).
When the angle picker switches to a negative input it should still have the minus sign on the left.

Note, if you type into the angle value (instead of using the picker) the minus sign also ends up on the right when it should be on the left.

### Actual Behavior

The minus sign is on the right as you move the angle picker, and the value is saved with the minus on the right after editing.

### Steps to Reproduce

Use the angle picker in Hebrew

epicfaace:
### Resolves

Resolves #1709, resolves #1708, resolves #1707 

### Proposed Changes

Make all number inputs LTR

### Reason for Changes

Number inputs should always be LTR, even when an RTL language is being used.

### Test Coverage

None

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -233,6 +233,10 @@ Blockly.FieldTextInput.prototype.showEditor_ = function(
   if (readOnly) {
     htmlInput.setAttribute('readonly', 'true');
   }
+  if (this.sourceBlock_.type === 'math_number') {
+    // All number fields are LTR.
+    htmlInput.setAttribute('dir', 'ltr');
+  }
   /** @type {!HTMLInputElement} */
   Blockly.FieldTextInput.htmlInput_ = htmlInput;
   div.appendChild(htmlInput);


picklesrus:
The fix works, but conditionally checking for math_number in field is a bit of a hack.  We want to remove the places we're already doing it (see discussion on https://github.com/google/blockly/issues/1802 if you're interested in details).  The fix really belongs in field_number, though might not be super straightforward because field number doesn't currently override the right methods.

picklesrus:
Going to close this for now given my previous comment. Thanks!

PR #1826, opened 2018-12-17
Duplicated block goes to mouse pointer, not corner of block

### LINKED ISSUE #1006: Duplicated block is placed at the upper left corner of the initial block, instead of at the mouse location.
### Expected Behavior

Duplicated blocks that attach to the mouse pointer should have the mouse pointer placed correctly relative to the upper left corner of the block.

### Actual Behavior

The block is positioned according to the upper left corner of the block that was duplicated, and then there's a jump the first time the mouse moves.

(After #1001 lands)

### Steps to Reproduce

Create a block
Right click the block in the lower right corner.
Select "Duplicate"
Move the mouse.

All

epicfaace:
### Resolves

Resolves #1006 ("Duplicated block is placed at the upper left corner of the initial block, instead of at the mouse location.")

### Proposed Changes

When someone duplicates a block, pass in the actual mouse event to the `duplicateAndDragCallback` function, so that the block starts at the location where the context menu was clicked (rather than the top-left or top-right corner)

### Reason for Changes

More intuitive

### Test Coverage

None

### DIFF
--- a/core/block_svg.js
+++ b/core/block_svg.js
@@ -673,11 +673,9 @@ Blockly.BlockSvg.prototype.showContextMenu_ = function(e) {
   // Save the current block in a variable for use in closures.
   var block = this;
   var menuOptions = [];
-
-  var isMouseEvent = Blockly.Touch.getTouchIdentifierFromEvent(e) == 'mouse';
   if (this.isDeletable() && this.isMovable() && !block.isInFlyout) {
     menuOptions.push(
-        Blockly.ContextMenu.blockDuplicateOption(block, isMouseEvent));
+        Blockly.ContextMenu.blockDuplicateOption(block, e));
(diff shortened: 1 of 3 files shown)


towerofnix:
Just to be totally sure, does this still work when the browser is zoomed in? When scratch-blocks itself is zoomed in? When you've scrolled round the workspace? (And zoomed in?)

epicfaace:
It does when I test it.

rachel-fenichel:
LGTM as long as it works in the cases @towerofnix mentioned.  Thanks @epicfaace!

picklesrus:
Nice simplification!  I'm a little confused why we did this the complicated way before...I looked through the events emitted when you duplicated and I think they're right too. I'm going to run it by one more person tomorrow to see if they can think of any reasons why we can't just use the event x & y.

PR #1828, opened 2018-12-17
Add restrictor to disallow text being typed in "direction" block input

### LINKED ISSUE #1779: Text can be typed into the "direction" block input
### Expected Behavior

You can't type text into the "direction" input.

### Actual Behavior

You can type any text into the direction input (i.e. in "point in direction"). The input will display with a red background and reset to the previous value if you press enter or de-focus the input.

### Steps to Reproduce

Drag out a block with a direction input, i.e. "point in direction", and type a non-digit value (e.g. "xyz").

Note that this is separate from the issue that you can type numbers into the scratch-gui "direction" input in the sprite info area (and into any other input there, e.g. x/y position).

(May be worth considering #1301 in fixing this. I'm taking a stab at that issue for fun (no guarantees I'll get anywhere).)

Firefox 64.0b6, Debian (Linux).

epicfaace:
### Resolves

Resolves #1779 - Text can be typed into the "direction" block input

### Proposed Changes

Add a restrictor that allows only numeric characters to be typed into the input

### Reason for Changes

Text should not be able to be typed into the "direction" block input

### Test Coverage

None

### DIFF
--- a/core/field_angle.js
+++ b/core/field_angle.js
@@ -47,10 +47,12 @@ Blockly.FieldAngle = function(opt_value, opt_validator) {
   // Add degree symbol: '360°' (LTR) or '°360' (RTL)
   this.symbol_ = Blockly.utils.createSvgElement('tspan', {}, null);
   this.symbol_.appendChild(document.createTextNode('\u00B0'));
+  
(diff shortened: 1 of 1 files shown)


picklesrus:
LGTM!  Sorry it took me so long to test this!

PR #1832, opened 2018-12-19
Use webdriver-selenium for tests

### LINKED ISSUE #1569: Use npm packages for selenium/chromedriver so they don't go out-of-date
Having the chromedriver and selenium configs in custom scripts is different from our other repos, where we use the npm packages that handle downloading, starting, stopping and correctly setting the path for the respective parts of this system.

If we don't do this, we will have another https://github.com/LLK/scratch-blocks/issues/1567 in a short while

epicfaace:
### Resolves

Fixes #1569 - Use npm packages for selenium/chromedriver so they don't go out-of-date

### Proposed Changes

Use webdriver-selenium for tests


epicfaace:
@thisandagain bump

thisandagain:
@picklesrus Any thoughts on this?

chrisgarrity (approved):
As predicted in #1569, chromedriver updates caused the tests to fail again. It's a little worrying that the version of the selenium webdriver is labelled 'alpha', but that's the version that's been out for a year or so. However, after updating the version of chromedriver to the latest, tests run correctly. I think the tests could be modified to wait for the results to appear instead of always waiting 5 seconds, but that's not worth the effort right now. Thanks @epicfaace for submitting this. Sorry we took a long time to land it.

PR #1833, opened 2018-12-19
Make refreshToolboxSelection_ package instead of private

### LINKED ISSUE #1354: Update refreshToolboxSelection
Update refreshToolboxSelection function and it's use case in procedures.js after it's been [updated in upstream blockly](https://github.com/google/blockly/issues/1554) to be made package instead of private.

This function is used in procedures.js as part of [fix](https://github.com/LLK/scratch-blocks/pull/1355) for [scratch-gui#1261](https://github.com/LLK/scratch-gui/issues/1261)

epicfaace:
### Resolves

Fixes #1354, Update refreshToolboxSelection

### Proposed Changes

Make refreshToolboxSelection_ a package method and rename it to refreshToolboxSelection.

### Reason for Changes

See the blockly issue - https://github.com/google/blockly/issues/1554

### DIFF
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -289,7 +289,7 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
         }
       }
       // The proc deletion was valid, update the toolbox.
-      ws.refreshToolboxSelection_();
+      ws.refreshToolboxSelection();
     });
   }
 };
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -568,10 +568,8 @@ Blockly.Procedures.deleteProcedureDefCallback = function(procCode,
   Blockly.Events.setGroup(true);
   definitionRoot.dispose();
   Blockly.Events.setGroup(false);
-
-  // TODO (#1354) Update this function when '_' is removed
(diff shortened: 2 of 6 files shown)


epicfaace:
Updates?

benjiwheeler:
@picklesrus Just wanted to check on this one, since it's linked to a change in scratch-gui. Do you think it would be good to review and merge this?

picklesrus:
@benjiwheeler I think this change is good. I lost it in the Scratch 3 launch period so it is old so it definitely needs to to be tested. 
It is technically a breaking change because it renames a method, but that method was not supposed to be used outside the blocks before so I'm okay with that. Of course GUI cheated and did use, so the GUI change would need to go in too. 
If we merge this, there's a good chance the GUI tests will break when the dependabot scratch-blocks version bump tries goes into GUI. Though our test coverage might not be great :-/

That was all a long way of saying that this and the GUI PR are probably fine, but they just need to be tested/shepherded through the release process.

kchadha:
Closing this PR since it's a breaking change and others (outside of Scratch might depend on the old function name). For future benefit, breaking changes like this should be made piecemeal where we make a new function that references the old function and add a deprecation warning.

The costs of merging this PR outweigh the benefits (a purely developer facing minor improvement) at this point.

PR #1834, opened 2018-12-19
Go to definition

### LINKED ISSUE #1136: ENA-189 Implement "go to procedure definition"
### Expected Behavior

The "define" option in the context menu on a custom command block should bring the procedure definition into view.

In Scratch 2.0 it just brings it into view--doesn't center it.  We may want to center and/or highlight it.
We have to decide on text for the context menu option.  In Scratch 2.0 it's 'define', which is not very clear.

epicfaace:
### Resolves

Fixes #1136, Implement "go to procedure definition"

### Proposed Changes

Add a "go to definition" option to the context menu for custom blocks.

image

### Reason for Changes

Now that `Blockly.WorkspaceSvg.prototype.centerOnBlock` has been added, we can implement the "go to definition" feature.

### DIFF
--- a/blocks_vertical/vertical_extensions.js
+++ b/blocks_vertical/vertical_extensions.js
@@ -202,13 +202,13 @@ Blockly.ScratchBlocks.VerticalExtensions.PROCEDURE_DEF_CONTEXTMENU = {
  */
 Blockly.ScratchBlocks.VerticalExtensions.PROCEDURE_CALL_CONTEXTMENU = {
   /**
-   * Add the "edit" option to the context menu.
-   * @todo Add "go to definition" option once implemented.
+   * Add the "edit" and "go to definition" options to the context menu.
    * @param {!Array.<!Object>} menuOptions List of menu options to edit.
    * @this Blockly.Block
    */
   customContextMenu: function(menuOptions) {
     menuOptions.push(Blockly.Procedures.makeEditOption(this));
+    menuOptions.push(Blockly.Procedures.makeShowDefinitionOption(this));
   }
 };
 
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -519,13 +519,15 @@ Blockly.Procedures.makeEditOption = function(block) {
 /**
  * Callback to show the procedure definition corresponding to a custom command
  * block.
- * TODO(#1136): Implement.
  * @param {!Blockly.Block} block The block that was right-clicked.
  * @private
  */
 Blockly.Procedures.showProcedureDefCallback_ = function(block) {
-  alert('TODO(#1136): implement showing procedure definition (procCode was "' +
-      block.procCode_ + '")');
+  var def = Blockly.Procedures.getDefineBlock(block.getProcCode(), block.workspace);
+  if (def) {
+    block.workspace.centerOnBlock(def.id);
+    def.select();
+  }
 };
 
 /**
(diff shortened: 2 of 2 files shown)


epicfaace:
@paulkaplan @picklesrus ?

picklesrus:
Sorry for the delay!

I linked this change into scratch-gui to check out it out.  It works quite well with a single workspace, but in the context Scratch (and not just a single workspace with blocks), we might need to make a few modifications.  Here are some things that we'll need to think about:

- What to do when the procedure definition is in another workspace (e.g. a sprite that isn't selected) 
- What to do when the procedure definition is long enough to take up the whole window.  It looks like currently scrolls to the middle of the stack. You don't really get much context when the define hat is off the top of the view.  The top of the define stack might be better.
- The option on blocks in toolbox doesn't do anything. We could
  -  not show it in the context menu on toolbox blocks
  - tell the toolbox version of the context menu to use the target workspace

I'll take a closer look at what 2.0 did in these situations and take this demo to the team for discussion next week.

towerofnix:
> What to do when the procedure definition is in another workspace (e.g. a sprite that isn't selected)

An interesting concept, but I'm not sure what you mean? If there is no "define" block corresponding to a procedure call block within the same sprite, that procedure call will do nothing. This is how it's been since custom blocks were introduced in 2.0 and it's still the case in 3.0. As such, it doesn't make sense to exactly have a blatantly "correct" behavior, since a custom block caller showing up in a sprite without a corresponding procedure definition is already in a sense an error. My suggestion is that it just doesn't do anything at all when you click define (I believe this is the behavior in 2.0, not sure though).

> The option on blocks in toolbox doesn't do anything.

**Most certainly** it should be changed to function correctly, i.e. scroll the target workspace to the position of that block just as though you'd clicked the option on a block dragged onto the workspace. I frequently am scrolled far away from a custom block as well as its callers; but I know which custom block I want to scroll to and modify. So finding it in the block palette and then selecting "define" from the context menu is exactly the behavior wanted here. (Otherwise, you would have to drag out the block, then select "define" - hence scrolling you away from the block you dragged out! And so it would be lying there, not doing anything besides functioning as a randomly-placed "shortcut" to the custom block definition. Frustrating for a variety of reasons, e.g. if you wanted to delete a custom block but now you can't because you dragged out a block to navigate to it.)

epicfaace:
@towerofnix @picklesrus The "go to definition" feature should work from a block in the toolbox now.

picklesrus:
Toolbox change works great!

After discussing with the team the main thing standing in the way of merging this is that, in this context, centerOnBlock doesn't lead to quite the right interaction. The hat block defining the function/procedure should be visible in the workspace when you click "Go to definition".  In a lot of cases it works nicely, but for a project like https://scratch.mit.edu/projects/222972802 where many of the procedures are longer than a screen, centerOnWorkspace to drops you in the middle of a stack.

If you want to change centerOnBlock, I'd be happy to merge this, but also understand that might be more work that you want to sign up for :)

towerofnix:
I know this PR isn't the one that adds the "go to definition" string, but perhaps it should be tweaked from "Go to definition" to "Go to Definition", to match the title-case of most buttons in the Scratch editor?

picklesrus:
I'm going to close this for now, but only because centerOnBlock doesn't have the right behavior.  Thank you for contributing!!

PR #1881, opened 2019-01-30
Manually calculate the bounding box of the flyout since we have all t…

picklesrus:
…he data. The browser does not reflow as a result of getBBox anymore.

### Proposed Changes

Make our own bounding box calculations based on the size of the blocks and buttons/labels in the flyout rather than asking the browser to do it for us.

### Reason for Changes

It makes things (like refreshing the toolbox) faster.

### Test Coverage

Compared the output of getBbox with the new function for toolboxes that start and end in blocks, buttons/labels and in rtl.

### DIFF
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -170,6 +170,50 @@ Blockly.VerticalFlyout.prototype.createDom = function(tagName) {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):
Code LGTM

PR #1883, opened 2019-01-31
Raise dragging layer z-index

chrisgarrity:
### Resolves

Resolving various z-index issues in gui and www.

### Proposed Changes

Raise the dragging layer z-index to match the gui.
### Reason for Changes

There are other ui elements (menus, alerts etc) that are above the toolbox. Anything being dragged should be above those. Gui uses 1000 (or greater) as the z-index when dragging costumes, sounds, etc. Blocks should match.

Looking at other z-index values, there didn't appear to be anything else between 50 and 1000. Raising it to 1000 matches the z-index for `blocklyDropDownDiv`. Should `blocklyDropDownDiv` get a small increment to maintain its position above the drag-layer? 
### Test Coverage

Doesn't change functionality so it should be covered by existing tests.

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -218,7 +218,7 @@ Blockly.Css.CONTENT = [
     'right: 0;',
     'bottom: 0;',
     'overflow: visible !important;',
-    'z-index: 50;', /* Display above the toolbox */
+    'z-index: 1000;', /* Display above the toolbox */
   '}',
 
   '.blocklyTooltipDiv {',


picklesrus:
Do we override blocks styles in gui or www? It kind of seems like this is specific to how we integrate scratch blocks into those and not scratch-blocks itself.  Thoughts?

chrisgarrity:
Good point - I had forgotten that we had a place to override blockly styles in the gui. Closing this PR in favor of https://github.com/LLK/scratch-gui/pull/4477

PR #1894, opened 2019-02-11
Fix flyout button

### LINKED ISSUE #1671: Dropping a block over a button in the blocks palette activates it
### Expected Behavior

The button is not activated, since it was not clicked.

### Actual Behavior

The button is activated.

### Steps to Reproduce

Drop a block over a button in the block palette, such as "Make a Variable".

Firefox 62.0b14, Debian (testing).

picklesrus:
…g in process.

This issue was caused by adding pointer-events:none on the drag surface 
element in https://github.com/LLK/scratch-gui/pull/2584

### Resolves

#1671 
### Proposed Changes

Bail out of the flyout button mouse up handler when a drag is in progress so that ending a drag doesn't cause a button click.


paulkaplan (approved):
LGTM

Issue #1855, opened 2019-01-04
Dragging a block sometimes offsets the whole workspace


towerofnix:
I can't reproduce this very reliably, but I noticed that sometimes the whole workspace gets offset while you're dragging a block. I've only ever seen this on scratch.mit.edu. I noticed that whenever I open a project, I can almost always repro the whole workspace appearing to be positioned above the backpack, with the flyout extending all the way to the bottom of the entire browser window, i.e. above the backpack.

### Operating System and Browser

Firefox 65.0b7, Linux (Debian).

I think this is Firefox-specific, since I can't repro it on Chromium.

BryceLTaylor:
@towerofnix can you describe what you mean by "offset?"  A screenshot or gif of the workspace would be helpful.  I can't reproduce, or if I am reproducing I'm not sure what you mean.

towerofnix:
@BryceLTaylor I couldn't repro the "offset" behavior per se - it was likely an instance of overflow issue of and related to LLK/scratch-gui#953. Here's a repro video: https://streamable.com/1sqke. I've already made a PR - #1895 - which almost certainly resolves this (it resolves 953).

PR #1895, opened 2019-02-11
Wrap drag surface so overflow is cut off

### LINKED ISSUE #953: Corners of editor area don't remain rounded when dragging blocks
In an extension of #318, when dragging blocks the corners of the editor area become square and return to rounded when the mouse is released.

towerofnix:
### Resolves

I don't think it's filed, but this fixes an issue with blocks that is the same in principle as LLK/scratch-gui#1791; particularly, fixes the scrollbar showing up when you drag a block past the edge of the screen. Also fixes LLK/scratch-gui#953 (since toggling overflow is no longer necessary).

Also is support for upcoming LLK/scratch-gui#4530.

While I can't be certain, I'm hoping this fixes or helps towards #1855 as well.

### Proposed Changes

Wraps the drag surface in a wrapper element that covers the whole screen and is marked as `overflow: hidden`. (This is the same technique as in LLK/scratch-gui#4527: If an element is outside of the screen, it will be cut off, and will not cause a scrollbar to show up.)

Since the SVG which follows the mouse as it is dragged is now positioned relative to this fixed element rather than the workspace, we use margin CSS properties to offset the element to the original position.

Note we don't need to place the element on the document body (as I originally thought we might, mentioned in https://github.com/LLK/scratch-gui/issues/1792#issuecomment-452839712); the `position: fixed` CSS makes the wrapper already effectively ignore whatever hierarchy it's in.

### Reason for Changes

To improve block dragging behavior in general.

### Test Coverage

Tested manually in a variety of cases:

* Blocks follow the mouse as it is dragged properly (and are dropped in the place they are visually dragged over, just as expected).
  * Even if the workspace is scaled (zoom buttons).
  * Even if the entire page is zoomed (browser zoom).
* Z-indexing is unchanged; dragged blocks are still positioned over the flyout.

Also tested inside of scratch-gui, where everything performs as expected.

### DIFF
--- a/core/block_drag_surface.js
+++ b/core/block_drag_surface.js
@@ -112,6 +112,8 @@ Blockly.BlockDragSurfaceSvg.prototype.createDom = function() {
   if (this.SVG_) {
     return;  // Already created.
   }
+  this.wrapper_ = goog.dom.createDom('div', 'blocklyDragSurfaceWrapper');
+  document.body.appendChild(this.wrapper_);
   this.SVG_ = Blockly.utils.createSvgElement('svg',
       {
         'xmlns': Blockly.SVG_NS,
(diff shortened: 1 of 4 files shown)


towerofnix:
~~Note ­— this shouldn't be merged yet, comment dragging apparently needs to be tweaked too.~~
Fixed: https://github.com/LLK/scratch-blocks/pull/1895/commits/38c8eb4272224b6e93e0a86f0e79ae2e155515e5

thisandagain:
@paulkaplan Bump

core/block_drag_surface.js
```diff
@@ -128,6 +130,16 @@ Blockly.BlockDragSurfaceSvg.prototype.createDom = function() {
...
 };
 
+/**
+ * Set the distance to offset the dragged SVG (in pixels).
+ * @param {number} left - Left offset.
+ * @param {number} top - Y offset.
+ */
+Blockly.BlockDragSurfaceSvg.prototype.setOffset = function(left, top) {
+  this.SVG_.style.marginLeft = left + 'px';
+  this.SVG_.style.marginTop = top + 'px';
```
adroitwhiz, line 140: It may be [more performant](https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/) to use `transform` if that's possible here.

PR #1898, opened 2019-02-13
make sure to have version of l10n that includes fa and ckb

chrisgarrity:
Update minimum scratch-l10n version to make sure Persian and Central Kurdish are included.


chrisgarrity:
redundant.

Issue #1856, opened 2019-01-05
Editing a procedure should not scroll flyout


towerofnix:
### Expected Behavior

Editing a procedure spec (the labels/inputs) should never scroll the flyout, since editing a procedure is usually done to change how existing calls to the procedure look, not to create new calls.

### Actual Behavior

Editing a procedure spec does scroll the flyout to the "My Blocks" section.

### Steps to Reproduce

Right-click an existing custom procedure block and press Cancel or OK.

### Operating System and Browser

Firefox 65.0b7, Linux (Debian).

kchadha:
Closing this issue as `won't fix` since it's a very low severity bug with a much higher-risk solution.

PR #1901, opened 2019-02-19
Tell external callback if procedure definition dialog is for editing

towerofnix:
### Resolves

Towards #1856. Should be merged before LLK/scratch-gui#4572.

### Proposed Changes

Adds a new `isEdit` argument (between `mutator` and `callback`) to `Blockly.Procedures.externalProcedureDefCallback` that tells whether or not the call came from editing an existing procedure.

### Reason for Changes

Support to fix the above bug! This could also let the "Make a Block" dialog display more relevant text, like "Edit Block".

### Test Coverage

Tested manually (as part of LLK/scratch-gui#4572).


kchadha:
Closing this PR, and the corresponding issue since the risk of this change far outweighs the benefits of fixing this very low severity bug.

PR #1907, opened 2019-03-02
"Relabel" context menu options

### LINKED ISSUE #1383: Operators should be able to be switched via right-click menu
### Expected Behavior
The `+`, `-`, `*`, etc. blocks in Scratch 2.0 provide a context menu to switch between operators.

### Actual Behavior
This has not yet been implemented.

towerofnix:
### Resolves

Resolves #1383. See also LLK/scratch-gui#1740, which discusses lots of potential relabel choices and should be reviewed before merging this, since they aren't all part of this PR and considering adding more of them is definitely welcome.

### Proposed Changes

Implements "relabeling" specific reporter blocks:

Demo of relabeling blocks - a context menu showing the options `+`, `*` (highlighted), and `-`, and the result of picking `*`, a `*` block showing beside it with the same inputs as the first block.

When you right-click specific blocks, the context menu will include options like in the screenshot above; when you pick one, it changes the block to a different opcode/type. (This actually happens by duplicating the block with a new opcode, then instantly replacing the original block with this new one.)

This PR specifically adds these groups of blocks that can be relabeled to each other:

* `+`, `-`, `*`, `/` (from 2.0)
* `<`, `=`, `>` (also from 2.0)
* `and`, `or` (this one's new, but a pretty popular suggestion on the forums)

It's easily extensible to have other groups of blocks - see the changes in `blocks_vertical/operators.js`. The only hard requirement is that input/field names be identical between the blocks.

### Reason for Changes

Relabeling blocks is a convenient way of swapping out one block for a similar one, which happens quite frequently when building large math-y scripts.

It's also a feature brought back from Scratch 2.0.

### Test Coverage

No new unit tests; tested manually.


towerofnix:
Travis is failing but I'm not sure why - I haven't worked with the translation system much so I don't really understand what the error means.

joker314:
@towerofnix It isn't your fault. Adding translations requires a Scratch Team member.

[(source)](https://github.com/LLK/scratch-blocks/wiki/Localization)
> If you submit a PR that changes `messages.js` and you are not a Scratch Team member, the translation tests should fail. Make sure you check the 'Allow edits from maintainers' option on your PR, and then tag a Scratch Team member to push a commit to your branch that includes translations. All the tests other than `test:messages` should run correctly before you submit the PR.

apple502j:
Resolves #1383  and LLK/scratch-gui#1740

towerofnix:
@apple502j Thanks for the links, updated the PR description.

msg/js/en.js
```diff
@@ -152,6 +152,8 @@ Blockly.Msg["OPERATORS_LT"] = "%1 < %2";
 Blockly.Msg["OPERATORS_EQUALS"] = "%1 = %2";
 Blockly.Msg["OPERATORS_AND"] = "%1 and %2";
 Blockly.Msg["OPERATORS_OR"] = "%1 or %2";
+Blockly.Msg["OPERATORS_RELABEL_AND"] = "And";
```
chrisgarrity: these should probably be lower case to be consistent with the blocks

chrisgarrity:
@towerofnix making any change to `msg/messages.js` requires pushing the new strings to our translation service (transifex) and pulling default translations back into scratch-blocks for all the languages supported. Since this requires developer access to transifex, it can only be done by a Scratch Team member. @picklesrus I'm happy to pair on syncing the translations. You'll need to push the updated scratch_msg file to @towerofnix 's branch after synching for the build to pass.

msg/js/en.js
towerofnix: I was debating over which was right here...thanks for the input! Resolved here: 67d6b127bd00b5150a43e80c658547f23ccb0144 Also thanks for the details on making translation changes.

picklesrus:
Hi @towerofnix.  Thanks for this PR!

Does the vm actually realize that the block has changed?  Do we get end up getting events through the call to connect in the callback?

I'm worried about this approach for a couple of reasons. Mutations (see the stop block as an example) are the blocks' way of communicating that a block changes shapes.  Adding this api seems like stepping around that. If you could use a mutation, I think that means that there would be a single operator block that takes different shapes (meaning +, -, *, etc.). We've already chosen 3 separate ones so combining into 1 means thinking about interactions with the VM, saving, and loading which we haven't had the chance to do yet. I'd also want to think pretty hard before we decided to add something to the json. We generally don't add things to it lightly, and this seems pretty specific to this one block and not something more generic. It could just be it needs a less specific name, but I still come back to mutations being the way for blocks to change.

Since I'm not ready to add this kind of api to the blocks, I'm going to close this PR for now.

towerofnix:
@picklesrus Thanks for the review. I've got some thoughts though.

> Does the vm actually realize that the block has changed? Do we get end up getting events through the call to connect in the callback?

Yes, we do. I've tested this in scratch-gui and it works exactly as anyone would expect.

>We generally don't add things to it lightly, and this seems pretty specific to this one block and not something more generic.

I understand that modifying the JSON area is not something done lightly but I ask *why* - this is a remarkably straightforward and clean implementation after all.

I certainly investigated mutations but couldn't figure out a nice way to use them. It sounds like mutations are usually used for making changes to an existing block - not for swapping one out with a new one. Accordingly, I created new code for this new behavior; and I made use of existing APIs to do so. (Again, this is literally no different from inserting a new block in place of the old one, then deleting the old one; it simply automates the action.)

Combining blocks such as +, -, * and /, and <, =, and > all into one block and writing a mutation function is blatantly overkill. Implementing general behavior which is easy to apply to any existing and new blocks, i.e. what has been done here, is much more practical and reliable, frankly.

Thoughts?

picklesrus:
Good questions!

I can see an argument for adding a generic api that replaces/swaps one block with another. I'm not sure I'd want that to be involved in the block definition though.  Also, this doesn't do that in a generic way. For example, what if you wanted to swap a statement block with another statement block? I think this only manages connections correctly for a reporter block. Also, as it stands now, I don't think events are being managed/grouped properly.  The undo/redo stack does weird things when after I switch from one operator to another.

It is possible that adding something like this in the future is the right thing to do, but until we have time to prioritize thinking through the implications of that with respect to upcoming infrastructural changes and ruling out using mutations as an option, I don't want to add it now.

PR #1928, opened 2019-04-21
Use opt_secondValue in addBlock as expected

### LINKED ISSUE #1927: Text of the second value is ignored in DataCategory.addBlock
### Expected Behavior

Screenshot from 2019-04-21 13-59-00

### Actual Behavior
See the last parameter of the block: `with 1`.
Screenshot from 2019-04-21 13-53-50

I don't think this is the desired behaviour. Take a look at this code:
https://github.com/LLK/scratch-blocks/blob/937e4793ef50dd066ce9d18ab6260f1202c1ab9e/core/data_category.js#L285
The value given is `Blockly.Msg.DEFAULT_LIST_ITEM`, which is `thing`. But the following code does not use it, rather uses the other parameter's value, which is `1`: https://github.com/LLK/scratch-blocks/blob/937e4793ef50dd066ce9d18ab6260f1202c1ab9e/core/data_category.js#L430

tansly:
### Resolves

Closes #1927.

### Proposed Changes

Use the parameter `opt_secondValue` instead of `opt_value` where appropriate.

### Reason for Changes

`opt_secondValue[2]` was not used at all in the code, which lead to behaviour seen in #1927.

### Test Coverage

No tests were added, no existing tests were broken.

### DIFF
--- a/core/data_category.js
+++ b/core/data_category.js
@@ -427,7 +427,7 @@ Blockly.DataCategory.addBlock = function(xmlList, variable, blockType,
     }
     if (opt_secondValue) {
       secondValueField = Blockly.DataCategory.createValue(opt_secondValue[0],
-          opt_secondValue[1], opt_value[2]);
+          opt_secondValue[1], opt_secondValue[2]);
     }
 
     var gap = 8;


picklesrus:
Looks good! Thank you!

PR #1888, opened 2019-02-04
Sort context menu variable list

towerofnix:
### Resolves

Fixes #1879.

### Proposed Changes

Sorts the variable list in the variable context menu:

A halfway-scrolled context menu with many alphabetically sorted variables in it.

Sorts list context menus, too:

The same deal, but in the context menu of a list reporter.

### Reason for Changes

To make the context menu of variable/list menus more useful, and more consistent with other lists of variables in the UI.

### Test Coverage

No new unit tests. Tested manually.

Issue #1879, opened 2019-01-27
Variables in context menu aren't sorted


towerofnix:
### Expected Behavior

Variables should be sorted in every menu they appear in.

### Actual Behavior

In the context menus of variables (as well as lists), variables are not sorted.

![A long context menu containing lots of arbitrarily arranged variable names.](https://user-images.githubusercontent.com/9948030/51803790-52d4dd00-222f-11e9-9f1d-652fba0e4df8.png)

### Steps to Reproduce

Right-click a variable or list reporter block. Note the variables are not sorted alphabetically (they're sorted by creation date, which is not the standard 3.0 behavior).

### Operating System and Browser

Firefox 66.0b2, Linux (Debian).

PR #1929, opened 2019-04-23
Sort list variables in context menu

towerofnix:
### Resolves

No issue, but it'd be corresponding to #1879.

### Proposed Changes

This is the exact same PR as #1888 from a couple months ago but for lists now. It sorts the variable (list) options in the list right-click menu.

### Reason for Changes

Consistency with other menus in Scratch 3.0.

### Test Coverage

Tested manually (created lists "a, z, b", dragged out an "a" block, then right-clicked it - on develop it shows "z, b", on this PR it shows "b, z" correctly).

### DIFF
--- a/blocks_vertical/data.js
+++ b/blocks_vertical/data.js
@@ -527,6 +527,7 @@ Blockly.Constants.Data.CUSTOM_CONTEXT_MENU_GET_VARIABLE_MIXIN = {
       for (var i = 0; i < variablesList.length; i++) {
         var varName = variablesList[i].name;
         if (varName == currentVarName) continue;
+
         var option = {enabled: true};
         option.text = varName;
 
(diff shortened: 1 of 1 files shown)


blocks_vertical/data.js
```diff
@@ -527,6 +527,7 @@ Blockly.Constants.Data.CUSTOM_CONTEXT_MENU_GET_VARIABLE_MIXIN = {
       for (var i = 0; i < variablesList.length; i++) {
         var varName = variablesList[i].name;
         if (varName == currentVarName) continue;
+
```
kchadha, line 530: Nit: stray newline

```suggestion
```

kchadha (commented):
@towerofnix, there's just a stray newline, but otherwise this looks good. Sorry this took so long to review!

blocks_vertical/data.js
towerofnix: This is so that it's consistent with the `CUSTOM_CONTEXT_MENU_GET_VARIABLE_MIXIN` function: https://github.com/towerofnix/scratch-blocks/blob/564d3ca8c22aa2c0c6f240ab3369b4093b2a1934/blocks_vertical/data.js#L527-L532

I can commit your suggestion if you want, but I figure code consistency is nice and it'll likely be a while (extensionification) before this code is touched again :P

No worries about the wait time! It was set in the May milestone after all :P

kchadha: Ah, I assume you mean so that it's consistent with the `..._LIST_MIXIN` function.

That's a good point, we can keep the newline. Thanks!

kchadha (approved):
LGTM, thank you!

Issue #1175, opened 2017-10-24
Blockly.Procedures.isNameUsed returning wrong values


smaameri:
The Blockly.Procedures.isNameUsed function is returing the opposite of what it says it should. 

As per the definition it says:

`//  @return {boolean} True if the name is used, otherwise return false`

however the code is returning opposite. Compared with the Blockly code, the return true/false locations are also switched. I.e the function is returning true when it should return false, and false when it should return true. Any reasons for this?

```
/**
 * Return if the given name is already a procedure name.
 * @param {string} name The questionable name.
 * @param {!Blockly.Workspace} workspace The workspace to scan for collisions.
 * @param {Blockly.Block=} opt_exclude Optional block to exclude from
 *     comparisons (one doesn't want to collide with oneself).
 * @return {boolean} True if the name is used, otherwise return false.
 */

Blockly.Procedures.isNameUsed = function(name, workspace, opt_exclude) {
  var blocks = workspace.getAllBlocks();
  // Iterate through every block and check the name.
  for (var i = 0; i < blocks.length; i++) {
    if (blocks[i] == opt_exclude) {
      continue;
    }
    if (blocks[i].getProcedureDef) {
      var procName = blocks[i].getProcedureDef();
      if (Blockly.Names.equals(procName[0], name)) {
        return false;
      }
    }
  }
  return true;
};

```

thisandagain:
@rachel-fenichel @paulkaplan Any thoughts on this?

rachel-fenichel:
Yup, that looks wrong!  Fixed in blockly [here](https://github.com/google/blockly/commit/234c53157f676978ab9de232ed28305720443003), with a commit comment of "Fix Blockly.Procedures.isNameUsed return values.
Add tests for Procedures.isNameUsed() so this bug never happens again."

However, fixing that won't fix `isNameUsed` right now--it's also calling `getProcedureDef` to check whether something is a procedure.  Since that's not defined on any blocks, it'll just always return true.

I'm working on implementing procedures right now, so I'll add this to my list.

rachel-fenichel:
However, this may be an argument for calling [`getProcCode`](https://github.com/LLK/scratch-blocks/blob/develop/blocks_vertical/procedures.js#L157) `getProcedureDef` instead, to keep compatibility.  Something to think about.

rachel-fenichel:
I'm going to leave this open and assigned to me, because it's a real issue with the code.

cwillisf:
[...]

PR #1930, opened 2019-04-24
Return correct values for isNameUsed function

aaimio:
### Resolves

No related Github issue, but it fixes `Blockly.Procedures.isNamedUsed` returning an incorrect value. Fix was already added to Blockly by @marisaleung 
- https://github.com/google/blockly/commit/234c53157f676978ab9de232ed28305720443003#diff-14d8bd917c33ff9fb0dfee355a789189

### Proposed Changes

Changes the value returned when checking if a procedure name already exists. Currently the function returns `false` when `Blockly.Names.equals(procName[0], name)`, however, it should return `true`. This proposed change updates those values.

### Reason for Changes

If left unchanged, the `isNameUsed` function creates an infinite loop if a procedure name already exists.

### DIFF
--- a/core/procedures.js
+++ b/core/procedures.js
@@ -181,11 +181,11 @@ Blockly.Procedures.isNameUsed = function(name, workspace, opt_exclude) {
     if (blocks[i].getProcedureDef) {
       var procName = blocks[i].getProcedureDef();
       if (Blockly.Names.equals(procName[0], name)) {
-        return false;
+        return true;
       }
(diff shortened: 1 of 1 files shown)


picklesrus (commented):
Whoops, That last comment was meant for a different PR. Sorry!

adroitwhiz:
Resolves https://github.com/LLK/scratch-blocks/issues/1175

PR #1937, opened 2019-05-06
Cache Field text widths with 2D Canvas's measureText

mzgoddard:
### Resolves

Field rendering performance.

### Proposed Changes

Related https://github.com/LLK/scratch-gui/pull/4669.

- Use CanvasRendering2DContext.measureText to determine text widths.
- Cache text widths by 2 dimensions className and textContext.

### Reason for Changes

> - Use CanvasRendering2DContext.measureText to determine text widths.

The first render of some group of blocks needs to determine the text widths for fields in the blocks. Currently getCachedWidth queries the produced text node layout to get the width. Requesting data from the dom layout in this scenario takes a lot of time. Instead of querying the text width for each field's text we can query the font for each class name.

CanvasRendering2DContext has a helper method for measuring text. Setting the font member of a CanvasRendering2DContext and calling measureText, we can get the width of some text without fetching that from the dom.

> - Cache text widths by 2 dimensions className and textContext.

measureText still has some cost as its walks the given text and looks up the related glyph in the font. So caching the produced widths is still very helpful. We can improve the cache lookup performance a little by using a nested lookup.

### DIFF
--- a/core/field.js
+++ b/core/field.js
@@ -117,6 +117,13 @@ Blockly.Field.fromJson = function(options) {
 Blockly.Field.cacheWidths_ = null;
 
 /**
+ * Temporary cache of text width measuring tools.
+ * @type {Object}
+ * @private
+ */
+Blockly.Field.cacheTextTools_ = null;
+
+/**
  * Number of current references to cache.
  * @type {number}
  * @private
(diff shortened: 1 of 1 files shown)


core/field.js
```diff
@@ -492,42 +499,172 @@ Blockly.Field.prototype.updateWidth = function() {
...
+    // These three are commonly used so lets initialize them together. This will
+    // reduce the cost of DOM style and layout while determining the font.
+    classNames = [
+      'blocklyFlyoutLabelText',
+      'blocklyText',
+      'blocklyText blocklyDropdownText'
+    ];
+  }
+
+  var root = document.body;
```
picklesrus, line 563: Can you explain what this root is - I'm confused :)
  I thought getRootNode would return the document, though when I look at this in the debugger, it is returning the <g> tag. Then later you're then adding another <g> and appending to that.  Is that right?

rachel-fenichel:
I am very interested in this change :D

core/field.js
mzgoddard: This version needs to determine the blockly fonts from CSS. (We could hard code them in JS.) So we need nodes in the DOM to fetch that style. Our common case is to get the fonts for all the blockly text node class names. This root is where we attach those nodes so we can get the font info for each class in one pass.

`document.body` is from a version of this I wrote where the text element may not have been attached to the dom yet. We can change this and require it be an element in the DOM. Which, the element is, in this version.

Adding a node, getting the info, removing it, adding the second, getting the info, removing, etc. will incur a restyle event and performance cost. Reusing the same node would also incur this cost. In the best case, we add all the nodes to the DOM, get the font info for each node, then remove all the nodes from the DOM.

The parent is here to conveniently let us detach the nodes after we get the font info for all of the nodes. We could add and remove the nodes to and from the root without the parent alternatively. As long as we add them all, get all their info, and remove them all.

core/field.js
```diff
@@ -492,42 +499,172 @@ Blockly.Field.prototype.updateWidth = function() {
...
-      return width;
-    }
-  }
-
-  // Attempt to compute fetch the width of the SVG text element.
-  try {
-    if (goog.userAgent.IE || goog.userAgent.EDGE) {
-      width = textElement.getBBox().width;
+    if (!Blockly.Field.cacheWidths_[className]) {
+      Blockly.Field.cacheWidths_[className] = {};
```
picklesrus, line 651: is it likely to hit this case? I thought this would have been initialized when initTextTools was called, but maybe I am misunderstanding the order in which things happen?

picklesrus (commented):
Sorry for the very long delay!  
This is looking good - I just want to clarify one thing.  See question below.

core/field.js
mzgoddard: Every run will execute this line a few times.

I can see the misunderstanding here. This PR introduces a second separate cache. And they both start with `cache` in their names.

This file already has a `Blockly.Field.cacheWidths_` cache. And the PR adds `Blockly.Field.cacheTextTools_`. `initTextTools` only works with `cacheTextTools_` and does not manage or interact with `cacheWidths_`. And `getCachedWidth` only works with `cacheWidths_` and does not manage or interact with `Blockly.Field.cacheTextTools_`. They are wholly independent caches. We can turn on one cache and leave the other off and they'll still function as expected.

We could move these caches out of Field and into helper classes that Field's class could make instances of. That could help distinguish the caches and functionality for text measuring. I'm not sure we need that. Everything should go through Field to get these values, so any new classes and modules from this would be ones Field depends on anyway.

PR #1942, opened 2019-05-23
Handle promise rejection for sound.play() in preload

mfrtrifork:
### Resolves

audio play() called before user interaction [Blockly fix](https://github.com/google/blockly/pull/2162)

### Proposed Changes

Handle promise rejection for sound.play() in preload. 

### Reason for Changes

This removes the error, "play() failed because the user didn't interact with the document first."

### Test Coverage

None

### DIFF
--- a/core/workspace_audio.js
+++ b/core/workspace_audio.js
@@ -112,7 +112,7 @@ Blockly.WorkspaceAudio.prototype.preload = function() {
   for (var name in this.SOUNDS_) {
     var sound = this.SOUNDS_[name];
     sound.volume = 0.01;
-    sound.play();
+    sound.play().catch(function() {});
     sound.pause();
     // iOS can only process one sound at a time.  Trying to load more than one
     // corrupts the earlier ones.  Just load one and leave the others uncached.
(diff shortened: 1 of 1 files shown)


epicfaace:
Be careful, I tried the same solution in blockly (https://github.com/google/blockly/pull/2162) but it ended up breaking something else.

mfrtrifork:
@epicfaace thanks for the warning. Is there a solution which is not breaking something else? - I see that your commit is still present in master on blockly.

rachel-fenichel:
The problem for Blockly was that we're still supporting ES5.  I'm pretty sure Scratch doesn't so you're probably fine.  I suggest finding the browser compatibility list for Scratch and directly checking against those browsers.

mfrtrifork:
According to: https://en.scratch-wiki.info/wiki/Compatibility_of_Versions_of_Scratch scratch supports:
Chrome 63+ or Mobile Chrome 63+
Edge 15+
Firefox 57+
Safari 11+ or Mobile Safari 11+

According to: https://www.w3schools.com/js/js_versions.asp ES6 is supported in:
Chrome | 58 | Apr 2017
Firefox | 54 | Jun 2017
Edge | 14 | Aug 2016
Safari | 10 | Sep 2016

Should be good for ES6 right @rachel-fenichel ?

An extended compatibility table can be found here: https://kangax.github.io/compat-table/es6/

paulkaplan (changes requested):
From that linked article, it looks like `.play` returning a promise is not supported in Edge, so this will throw an exception on Edge. If this is the case, you may want to check if a `then` function is present on the return value of `play` before using it, falling back to calling pause right away if it doesn't return a promise.

mathiassoeholm:
Well spotted @paulkaplan :-)

My colleague @mfrtrifork is on vacation, so i made a new PR with the request changes: https://github.com/LLK/scratch-blocks/pull/1959

PR #1944, opened 2019-05-24
Fix style matching in Firefox 67 (port from Blockly)

### LINKED ISSUE #1915: Blocks are not dragged from flyout at accurate position
### Expected Behavior

Dragging a block should always attach it to the mouse according to the position where it was grabbed.

### Actual Behavior

This doesn't happen accurately when the block is dragged from the flyout.

### Steps to Reproduce

Drag a block from the flyout, from various different positions. Note it always starts moving from the same position (slightly to the left of where the block is in the flyout).

Firefox 67.0b2.

### LINKED ISSUE #1943: Dragging from flyout is offset in Firefox 67
See https://github.com/google/blockly/issues/2482

Firefox 67 was released yesterday and we got a very fast bug report on this.

This is a blockly bug, but was only exposed because of a change in how Firefox's style property is formatted.  We cherry-picked the fix (https://github.com/google/blockly/pull/2485) into master.

towerofnix:
### Resolves

Fixes #1915 and fixes #1943.

### Proposed Changes

This PR is ported from google/blockly#2485. The description from that PR:

>Chrome returns ‘transform: translate(107px, 0px);’ whereas Firefox now
>returns ‘transform: translate(107px);’ if the y value is 0.  This is
>consistent with existing behaviour in the translate SVG attribute.
>
>The comment in our code specifically states:
>`// Accounts for same exceptions as XY_REGEX_`
>Yet that was not true at all.
>
>This PR makes the y argument optional (as falsely described in the
>comment).  It also merges the 2D and 3D regeps together to simplify the
>code.

### Reason for Changes

Fixes blocks being dragged from the flyout getting offset in Firefox 67+ (now the current release).

### Test Coverage

Tested manually on Firefox 68.0b3 and Chromium 68.0.3440.75. Works correctly in both browsers.

/cc @rachel-fenichel @NeilFraser

### DIFF
--- a/core/utils.js
+++ b/core/utils.js
@@ -182,11 +182,7 @@ Blockly.utils.getRelativeXY = function(element) {
   // Then check for style = transform: translate(...) or translate3d(...)
   var style = element.getAttribute('style');
   if (style && style.indexOf('translate') > -1) {
-    var styleComponents = style.match(Blockly.utils.getRelativeXY.XY_2D_REGEX_);
-    // Try transform3d if 2d transform wasn't there.
-    if (!styleComponents) {
-      styleComponents = style.match(Blockly.utils.getRelativeXY.XY_3D_REGEX_);
-    }
+    var styleComponents = style.match(Blockly.utils.getRelativeXY.XY_STYLE_REGEX_);
     if (styleComponents) {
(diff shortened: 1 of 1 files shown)


picklesrus (approved):
LGTM!

PR #1948, opened 2019-06-10
Make checkboxes and extension status buttons easier to tap on touch

paulkaplan:
1. Add a transparent clickable rect around extension status button that extends farther to make the button more easily clicked.
2. Do the same for checkboxes.
3. Increase the visual size of the checkboxes as well.
4. Changed the vertical playground initial scale to match what is used on production scratch to make it easier to compare.

New on the left, old on the right
image
image

/cc @carljbowman

### DIFF
--- a/tests/vertical_playground.html
+++ b/tests/vertical_playground.html
@@ -84,7 +84,7 @@
           zoom: {
             controls: true,
             wheel: true,
-            startScale: 0.75,
+            startScale: 0.675,
             maxScale: 4,
             minScale: 0.25,
             scaleSpeed: 1.1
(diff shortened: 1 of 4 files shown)


carljbowman:
@paulkaplan - Thanks for taking a look at this! Looks like a great first step in some of the touch issues.

Let me know how you want to follow up on some of the other spacing and element sizing we have been discussing. Happy to grab a hour to pair and iron out the deets.

tests/vertical_playground.html
```diff
@@ -84,7 +84,7 @@
           zoom: {
             controls: true,
             wheel: true,
-            startScale: 0.75,
+            startScale: 0.675,
```
chrisgarrity, line 87: Thanks for making this consistent!

chrisgarrity (approved):
Looks good.

PR #1949, opened 2019-06-11
ENA-189 Add "go to definition" menu option so it's easy to find your custom blocks/procedures

### LINKED ISSUE #1136: ENA-189 Implement "go to procedure definition"

towerofnix:
### Resolves

Resolves #1136.

### Proposed Changes

This PR is based off of @epicfaace's changes in #1834. It adds a "go to definition" menu option to custom procedure callers, which now makes the workspace viewport center on the prototype (top) of the custom block.

Screenshot summarizing the new feature: the context menu with the new option highlighted, opened on the caller block in the palette, and the result of clicking it, the viewport centered on the prototype block of the custom procedure.

### Reason for Changes

The "go to definition" feature is much requested, and was part of Scratch 2.0 (where the block was called "define").

This particular PR contains the additional change of centering on the block's prototype instead of the entire script (its "definition"). However, I'm not 110% sure this is the best behavior. I feel it would be better if, when the entire script *can* be fit inside the workspace, the full script would be centered upon. When it *can't* fit, the top of the workspace (with some padding) would be aligned with the top of the custom block (the prototype), so as to show as much of the script as possible. (We'd want to take care to do the same check for the X-axis, so that custom blocks with very wide blocks don't have their start-end cut off. Worth being sure of LTR/RTL support here!)

### Test Coverage

Only manual testing on both Firefox and Chromium, on project ID `316338651` (unshared) and [Bubble Scratch v0.9](https://scratch.mit.edu/projects/222972802).

(/cc @picklesrus since you reviewed the original PR by @epicfaace!)

### DIFF
--- a/blocks_vertical/vertical_extensions.js
+++ b/blocks_vertical/vertical_extensions.js
@@ -202,13 +202,13 @@ Blockly.ScratchBlocks.VerticalExtensions.PROCEDURE_DEF_CONTEXTMENU = {
  */
 Blockly.ScratchBlocks.VerticalExtensions.PROCEDURE_CALL_CONTEXTMENU = {
   /**
-   * Add the "edit" option to the context menu.
-   * @todo Add "go to definition" option once implemented.
+   * Add the "edit" and "go to definition" options to the context menu.
    * @param {!Array.<!Object>} menuOptions List of menu options to edit.
    * @this Blockly.Block
    */
   customContextMenu: function(menuOptions) {
     menuOptions.push(Blockly.Procedures.makeEditOption(this));
+    menuOptions.push(Blockly.Procedures.makeShowDefinitionOption(this));
   }
 };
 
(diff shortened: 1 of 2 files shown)


rebca-online:
Why hasn't this feature been merged? If you don't want it, can you at least give a reason?

benjiwheeler:
@towerofnix @actually-reb Sorry for the delay, and thanks for your persistence. We're kicking off several internal projects this month, and we've been occupied with that, but hopefully we'll get to it soon. Thanks for making the case for it!

benjiwheeler:
I gave severity and impact labels to this. Though it's not critical, it would be good to triage. It would improve user ergonomics, and it's a small change -- see @towerofnix's comment above

PR #1950, opened 2019-06-12
Allow selecting the last two piano keys

aeons:
### Proposed Changes

Currently, the last two piano keys in the `note_field` type are not selectable. This PR changes the max note for the field validation, so that you can select the last two piano keys.

### Reason for Changes

Either the last two blocks should be selectable or the last octave should be removed (IMO :))

### DIFF
--- a/core/field_note.js
+++ b/core/field_note.js
@@ -309,7 +309,7 @@ Blockly.FieldNote.KEY_INFO = [
  * @type {number}
  * @const
  */
-Blockly.FieldNote.MAX_NOTE = 130;
+Blockly.FieldNote.MAX_NOTE = 132;
 
 /**
  * The fraction of the distance to the target location to move the piano at each


ericrosenbaum:
Hi @aeons - I agree that the behavior of the note picker in the top octave is a bit imperfect (i.e. you can only click to select and play up to note 130, and if you click 131 or 132, it plays and selects 130). This is the best design we could come up with to meet our requirements. The maximum note value for the Scratch "play note" block is 130, so we don't want to let people select notes above that. Any value you input above 130 is clamped to 130 in the VM, so constraining the input to 130 keeps it consistent. We cannot simply change the maximum value in the VM because in general we make sure that the behavior of Scratch 3.0 is consistent with the behavior of earlier versions, and the maximum note has always been constrained to 130, so if we changed it, it could change the behavior of existing projects. 

So that's why we can't, IMO, accept this PR, or your other suggestion that we remove the top octave. Forks that want different behavior are of course welcome to make their own changes and improvements. If you'd like to suggest a design change for Scratch 3.0 you can file an issue for our design team to discuss. Thanks!

aeons:
Alright, that makes sense. Our implementation stops at 108 which is on an octave boundary.

ericrosenbaum:
hm, it occurred to me double-check and it looks like I was incorrect- the max note in both scratch 1 and [scratch 2](https://github.com/LLK/scratch-flash/blob/646523e6846ad0dd993213a38b46fe2ea511d026/src/sound/NotePlayer.as#L84) was 127 (to match the MIDI spec I think). So, looks like I introduced a subtle bug in 3.0 of exactly the kind I was describing above. Darn. Unfortunately, 127 is a G, which would make the UX problem of broken-note-selector-notes even worse... I'm not sure what to do at this point. :/ Probably just leave it as is. Sorry for the confusion.

PR #1952, opened 2019-06-19
Update blocks translations

chrisgarrity:
Update blocks translations in preparation for a Desktop release.


chrisgarrity:
Missed the window for updates due to other problems building scratch-blocks. These will get updated at the end of the month.

PR #1956, opened 2019-06-25
Fix broken image for numpad delete button

### LINKED ISSUE #1955: On Android tablet, number input dropdown's delete button has broken image
### Expected Behavior

On touch devices, the number input dropdown should have a delete button with a delete icon on it.

### Actual Behavior

On Chrome on an Android tablet, the delete button on the number input dropdown has a broken image.

paulkaplan:
It turns out we are creating a data-uri SVG for the delete button, but are using string interpolation for the color. If you use hex, this breaks the image on Chrome (but not Safari). Using a standard css color keyword (alternatively using RGB) fix the issue.

### Resolves

Fixes https://github.com/LLK/scratch-blocks/issues/1955

This can be tested by force-enabling the numpad and trying number inputs in the vertical playground: in field_number.js#168, set showNumPad=true;

### DIFF
--- a/core/colours.js
+++ b/core/colours.js
@@ -108,7 +108,7 @@ Blockly.Colours = {
   "numPadBackground": "#547AB2",
   "numPadBorder": "#435F91",
   "numPadActiveBackground": "#435F91",
-  "numPadText": "#FFFFFF",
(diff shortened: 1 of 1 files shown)


picklesrus (approved):
Looks good to me. 
Note, however that the delete button doesn't actually work... :-/

PR #1959, opened 2019-07-01
Update to "Handle promise rejection for sound.play() in preload"

mathiassoeholm:
Since my colleague Michael is on vacation, I added the requested changes to [his pull request](https://github.com/LLK/scratch-blocks/pull/1942).

### Resolves

audio play() called before user interaction [Blockly fix](https://github.com/google/blockly/pull/2162)

### Proposed Changes

Handle promise rejection for sound.play() in preload.

### Reason for Changes

This removes the error, "play() failed because the user didn't interact with the document first."
This is good in itself, but that exception turned out to be the reason for a several seconds delay after the first interaction on iOS.

### Test Coverage

None

### DIFF
--- a/core/workspace_audio.js
+++ b/core/workspace_audio.js
@@ -112,7 +112,7 @@ Blockly.WorkspaceAudio.prototype.preload = function() {
   for (var name in this.SOUNDS_) {
     var sound = this.SOUNDS_[name];
     sound.volume = 0.01;
-    sound.play();
+    sound.play().catch(function() {});
     sound.pause();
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):
LGTM! Thanks!

epicfaace:
Should we make a PR for a similar change into blockly to resolve the issues created by https://github.com/google/blockly/pull/2162?

mathiassoeholm:
@epicfaace I think that would be a good idea, I can see that a similar change has already been proposed in the PR you linked.

PR #1960, opened 2019-07-01
Fix numpad to work correctly with moving cursors

paulkaplan:
### Resolves

Fixes issues with the numpad mentioned by @picklesrus  in https://github.com/LLK/scratch-blocks/pull/1956

### Proposed Changes

Adjust the logic around the numpad in a few ways:

1. Prevent default on the button presses to prevent the (read-only) input from losing focus. When this input loses focus, its selectionStart/End get cleared (at least on Chrome), which makes successive numpad clicks appear to do pretty random looking things.
2. Make the numpad work with selections that are not just the entire thing or just add the end. This makes it possible to use the numpad in an environment where you can also use a mouse. 

/cc @picklesrus testing this is similar to #1956. I recommend testing it both in read-only mode as well as in not-read-only mode, since that will be the situation of "mixed input" where the user touches the number input, showing the numpad, but then uses the keyboard. You should be able to mix keyboard and numpad seamlessly. To get a normal not-read-only input AND the numpad, just remove the `if` at line 173 of field_number and just always show it.

### DIFF
--- a/core/field_number.js
+++ b/core/field_number.js
@@ -282,8 +282,9 @@ Blockly.FieldNumber.prototype.addButtons_ = function(contentDiv) {
 /**
  * Call for when a num-pad number or punctuation button is touched.
  * Determine what the user is inputting and update the text field appropriately.
+ * @param {Event} e DOM event triggering the touch.
  */
-Blockly.FieldNumber.numPadButtonTouch = function() {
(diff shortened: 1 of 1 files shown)


picklesrus (approved):
I can't actually test this on a real touch device, but I played around with it in simulated touch in chrome both with and without readonly set to true. Assuming you tried it out on a real one, this looks good to me!

PR #1961, opened 2019-07-01
Allow quiet inputs to be focused by typing

paulkaplan:
This PR allows fields that prevent the onscreen keyboard using either quiet or read-only mode to be focused by keyboard input. This allows the editors to be used seamlessly on devices that have both a touchscreen and a keyboard.

/cc @picklesrus this is another part of supporting mixed input in a good way. The final part of this will involve changing the way we choose when to try to suppress the onscreen keyboard, so you still need to force it into readonly/quiet mode while testing.

### DIFF
--- a/core/field_textinput.js
+++ b/core/field_textinput.js
@@ -278,7 +278,7 @@ Blockly.FieldTextInput.prototype.showEditor_ = function(
     htmlInput.setSelectionRange(0, 99999);
   }
 
-  this.bindEvents_(htmlInput);
+  this.bindEvents_(htmlInput, quietInput || readOnly);
 
   // Add animation transition properties
   var transitionProperties = 'box-shadow ' + Blockly.FieldTextInput.ANIMATION_TIME + 's';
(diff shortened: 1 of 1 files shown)


picklesrus (commented):
IIUC, you're binding a keypress on the document when the field editor opens and then listening for events from that. Then when the editor gets closed, you remove that listener.  If that's true, here's a question:
In the case where you click in the number field, then click in another text field outside of the blocks (e.g. the sprite info pane) and type. I think this works because you're checking whether the event target is the htmlInput, but I don't have a device with a soft keyboard to check with. It seems to work chrome devtools touch simulation... but I don't trust that.  If you've tested that case on iOS and android/chromeos, then this PR looks good!  (Marking as needs work, but only to get your attention about the question)

paulkaplan:
@picklesrus yup that is a good question. I tested on the blocks playground when you type into the XML playground input, and it does not trigger typing into the block editor (the correct behavior). The reason is that we specifically check that the activeElement is _either_ the input or _not_ an input/textarea. This is similar to how we redirect keyboard input for the stage. But I agree this should get some specific testing before release.

Issue #1762, opened 2018-11-01
Comment does not work well on iOS devices. Cannot input text, and workspace is frozen when remove comment.


xys1836:
### Expected Behavior
Comment should work well on iOS devices.
User should be able to input text to comment text area on iOS devices.
Workspace should not be frozen when remove a comment on iOS devices.
### Actual Behavior
Comment does not work well neither on workspace nor on blocks on iOS devices. 
Keyboard does not appear when touching the text area of a comment. 
Comment can be removed by touching on the top-right "x", however, once the comment is removed, the workspace is frozen. Cannot drag any block from toolbox any more.

### Steps to Reproduce

Long press workspace, add a comment, try to input some text into comment.
Touch the "x" on the top right of a comment, the workspace is frozen. 

### Operating System and Browser
iPad air, iOS 12.1, safari, chrome
iPhone 8, iOS 12.1, safari, chrome

thisandagain:
Thanks for reporting @xys1836. We are in the middle of redoing comments on project pages for the release of 3.0 which address many of these issues.

picklesrus:
I think PR #1967 fixes part of this (the same part that #1963) was trying to fix.  Here's a summary of what else is broken.  

- Removing a comment throws a javascript error: 
```
Cannot read property 'getAttribute' of null
    at Object.Blockly.utils.removeClass (utils.js:100)
    at Blockly.WorkspaceCommentSvg.removeFocus (workspace_comment_svg.js:260)
```

- The minimize triangle and delete x are not clickable when the comment isn't focused. There's a transparent rectangle that sits over the whole comment when it isn't focused so that it is easy to start dragging it from the middle.  That rectangle doesn't belong on top of the workspace comment top bar. I moved it down by changing the y position of the rectangle down by changing  [y coordinate ](https://github.com/LLK/scratch-blocks/blob/develop/core/workspace_comment_render_svg.js#L116) to `'y': Blockly.WorkspaceCommentSvg.TOP_BAR_HEIGHT` . That definitely helps, but the height also needs to be adjusted.
- Selection on iOS is bad. There's no way to edit the existing text "Say something..." because the magnifying glass tool doesn't work at all.
- If you do manage to click the x or minimize buttons, the workspace freezes (it seems to be waiting for a gesture to end?). The only way to get it unstuck is to click on something specific outside of the workspace area (e.g. one of the toolbox categories)

PR #1962, opened 2019-07-03
Fix freeze workspace when delete comments

paulkaplan:
This PR resolves one part of https://github.com/LLK/scratch-blocks/issues/1762, where clicking to delete or minimize a comment on touch would freeze the workspace. 

We were stopping event propagation but not clearing the touch identifier, resulting in a gesture sequence that never terminated.

### DIFF
--- a/core/scratch_bubble.js
+++ b/core/scratch_bubble.js
@@ -354,6 +354,7 @@ Blockly.ScratchBubble.prototype.minimizeArrowMouseUp_ = function(e) {
     }
   }
   e.stopPropagation();
+  Blockly.Touch.clearTouchIdentifier();
 };
 
 /**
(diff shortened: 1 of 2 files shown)


picklesrus (commented):
Are these comment mouse/touch interactions not going through a gesture? I guess I was expecting something like gesture.dispose to get called. If this code path sidesteps that though, this makes sense.

(Doesn't necessarily need work other than answering the question, but not sure if there's a better label)

picklesrus:
In another PR I'm reviewing (https://github.com/LLK/scratch-blocks/pull/1963), I noticed that we are using gestures. Before submitting, I would want to make sure that for these cases we're either not starting one, or, if we are, that we actually cancel it. iOS' behavior here is different than everywhere else, but hopefully I'll get to the bottom of that as part of that review I mentioned.

paulkaplan:
Closing this because we should be using the opt_noCaptureIdentifier option in bindEventWithChecks instead.

PR #1963, opened 2019-07-04
Allow comment textarea to receive focus on touch devices

flemcito:
### Resolves
(partially)
https://github.com/LLK/scratch-blocks/issues/1762

### Proposed Changes

This PR adds a click/touch event handler that programmatically sets focus on the comment textarea.

### Reason for Changes

Behaviour is broken on touch devices as it is impossible to write comments.

### Test Coverage

I have added no additional tests for this

### DIFF
--- a/core/workspace_comment_render_svg.js
+++ b/core/workspace_comment_render_svg.js
@@ -187,7 +187,7 @@ Blockly.WorkspaceCommentSvg.prototype.createEditor_ = function() {
   this.textarea_ = textarea;
   this.textarea_.style.margin = (Blockly.WorkspaceCommentSvg.TEXTAREA_OFFSET) + 'px';
   this.foreignObject_.appendChild(body);
-  Blockly.bindEventWithChecks_(textarea, 'mousedown', this, function(e) {
+  Blockly.bindEventWithChecks_(textarea, 'mousedown', this, function() {
     textarea.focus();
   });
   // Don't zoom with mousewheel.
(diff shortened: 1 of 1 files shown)


picklesrus:
Just wanted to let you know that I haven't forgotten about this.  I see that this works, but I'm still trying to understand why the mechanism we already have in place through the gesture class isn't working on iOS. A workspace comment click/tap should get recognized as a "bubbleClick" For some reason, the touchstart and touchend events are getting split between gestures, so by the time the touchend happens, we think it click was on a workspace not on a comment (or a bubble as it is called in the code).  Now that I finally have a good iOS debugging setup, I plan to dig in some more tomorrow. So sorry for the delay.

picklesrus:
I think I figured it out!
Instead of adding a new mousedown handler, I think what we need is to move the call to textarea focus outside of the setTimeout here: https://github.com/LLK/scratch-blocks/blob/develop/core/workspace_comment_render_svg.js#L667 . That seems to be working for me on an ipad running iOS 11.

picklesrus:
After playing around with this a little more, I do believe the focus change fixes the ability to write in comments, but many other issues are now apparent. Here's a brain dump that we probably should turn into actual issues once we decide what to commit for now.

The transparent rectangle that sits on top of a comment when it isn't focused (so that it is draggable from anywhere in the yellow rectangle) causes problems with the minimize and delete buttons.  Those buttons are only usable when a comment is focused. 

Also, the "says something..." text that sits in the comment isn't editable on mobile. On Android, you can delete it when the comment first appears if you hit delete immediately, but the iOS cursor starts at the beginning of the text and the cursor/magnifying glass tool doesn't work.

There's also this known problem about the workspace freezing.  It may also have to do with the focus. https://github.com/LLK/scratch-blocks/pull/1962

When you delete a comment, it throws a JS error.

flemcito:
Thanks for the update @picklesrus (just got back from vacation). 
Glad to see that there is progress.

paulkaplan:
Closing this in favor of https://github.com/LLK/scratch-blocks/pull/1967 which solved the same problem

Issue #1940, opened 2019-05-21
Please fix it in Python 3.x


Tsepu:
### Expected Behavior

Install with Python 3.x.x

### Actual Behavior

It is only possible in Python 2.x.x

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

towerofnix:
This is already being tracked issue #1679.

Kenny2github:
Duplicate of #1679

PR #1964, opened 2019-07-12
Make build.py compatible with Python 2 and 3

### LINKED ISSUE #1679: build.py requires Python 2.x whose EOL date is 2020
### Expected Behavior

We can build and install scratch-blocks using Python 3.x.

### Actual Behavior

`build.py` requires to use Python 2.x. See [these lines](https://github.com/LLK/scratch-blocks/blob/f5ff6029c223f0f136afe6f9070a267841a42844/build.py#L38).

[PEP 373](https://legacy.python.org/dev/peps/pep-0373/) and <https://pythonclock.org/> says the End Of Life date of Python 2.x is January 1st, 2020. Therefore we should use Python 3.x instead in order to enjoy official security patches quickly.

Is there any plan to update the script?

tansly:
### Resolves

Closes #1679 (and #1940 (duplicate))

### Proposed Changes

Made the build.py script compatible with Python 2 and 3.
Ported the relevant changes from https://github.com/google/blockly/pull/2123 and https://github.com/google/blockly/pull/2125. There were errors remaining, though; fixed those errors as well.

### Reason for Changes

As rationalized in #1679, Python 2 is EOL in 2020.

### Test Coverage

Not added any additional tests. The build succeeded on my machine with Python 2.7.16 and Python 3.7.3. (ran `npm install` and `npm link`. Tested by linking it to `scratch-gui` and running the GUI.)

### DIFF
--- a/i18n/create_messages.py
+++ b/i18n/create_messages.py
@@ -29,11 +29,8 @@ _NEWLINE_PATTERN = re.compile('[\n\r]')
 
 
 def string_is_ascii(s):
-  try:
-    # This approach is better for compatibility 
-    return all(ord(c) < 128 for c in s)
-  except UnicodeEncodeError:
-    return False
+  # This approach is better for compatibility
+  return all(ord(c) < 128 for c in s)
 
 def load_constants(filename):
   """Read in constants file, which must be output in every language."""
--- a/i18n/create_messages.py
+++ b/i18n/create_messages.py
@@ -30,8 +30,8 @@ _NEWLINE_PATTERN = re.compile('[\n\r]')
 
 def string_is_ascii(s):
   try:
-    s.decode('ascii')
-    return True
+    # This approach is better for compatibility 
+    return all(ord(c) < 128 for c in s)
   except UnicodeEncodeError:
     return False
 
@@ -81,14 +81,14 @@ def main():
       print('ERROR: definition of {0} in {1} contained a newline character.'.
             format(key, args.source_lang_file))
       sys.exit(1)
-  sorted_keys = source_defs.keys()
-  sorted_keys.sort()
+  sorted_keys = sorted(source_defs.keys())
 
   # Read in synonyms file, which must be output in every language.
   synonym_defs = read_json_file(os.path.join(
       os.curdir, args.source_synonym_file))
+  # synonym_defs is also being sorted to ensure the same order is kept
   synonym_text = '\n'.join(['Blockly.Msg.{0} = Blockly.Msg.{1};'.format(
-      key, synonym_defs[key]) for key in synonym_defs])
+      key, synonym_defs[key]) for key in sorted(synonym_defs)])
 
   # Read in constants file, which must be output in every language.
   constants_text = load_constants(os.path.join(os.curdir, args.source_constants_file))
--- a/build.py
+++ b/build.py
@@ -36,7 +36,7 @@
 
 import sys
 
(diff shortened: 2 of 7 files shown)


filips123:
@thisandagain Any update? Python 2.7 support will be discontinued in about 2 months.

adroitwhiz:
Python 2 is now EOL.

i18n/create_messages.py
```diff
@@ -30,8 +30,8 @@
 
 def string_is_ascii(s):
   try:
-    s.decode('ascii')
-    return True
+    # This approach is better for compatibility 
+    return all(ord(c) < 128 for c in s)
   except UnicodeEncodeError:
     return False
```
adroitwhiz: Will `all(ord(c) < 128 for c in s)` raise a `UnicodeEncodeError` or can this try block be removed?

apple502j: I think this could be removed.

tansly: I think you're right, I've removed it.

ericrosenbaum:
@picklesrus do you think this would also solve the problem in https://github.com/LLK/scratch-blocks/issues/2092? Perhaps we should put this back in the queue to review

apple502j:
@ericrosenbaum It should (because with this PR python version doesn't matter)

tansly:
When I opened the PR, I tested this both with `python2` and `python3` and it worked.

However that's a quite a while ago. I haven't been following the repo actively; were there any other big changes to `build.py` since then? If so, we might need to take another look and test it again to make sure there weren't any new changes that introduced other incompatibilities.

apple502j:
@tansly There is one change, but it shouldn't matter, there is no conflict after all

thebravoman:
Hey @thisandagain, @picklesrus,

I wanted to reach out to you again concerning this issue. I've shared my thoughts in https://github.com/LLK/scratch-blocks/issues/1679#issuecomment-1305244999. Main point is that it is difficult to involve people in 2022 if this still depends on python2. 

Any chance this PR could be merged?

thebravoman:
Merged branch tansly:build-python3 locally. sh -c did not work and python was again not found, but did only python and got:
 
```
% python build.py && webpack
  File "/Users/kireto/axles/code/scratch-blocks/build.py", line 580, in <module>
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1821, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'google-closure-compiler'
```

csarnataro:
Hi @thebravoman, I came here to ask the same thing, that is, I would love to see this PR merged.
After downloading the branch `tansly:build-python3`, I launched `npm install` and it worked at first attempt. So I would just say that the new build script is working. :) BTW, thanks @tansly for your work.

Regarding your specific issue `No such file or directory: 'google-closure-compiler'` I assume that the build process, as a whole, requires to install all dependencies, e.g. google-closure-compiler, via `npm install`.
What I mean is that probably just `pyhton build.py` is not enough if you don't have all dependencies installed before.

delasare:
Reviewed, Tested and Merged for move to GitHub Actions.

PR #1967, opened 2019-07-24
Move call to focus element outside of the the setTimeout. iOS is pick…

picklesrus:
…y about what things you can do outside of a direct user action and I think focus might be one of them.  With this change comments are now editable on iOS.

### Resolves

Part of https://github.com/LLK/scratch-blocks/issues/1762

### Proposed Changes

On workspace comment click, call focus immediately rather than in a set timeout

### Reason for Changes

The call to focus isn't working on iOS in the setTimeout. I think this is because it is no longer part of the user action/click.


kchadha (approved):
LGTM!

PR #1968, opened 2019-07-26
Use quiet inputs/numpad for touch interaction instead of based on user agent

paulkaplan:
Instead of using user agent detection for showing the numpad and deciding whether or not to use quiet inputs, use the original interaction on the field to determine that. The goal is to show the touch ui for anyone interacting via touch, not based on device.

/ht to @picklesrus  for pairing on this


picklesrus (approved):
LGTM

rachel-fenichel:
@samelhusseini FYI

PR #1970, opened 2019-07-29
Several fixes for comments

### LINKED ISSUE #1631: Cannot edit the text of a comment on touch devices
### Expected Behavior

Should be able to edit the text of a comment.

### Actual Behavior

Cannot edit the text of a comment on a touch device.

### LINKED ISSUE #1758: Firefox requires two clicks to focus a block comment
### Expected Behavior

You should be able to type into the comment after clicking once. On Firefox on a mac it takes two clicks.

This is a tradeoff of fixing #1631.

Pasting the relevant history from that bug here, but #1631 actually has code links, so probably easier to read there. 

setTextArea focus on the Comment class gets called on mouseup and is bound here. It is bound on mouse up due to the previously mentioned FIrefox problem.

In order to fix that, the right thing to do is probably to bind on mouse down and let the code proceed through the usual gesture handling class. That means doing something more like what workspace_comment_svg does in pathMouseDown i.e.
var gesture = this.workspace.getGesture(e); if (gesture) { gesture.handleBubbleStart(.....); }

But, because the mousedown is on a text area (and utils.isTargetInput returns true, doStart gives up and cancels the gesture. I believe this is why the workspace comment has layer of svg in it. I think that layer catches the bubble click and gets it past the target input check.

Next, even if we fixed that, by the time we decide this was a bubble click gesture's handleUp, doBubbleClick tries to call startBubble_.setFocus(). Except, startBubble_ (which is of type ScratchBubble) doesn't have the setFocus method on it. Only WorkspaceCommentSvg has this method. It turns out, in the case of workspace comments, the startBubble_ member variable isn't a Bubble or a subclass of Bubble. Instead, it just happens to have the setFocus method focus and works. So, we have some mess to clean up. I haven't internalized the reach of cleaning it up yet so I'm going to put this aside for now.

### LINKED ISSUE #1762: Comment does not work well on iOS devices. Cannot input text, and workspace is frozen when remove comment.

paulkaplan:
Together with @picklesrus and @kchadha, here are several fixes for comments, block and workspace, for different platforms.

I'll break down the things to test by commit:
1. d95476f You should not need to click/tap more than once to delete or minimize a workspace comment
2. 3a992ac Delete and minimize icons no longer cause the workspace to freeze
3. b900bd2 Clicking the comments now focuses them correctly on mobile. Also you should now be able to select the comment text using the system text selection tools (i.e. text selection cursors should show up on iOS, you can delete the placeholder text, etc). Also this fixes #1758 
4. 469fe8b No errors should be thrown when deleting a workspace comment

Fixes https://github.com/LLK/scratch-blocks/issues/1758
Fixes https://github.com/LLK/scratch-blocks/issues/1631
Fixes https://github.com/LLK/scratch-blocks/issues/1762

### DIFF
--- a/core/scratch_block_comment.js
+++ b/core/scratch_block_comment.js
@@ -231,7 +231,8 @@ Blockly.ScratchBlockComment.prototype.createEditor_ = function() {
   this.textarea_ = textarea;
   this.textarea_.style.margin = (Blockly.ScratchBlockComment.TEXTAREA_OFFSET) + 'px';
   this.foreignObject_.appendChild(body);
-  Blockly.bindEventWithChecks_(textarea, 'mousedown', this, this.textareaFocus_);
+  Blockly.bindEventWithChecks_(textarea, 'mousedown', this,
+      this.textareaFocus_, true, true); // noCapture and do not prevent default
   // Don't zoom with mousewheel.
   Blockly.bindEventWithChecks_(textarea, 'wheel', this, function(e) {
     e.stopPropagation();
@@ -255,6 +256,18 @@ Blockly.ScratchBlockComment.prototype.createEditor_ = function() {
     labelText: this.label_
   };
 };
+
+/**
+ * Handle text area click, make sure to stop propagation to allow default selection behavior.
+ * @param {!Event} e Mouse up event.
+ * @private
+ */
+Blockly.ScratchBlockComment.prototype.textareaFocus_ = function(e) {
+  Blockly.ScratchBlockComment.superClass_.textareaFocus_.call(this, e);
+  // Stop event from propagating to the workspace to make sure preventDefault _is not called_.
+  e.stopPropagation();
+};
+
 
 /**
  * Callback function triggered when the bubble has resized.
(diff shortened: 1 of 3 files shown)


picklesrus (approved):
LGTM

kchadha (approved):
LGTM

PR #1971, opened 2019-07-29
Use placeholders instead of default text for comments, fix focusing

### LINKED ISSUE #1601: Switching to a sprite focuses on most recent block comment
### Expected Behavior

Selecting a sprite should not select/focus on any of the comments.

### Actual Behavior

If there are any block comments on a sprite you switch to it will add a blinking cursor to the comment.  If the project is controlled by keyboard keys you could end up typing in the comment.

### Steps to Reproduce

Create a project
Add a second sprite
In one sprite drag a block onto the workspace
right click the block and add a comment
type something in the comment
Switch to the other sprite
switch back
Note: the cursor is on the comment.

Note: This is not the case with workspace comments.
Note: if there is more than one block comment it will highlight the newer one.

Chrome on Mac

paulkaplan:
After chatting with @carljbowman, we figured the best behavior w/r/t creating and focusing comments is to use the `placeholder` property for the "say something..." instructions, and to fix block comment focus so they can be focused when created from the contextmenu callback.

Just want to check though, @rachel-fenichel @picklesrus @kchadha was there a reason we went with a default value instead of a placeholder originally?

In addition to fixing the placeholder issue, it also fixes focusing after creation, which does not work on iOS because the element is not attached to the dom during `createEditor`, so we cannot remove the setTimeout to get focus to work. It needs to be called after createEditor is done, with workspace comments this is done explicitly in the contextmenu code, so i followed that pattern.

If you want these as two separate PRs, i can do that.

By moving focus out of `createEditor` and explicitly calling it in the context menu code, this also fixes https://github.com/LLK/scratch-blocks/issues/1601

### DIFF
--- a/core/contextmenu.js
+++ b/core/contextmenu.js
@@ -292,7 +292,7 @@ Blockly.ContextMenu.blockCommentOption = function(block) {
     // If there's no comment, add an option to create a comment.
     commentOption.text = Blockly.Msg.ADD_COMMENT;
     commentOption.callback = function() {
-      block.setCommentText(Blockly.Msg.WORKSPACE_COMMENT_DEFAULT_TEXT);
+      block.setCommentText('');
     };
   }
   return commentOption;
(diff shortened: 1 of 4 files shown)


rachel-fenichel:
I didn't know about that property!  This looks good to me (and simpler than the previous version).

kchadha (approved):
LGTM!

PR #1983, opened 2019-08-07
Use correct msg when showing delete blocks confirm dialog.

mfrtrifork:
### Resolves

The incorrect text "Delete all 1 blocks?" and replaces it with "Delete block".

### Proposed Changes

Simply checks the number of blocks to delete and if only 1 then use another message.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1504,7 +1504,8 @@ Blockly.WorkspaceSvg.prototype.showContextMenu_ = function(e) {
         deleteNext();
       } else {
         Blockly.confirm(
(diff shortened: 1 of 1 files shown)


thisandagain:
@picklesrus Would you mind taking a look at this over the next week or so?

picklesrus:
The fact that were a dialog when there was only 1 block was actually a bug. We only showed it for blocks with inputs (e.g. move (10) steps) but not for blocks without them (e.g when green flag clicked).  #1990 fixed that which should mean we never show this message when deleting a single bock.  Let me know if I'm misunderstanding though!  Thanks!

PR #1985, opened 2019-08-07
Use click event to avoid gesture problems freezing workspace

### LINKED ISSUE #4920: Cannot use blocks color eyedropper more than once, it freezes on iOS
--- Edited to add details by @paulkaplan 
Trying to use the blocks eye dropper more than once causes the UI to freeze when using iOS.
Also dragging the eyedropper around causes the whole page to move around.

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/4920 and an unreported issue where you could not use the "clear" and "fill" buttons on the microbit extension.

### Proposed Changes

Instead of using `mousedown`, for events, and then having to disable the gestures, use `click` instead, which both works on touch and does not start a gesture. 

This incidentally also fixes a problem with the `field_matrix` where the `e.button` property did not exist for touch, causing those buttons to not work. But the button property does exist for `click` events, even on touch.

/cc @khanning might be interested

### DIFF
--- a/core/field_colour_slider.js
+++ b/core/field_colour_slider.js
@@ -341,7 +341,7 @@ Blockly.FieldColourSlider.prototype.showEditor_ = function() {
     button.appendChild(image);
     div.appendChild(button);
     Blockly.FieldColourSlider.eyedropperEventData_ =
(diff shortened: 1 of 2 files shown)


kchadha (approved):
LGTM!

PR #1990, opened 2019-08-13
Now using correct delete count to check if should show dialog

mathiassoeholm:
# Proposed Changes

Checks the same delete count var, which is shown in the confirmation dialog.

### Reason for Changes

Currently we check if deleteList.length < 2 and then show a dialog with deleteCount if its above.
deleteList.length and deleteCount are not always the same. In a block with input, each input gets added to the deleteList, hence the size is larger than the amount og blocks.
Therefore it's the wrong value to check.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -1500,7 +1500,7 @@ Blockly.WorkspaceSvg.prototype.showContextMenu_ = function(e) {
       if (ws.currentGesture_) {
         ws.currentGesture_.cancel();
       }
-      if (deleteList.length < 2 ) {
+      if (deleteCount < 2 ) {
         deleteNext();
       } else {
         Blockly.confirm(


kchadha:
@mathiassoeholm Do you mind providing steps to reproduce the issue that this PR is fixing?

mathiassoeholm:
Sure :-)

1. Go to https://scratch.mit.edu/projects/editor
2. Drag a "move (10) steps" block into the canvas
3. Right-click the canvas, press Delete 1 block

Result:
There's a confirmation dialog asking: Delete all 1 blocks?

Expected result:
There should not be a confirmation dialog, since we're only deleting one block.

Now try the same steps but with the "when green flag clicked" block.
Result: It behaves correctly, because this block has no inputs.

picklesrus (approved):
Thank you! Sorry for the delay on the review. LGTM!

PR #2023, opened 2019-12-18
Fix offset of comments in stacks

### LINKED ISSUE #2022: The comment's line is too long.
### Expected Behavior

The comment line needs to be close to the block

### Actual Behavior

The comment line is so long that it's too far away from the blocks

### Steps to Reproduce

* Add two blocks like the follow images:
The input field need input long text.
image
* Add comment for this two blocks.
comment too long

Chrome 78.0.3904.108 x64

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/2022

### Proposed Changes

Instead of bumping comments over by a factor of the stack width, bump comments over by the "overhang", meaning the difference between the width of the stack (this block and all next) and the width of this block. The anchor is already at the right edge so we don't need to offset by a full block width ever.

Thanks @picklesrus for helping pair on this!
### Reason for Changes

Current behavior
Screen Shot 2019-12-18 at 3 39 11 PM

New behavior
Screen Shot 2019-12-18 at 3 38 29 PM

### DIFF
--- a/core/scratch_block_comment.js
+++ b/core/scratch_block_comment.js
@@ -313,20 +313,17 @@ Blockly.ScratchBlockComment.prototype.autoPosition_ = function() {
         this.iconXY_.x + minimizedOffset;
     this.y_ = this.iconXY_.y - (Blockly.ScratchBubble.TOP_BAR_HEIGHT / 2);
   } else {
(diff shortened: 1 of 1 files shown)


kchadha (approved):
LGTM!

PR #2036, opened 2020-01-29
Scripts for automating translation in travis.

### LINKED ISSUE #1608: Automate translation updates
### Expected Behavior

When changes to the `messages.js` file merged, the `en.json` source on Transifex should be updated. Updated translations should also be downloaded to make sure that all languages include any new keys (even though they will initially be the default English)

On deploys we should download currently reviewed translations and regenerate `scratch_msgs.js`

chrisgarrity:
### Resolves

Fixes #1608 

### Proposed Changes

i18n/sync_translations.sh: doesn’t do anything if it isn’t a cron job. Otherwise, it updates and tests messages, then commits and pushes to develop branch if there are no errors.

.travis.yml: adds syncing translations before the build script. Trying to keep the .travis config changes minimal.

#### Questions
* Does the sync script need to check out develop since that is the branch that is getting built by the cron job?
* For now it's only staging changes in the msg directory (that's where translations show up). Is there any reason to stage more?
* added the redirect to `/dev/null` on the `git remote` command to avoid leaking the $GH_TOKEN, does the `git push` also need the redirect?
* would it be better to do a more major refactor of the travis.yml to have a `jobs` section?
* Since this does a commit on develop, I think it will kick off another travis build (Yes?). I'm assuming that build should be the one to build and release, so the cron job should skip releasing.

### Reason for Changes
Better ergonomics

### Testing

Testing involves getting travis to build and release.

### DIFF
--- a/i18n/sync_translations.sh
+++ b/i18n/sync_translations.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+if [ "${TRAVIS_EVENT_TYPE}" == "cron" ]
+  then
+    echo "Starting translation sync"
+    set -ev
+    git checkout develop
+    # update translations, and test any updated messages
+    npm run translate
+    npm run translate:update
+    npm run test:messages
+    # stage any changes in the msg directory
+    git add ./msg
+    git commit -m 'update translations from transifex'
+    # add remote, make sure that API token doesn't end up in the log
+    git remote add origin-translation https://${GH_TOKEN}@github.com/LLK/scratch-blocks.git > /dev/null 2>&1
+    git push --set-upstream origin-translation develop
+fi
--- a/i18n/sync_translations.sh
+++ b/i18n/sync_translations.sh
@@ -10,7 +10,7 @@ if [ "${TRAVIS_EVENT_TYPE}" == "cron" ]
     npm run test:messages
     # stage any changes in the msg directory
     git add ./msg
-    git commit -m 'update translations from transifex'
+    git commit -m '[skip ci] Update translations from transifex'
     # add remote, make sure that API token doesn't end up in the log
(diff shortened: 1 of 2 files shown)


i18n/sync_translations.sh
```diff
@@ -0,0 +1,17 @@
...
+    echo "Starting translation sync"
+    set -ev
+    git checkout develop
+    # update translations, and test any updated messages
+    npm run translate
+    npm run translate:update
+    npm run test:messages
+    # stage any changes in the msg directory
+    git add ./msg
+    git commit -m 'update translations from transifex'
```
rschamp: When this gets pushed to develop, I think it will kick off another job, right? It looks like the build published by this cron job will include the changes to `msg`, so that second build will be redundant. I think we should try to make this silent to ci, so that it makes the commit and pushes it, but that one doesn't kick off a new build (the capitalization also made me twitch :)

```suggestion
    git commit -m '[skip ci] Update translations from transifex'
```

rschamp (changes requested):

rschamp (approved):
> Does the sync script need to check out develop since that is the branch that is getting built by the cron job?

I don't think it's necessary, but it's probably good to be explicit. Although I suppose you could argue that it would be more flexible if we didn't hard-code the branch, so the cron job could configure which branch to build? I think it's fine for now as-is.

> For now it's only staging changes in the msg directory (that's where translations show up). Is there any reason to stage more?

I think it's good that we're telling it exactly what to commit, since otherwise we might push something we don't expect.

> added the redirect to /dev/null on the git remote command to avoid leaking the $GH_TOKEN, does the git push also need the redirect?

I think since Travis blanks out any instance of secure variable values in logs we don't even need that redirect on the remote command. No problem with it being there though.

> would it be better to do a more major refactor of the travis.yml to have a jobs section?

Better to just port to CircleCI, and definitely not necessary for this change.

PR #2058, opened 2020-04-07
Localize custom block used alert

### LINKED ISSUE #2050: Delete custom block alert text is not localized
### Expected Behavior

If you try to delete a custom block definition without deleting all the users, you get an alert pop up with a message (see #1151). It should have the message in the currently selected language.

### Actual Behavior

 It does not use a Scratch Msg, so it is not translatable and the message is always in English. See:
https://github.com/LLK/scratch-blocks/blob/develop/core/block_dragger.js#L283..L285

### Steps to Reproduce
* switch to a language other than English
* create a custom block
* use it in another stack
* Try to delete the definition

apple502j:
### Resolves
Resolves #2050

### Proposed Changes
Adds a l10n message for trying to delete custom blocks in use

### Reason for Changes
Localization!

### Test Coverage
Tests FAIL, and it's expected: https://github.com/LLK/scratch-blocks/wiki/Localization


core/block_dragger.js
```diff
@@ -282,7 +282,7 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
           // Check for call blocks with no associated define block.
           if (!Blockly.Procedures.getDefineBlock(procCode, ws)) {
             // TODO:(#1151)
-            alert('To delete a block definition, first remove all uses of the block');
+            alert(Blockly.Msg.PROCEDURE_USED);
```
chrisgarrity, line 284: This will only get the pop-up if you drag the definition off the workspace. There is also a version in the context menu procedure: https://github.com/LLK/scratch-blocks/blob/a355ca0b6353cbcfa906472d01a21e9940a390c9/blocks_vertical/vertical_extensions.js#L178-L180

Also, I'm not quite sure whether these are going to get dynamically updated if the language changes.

chrisgarrity (changes requested):
There's a second place that needs to change. 

Also, before this can be merged it will need to be rebased on the latest develop branch. I'll need to push a commit that pulls all the translations for the new string. Since we recently added Armenian, I can't easily do that without introducing conflicts.

chrisgarrity (approved):
Sorry for the additional commits that got pulled in when I tried to rebase/update strings (we'd added another language). The tests should pass now. I'll merge once that's confirmed (they ran locally for me).
Thanks!

PR #2064, opened 2020-05-07
Update chromedriver and selenium to fix failing tests

chrisgarrity:
Chromedriver 81 is available, but gets an error running the install script. Using chromedriver 80 instead. It fixed the tests for me - hopefuly it does for travis too.

@ericrosenbaum I'm tagging you for review since you're next on the deploy schedule and this needs to be fixed by next week. We haven't updated translations for 2 deploy cycles due to the tests failing. 

Testing:
If the travis build of the PR doesn't fail, then it's fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "devDependencies": {
     "async": "2.6.0",
-    "chromedriver": "^75.0",
+    "chromedriver": "80.0.1",
     "copy-webpack-plugin": "4.5.1",
     "eslint": "^4.16",
     "event-stream": "3.3.4",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
   },
   "devDependencies": {
     "async": "2.6.0",
-    "chromedriver": "^75.0",
+    "chromedriver": "^80.0.0",
```
apple502j:
```suggestion
    "chromedriver": "80.0.1",
```

Node v8 support was dropped on `80.0.2`:  giggio/node-chromedriver#261

apple502j:
@chrisgarrity It seems like we need to either use `80.0.1` or bump Node version to at least 10.

chrisgarrity:
@apple502j thanks for the pointer to the chromedriver issue - that explains a lot.

package.json
chrisgarrity: 👍

chrisgarrity:
@rschamp since this was fixed by upgrading node to version 12, I'm closing this PR. It might still be a good idea to change the chromedriver version to `^81.0` at some point

PR #2066, opened 2020-05-11
Use FieldColourSlider's internal hue_, saturation_, brightness_ in slider callbacks

### LINKED ISSUE #1858: Colour slider show the wrong color
### Expected Behavior

Show the correct color.

### Actual Behavior

Show the wrong color.

### Steps to Reproduce

The main step is to drag the brightness slider to 0.

feature

Win 10 1809

Chrome Version 71.0.3578.98 (Official Build) (64-bit)

----

It seems because `goog.color.hsvToHex(hue, saturation, 0)` always return `#00000`.

So I try to return the hex from hsv

```js
/**
 * Return the current colour.
 * @return {string} Current colour in '#rrggbb' format.
 */
Blockly.FieldColourSlider.prototype.getValue = function() {
  if(this.hue_){
    return goog.color.hsvToHex(this.hue_, this.saturation_, this.brightness_);
  }
  else {
    return this.colour_;
  }
};
```

And it works.

I hope this will be helpful.

### LINKED ISSUE #2048: In the touching color block, reducing the saturation to 0 changes the color to red
### Expected Behavior

If you reduce the saturation to 0 on the touching color block and then increase it again the color selected remains the same as it was before.

### Actual Behavior

Whatever color was selected before reducing the saturation to 0, it becomes red after doing so.

saturationBug

### Steps to Reproduce

Create a new project
Drag out a "touching color" or a "color is touching" block
Click on one of the colors in the block to bring up the color selectr
Make sure the color value is set to something other than red
Drag the "saturation" value to 0 
Drag the "saturation" value back up so it reveals a color
Note: the value for color has not changed, but the color that appears in the color selected is red
Click off the block to close the color selector
Open the color selector again
Note: it shows the color value to be 0 (red)

This does not happen in the paint editor.
This is similar to https://github.com/LLK/scratch-blocks/issues/1858 and https://github.com/LLK/scratch-gui/issues/3985, but the behavior is slightly different.

Mac Chrome

### LINKED ISSUE #3985: FieldColourSlider: error when brightness from zero to other value
### Expected Behavior

Block's color field slider can work well.

### Actual Behavior

Color field get unexpect color when brightness from 0 to other value.

### Steps to Reproduce

See the gif: 
slide color

Win10 x64, Google Chrome 70.0.3538.110

adroitwhiz:
### Resolves

Resolves #2048
Resolves #1858
Resolves https://github.com/LLK/scratch-gui/issues/3985

### Proposed Changes

This PR changes the slider callbacks in `FieldColourSlider` to set the field's color based on the field's internal `hue_`, `saturation_`, and `brightness_` introduced in #1148, rather than converting the field's RGB hex color into HSV and using those values.

### Reason for Changes

If you convert an HSV color value into RGB, it's impossible to determine its hue if its saturation or brightness are 0. By converting the color field's stored RGB color back into HSV every time a slider is changed, the hue will thus be lost and reset to 0 every time the saturation or brightness sliders are dragged down to 0, and the saturation will be lost and reset to 0 when the brightness slider is 0 as well.

To avoid this, just use the HSV values already stored in the field, which are only reinitialized from RGB when the color picker is opened.

### Test Coverage

Tested manually

### DIFF
--- a/core/field_colour_slider.js
+++ b/core/field_colour_slider.js
@@ -245,19 +245,18 @@ Blockly.FieldColourSlider.prototype.sliderCallbackFactory_ = function(channel) {
   return function(event) {
     if (!thisField.sliderCallbacksEnabled_) return;
     var channelValue = event.target.getValue();
-    var hsv = goog.color.hexToHsv(thisField.getValue());
     switch (channel) {
       case 'hue':
-        hsv[0] = thisField.hue_ = channelValue;
+        thisField.hue_ = channelValue;
         break;
       case 'saturation':
-        hsv[1] = thisField.saturation_ = channelValue;
+        thisField.saturation_ = channelValue;
         break;
       case 'brightness':
-        hsv[2] = thisField.brightness_ = channelValue;
+        thisField.brightness_ = channelValue;
         break;
     }
-    var colour = goog.color.hsvToHex(hsv[0], hsv[1], hsv[2]);
+    var colour = goog.color.hsvToHex(thisField.hue_, thisField.saturation_, thisField.brightness_);
     if (thisField.sourceBlock_) {
       // Call any validation function, and allow it to override.
       colour = thisField.callValidator(colour);


fsih (approved):
This fix looks good to me!

We were unable to figure out why to was written this way in the first place

PR #2068, opened 2020-05-13
Make event filtering stricter for move events (blockly#1892)

### LINKED ISSUE #4400: Changing  sprite of a "(attribute) of (sprite)" in block palette rearranges blocks in the workspace (sometimes)
### Expected Behavior

The "(attribute) of (sprite)" block does not destroy your project.

### Actual Behavior

It does.

Case in point:

Many complex scripts all layered onto each other.

Every script is positioned at (0, 0).

Zip file of this project attached (see "Battle Controller" sprite): [Untitled-297.zip](https://github.com/LLK/scratch-gui/files/2776769/Untitled-297.zip)

### Steps to Reproduce

I can't reliably repro this, unfortunately, but it's happened to me several times while developing that large project. This only ever occurs immediately after I change the sprite selected in the dropdown menu. It's a huge pain and always causes either lost progress (restoring from a page reload) or rearranging all the scripts back to where they're supposed to be.

Firefox 65.0b11, Linux (Debian).

adroitwhiz:
Well, this was a fun way to spend 3 hours!

### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/4400

### Proposed Changes

This PR brings over the event merging changes from https://github.com/google/blockly/pull/1892/.

Regarding the other commits:
- "Fix undoing a drag to connect" is part of the `wouldConnectBlock` changes. I was wondering why `applyConnections` and `render` were being called after `fireMoveEvent_` in Blockly but before it here, and it turns out that was part of another bugfix. I think it's best to apply it here too; [fixing the event merging would otherwise break behavior](https://github.com/google/blockly/issues/1780#issuecomment-385087792).
- "Add wouldConnectBlock to InsertionMarkerManager": see below discussion.

### Reason for Changes

Changing the sprite menu in the "(attribute) of (sprite)" block triggers a workspace update, which rebuilds the entire Blockly workspace.

When rebuilding a workspace, blocks are first created with their position at the origin and then moved to their proper positions. The VM should be able to track that via events emitted from Blockly, e.g.:
- Block A is created at position (0, 0).
- Block A is moved to position (500, 100).

However, in certain cases (e.g. when you change the menu in the toolbox/flyout, maybe other cases?), these events are emitted twice:
- Block A is created at position (0, 0).
- Block A is moved to position (500, 100).
- Block A is created at position (0, 0).
- Block A is moved to position (500, 100).

The previous event merging behavior would incorrectly merge the second move event into the first, as such:
- Block A is created at position (0, 0).
- Block A is moved to position (500, 100).
- Block A is created at position (0, 0).

This would cause the VM to think that Block A should remain at (0, 0). This would be repeated for every top-level block in the current sprite's workspace, causing them all to be reset to (0, 0).

### Test Coverage

Tests were ported from Blockly

### DIFF
--- a/core/insertion_marker_manager.js
+++ b/core/insertion_marker_manager.js
@@ -188,6 +188,17 @@ Blockly.InsertionMarkerManager.prototype.wouldDeleteBlock = function() {
 };
 
 /**
+ * Return whether the block would be connected if dropped immediately, based on
+ * information from the most recent move event.
+ * @return {boolean} True if the block would be connected if dropped
+ *   immediately.
+ * @package
+ */
+Blockly.InsertionMarkerManager.prototype.wouldConnectBlock = function() {
+  return !!this.closestConnection_;
+};
+
+/**
  * Connect to the closest connection and render the results.
  * This should be called at the end of a drag.
  * @package
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -245,9 +245,9 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
     // These are expensive and don't need to be done if we're deleting.
     this.draggingBlock_.moveConnections_(delta.x, delta.y);
     this.draggingBlock_.setDragging(false);
+    this.fireMoveEvent_();
     this.draggedConnectionManager_.applyConnections();
     this.draggingBlock_.render();
-    this.fireMoveEvent_();
     this.draggingBlock_.scheduleSnapAndBump();
   }
   this.workspace_.setResizesEnabled(true);
--- a/core/block_dragger.js
+++ b/core/block_dragger.js
@@ -246,8 +246,12 @@ Blockly.BlockDragger.prototype.endBlockDrag = function(e, currentDragDeltaXY) {
     this.draggingBlock_.moveConnections_(delta.x, delta.y);
     this.draggingBlock_.setDragging(false);
     this.fireMoveEvent_();
-    this.draggedConnectionManager_.applyConnections();
-    this.draggingBlock_.render();
+    if (this.draggedConnectionManager_.wouldConnectBlock()) {
+      // Applying connections also rerenders the relevant blocks.
+      this.draggedConnectionManager_.applyConnections();
+    } else {
+      this.draggingBlock_.render();
+    }
     this.draggingBlock_.scheduleSnapAndBump();
   }
   this.workspace_.setResizesEnabled(true);
--- a/core/dragged_connection_manager.js
+++ b/core/dragged_connection_manager.js
@@ -126,6 +126,16 @@ Blockly.DraggedConnectionManager.prototype.wouldDeleteBlock = function() {
 };
 
 /**
+ * Return whether the block would be deleted if dropped immediately, based on
+ * information from the most recent move event.
+ * @return {boolean} true if the block would be deleted if dropped immediately.
+ * @package
+ */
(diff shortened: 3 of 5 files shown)


adroitwhiz:
@rachel-fenichel Sorry about the lack of attribution on https://github.com/LLK/scratch-blocks/pull/2068/commits/fccb5e11d33b0a823a4964870e65ae3e12092d9b! Fixed it now

rachel-fenichel:
Oh I'm not worried about attribution! Just sad that it took lots of digging to find a previously solved problem.

cwillisf:
@picklesrus any chance you have time to review this? I feel like I'm in over my head in terms of scratch-blocks. In talking with @adroitwhiz (and referring to 6275e1137c9be43ff8a94dcf8139196cb384e657) it does look like adding `wouldConnectBlock` to both `DraggedConnectionManager` and `InsertionMarkerManager` makes sense.

EDIT: to clarify, I mean that adding `wouldConnectBlock` to `InsertionMarkerManager` makes sense to me given that the original change added `wouldConnectBlock` to `DraggedConnectionManager` :)

core/insertion_marker_manager.js
```diff
@@ -187,6 +187,17 @@ Blockly.InsertionMarkerManager.prototype.wouldDeleteBlock = function() {
...
 };
 
+/**
+ * Return whether the block would be connected if dropped immediately, based on
+ * information from the most recent move event.
+ * @return {boolean} True if the block would be connected if dropped
+ *   immediately.
+ * @package
+ */
+Blockly.InsertionMarkerManager.prototype.wouldConnectBlock = function() {
```
fsih, line 197: Why do we need this function?

fsih (approved):
LGTM! Since it's bringing in an upstream change I think it should be safe, but would feel better if @picklesrus, @kchadha or @paulkaplan sanity-checked

rachel-fenichel:
As I recall, scratch-blocks doesn't use the dragged connection manager and so it doesn't matter if you add the code in there or not.

The dragged connection manager handled blockly-style yellow highlights. The insertion marker manager handles insertion markers. They share an API and for a while both files existed next to each other, but I don't think you're ever instantiating a dragged connection manager. And there's no inheritance going on either.

picklesrus:
What Rachel said :) Scratch blocks instantiates InsertionmarkerManager in block dragger, not DraggedConnectionManager so that code shouldn't matter.  Looks like we're setting closestConnection_ in the same way that Blockly does too so I'm pretty sure this is fine for the blocks. During testing we should verify that insertion markers continue to work as we expect. 

To me, knowing whether the VM is somehow depending on the event that was merged is trickier to verify. @adroitwhiz would you work with folks to make a test plan for this? Happy to help!  In addition to testing that the bug with that sensing block is fixed and insertion markers still work as expected, here are some random initial thoughts on things that may generate events: switching sprites, switching between code/sound/paint tabs, switching to the stage & back to a sprite, editor->project page->editor, share the love, changing values in the sprite pane, moving blocks around, undo/redo, clean up blocks, deleting blocks in a way that resizes the workspace, load project from a file.   I think considering something like this list while the project is running is important too. Maybe @cwillisf has thoughts on whether I should be concerned about how these things interact with the VM or not?

fsih:
Thanks Katelyn for helping us ideate about a test plan!

picklesrus (changes requested):
Whoops! Sorry! Previous comment (deleted now) was meant for a different PR. Ignore me -- I have too many tabs open :)

PR #2069, opened 2020-05-15
Don't scroll workspace when clearing it upon reloading from XML

### LINKED ISSUE #4615: changing the ____ of ____ sensing block sprite scrolls the workspace to the top
### Expected Behavior

If you change the sprite/backdrop of the ____ of _____ sensing block should not change the view of the block workspace.

### Actual Behavior

When you change the sprite it scrolls the workspace to the top.  

This issue was brought to our attention from the community by user 34oo3958.

### Steps to Reproduce

Create a new project
Add a second sprite (any method)
On the second sprite drag a block (any) to the top of the workspace
Add a ___ of ___ block to the workspace
Scroll down and bring the ___ of ___ block with you.  
Make sure you can't see the first block you dragged out
Change the second part of that block, the sprite/stage
Note: you have now been scrolled to the top.

Mac Chrome

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/4615

(See also [this discussion on scratch-gui#4400](https://github.com/LLK/scratch-gui/issues/4400#issuecomment-462994133))

### Proposed Changes

This PR adds a parameter to `WorkspaceSvg.clear` that allows callers to specify whether the workspace should be resized afterwards. It defaults to true, meaning workspaces will be resized by default as they currently are.

When `clearWorkspaceAndLoadFromXml` calls `clear`, it will now tell it *not* to resize the workspace, however.

### Reason for Changes

`clearWorkspaceAndLoadFromXml` removes all items from the workspace and then places new ones into it. When doing this, the scroll position should be preserved.

However, `clear` would cause the scroll position to be reset by resizing the workspace after clearing it. This manifested as the workspace being scrolled back up to the top left every time it was reloaded from XML, as changing the menu in the "() of ()" block did.

### DIFF
--- a/core/workspace_svg.js
+++ b/core/workspace_svg.js
@@ -2097,11 +2097,14 @@ Blockly.WorkspaceSvg.prototype.setToolboxRefreshEnabled = function(enabled) {
 
 /**
  * Dispose of all blocks in workspace, with an optimization to prevent resizes.
+ * @param {boolean} [resizeAfterwards=true] Whether the workspace should be resized after clearing.
  */
-Blockly.WorkspaceSvg.prototype.clear = function() {
-  this.setResizesEnabled(false);
+Blockly.WorkspaceSvg.prototype.clear = function(resizeAfterwards) {
+  // Default resizeAfterwards to true if not explicitly passed
+  resizeAfterwards = resizeAfterwards !== false;
+  if (resizeAfterwards) this.setResizesEnabled(false);
   Blockly.WorkspaceSvg.superClass_.clear.call(this);
-  this.setResizesEnabled(true);
+  if (resizeAfterwards) this.setResizesEnabled(true);
 };
 
 /**
--- a/core/xml.js
+++ b/core/xml.js
@@ -408,7 +408,9 @@ Blockly.Xml.textToDom = function(text) {
 Blockly.Xml.clearWorkspaceAndLoadFromXml = function(xml, workspace) {
   workspace.setResizesEnabled(false);
   workspace.setToolboxRefreshEnabled(false);
-  workspace.clear();
+  // If we clear the workspace and allow it to resize, the scroll position will be clamped around (0, 0)
+  // because the workspace will be empty when it is resized.
+  workspace.clear(false /* resizeAfterwards */);
   var blockIds = Blockly.Xml.domToWorkspace(xml, workspace);
   workspace.setResizesEnabled(true);
   workspace.setToolboxRefreshEnabled(true);


adroitwhiz:
@rachel-fenichel Looks like this may affect upstream Blockly as well?

rachel-fenichel:
Yes, but I don't want to add a new parameter to something that's so established as part of our core API.

adroitwhiz:
> I don't want to add a new parameter to something that's so established as part of our core API.

That's a good point. Would it perhaps be better to duplicate 
```js
Blockly.WorkspaceSvg.superClass_.clear.call(this);
```
inside `clearWorkspaceAndLoadFromXml`, or would you solve this another way entirely?

rachel-fenichel:
Taking a step back, why does changing a single block reload the entire workspace? The changes you're proposing seem to be workarounds for something that's already terribly non-performant and may be unnecessary.

adroitwhiz:
That's a good question!

The "(attribute) of (sprite)" block needs to update the "attribute" menu when you change the "sprite" menu, and apparently this is done by reloading the entire workspace.

There are three PRs across the VM, GUI, and this repo (https://github.com/LLK/scratch-vm/pull/2034, https://github.com/LLK/scratch-gui/pull/4613, https://github.com/LLK/scratch-blocks/pull/1908) that were supposed to *only* reload that one specific block, but they were closed for adding tech debt. The PR in scratch-gui (https://github.com/LLK/scratch-gui/pull/4613) goes into detail as to why.

In my mind, the workspace being reloaded completely when you change one menu is a problem, but workspace reloads not functioning correctly (for some definition of "correctly") is also a problem, and fixing both is a good idea.

[The GUI attempts to re-scroll the workspace after reloading it](https://github.com/LLK/scratch-gui/blob/dee7462c57e5fe2cd726ef12ef7f338f2ee01868/src/containers/blocks.jsx#L384-L390), but this fails due to what appears to be mysterious React interactions-- from what I can tell, Blockly emits an event which somehow causes the GUI to set `this.state.workspaceMetrics[this.props.vm.editingTarget.id]` in the same synchronous JS "tick", meaning that the workspace re-scroll doesn't work. If resetting the workspace scroll position on reload is intended behavior, I can fix the GUI re-scrolling code, but I wasn't entirely sure that's the case.

rachel-fenichel:
Clearing and repopulating the workspace just isn't intended to be a way to change one thing on the workspace while keeping everything else the same. So I don't think that it's not functioning correctly. As a result, fixing the GUI re-scrolling code is a better solution imo.

There's an open [issue](https://github.com/google/blockly/issues/1052) around tracking and firing events for viewport changes that I think is relevant, but it seems like a big problem to tackle at the wrong end. If you want to do it that way, you could implement those APIs and then grab the position before the clear, then scroll back to it after the clear. That's essentially what the GUI is doing, but using APIs instead of directly reaching in for workspace metrics properties.

adroitwhiz:
Alright, I've submitted a new fix over on the GUI (https://github.com/LLK/scratch-gui/pull/5774).

Another quick semi-related question: It seems like `setResizesEnabled` gets toggled a lot-- `clearWorkspaceAndLoadFromXml` sets it to `false` and then `true` after calling `clear` and `domToWorkspace`, but both of those functions themselves `setResizesEnabled` to `false` and then back to `true` once their work is done.

Would it make more sense to treat "resizes enabled" as a stack/counter? So e.g. setting it to `false` would increment a counter and setting it to `true` would decrement that counter, and resizes would only actually be re-enabled once the counter hits 0. I could see this being more trouble than it's worth but it may cut down on unnecessary resizes.

rachel-fenichel:
I agree, making it a counter would be a good idea. That would match what we do with events.  I've filed an issue: https://github.com/google/blockly/issues/3905

PR #2072, opened 2020-05-23
Change checkboxes_ from array to object map

### LINKED ISSUE #2071: Toggling checkboxes requires looping over many checkboxes
### Expected Behavior

Toggling a checkbox should be a constant-time operation

### Actual Behavior

Currently, toggling a checkbox [requires going over all checkboxes in the flyout until reaching one with the proper block ID](https://github.com/LLK/scratch-blocks/blob/20ae998b7ffeb82aa26df35087f27d49b74beb32/core/flyout_vertical.js#L667-L688).

This shouldn't cause much of a problem with toggling one checkbox at a time, but https://github.com/LLK/scratch-gui/pull/4147 loops over all monitors and sets their checkbox state, which will cause the number of array lookups to increase *exponentially* with the number of checkboxes.

adroitwhiz:
### Resolves

Resolves #2071

### Proposed Changes

This PR changes `flyoutVertical.checkboxes_` from an array to an object whose values are checkboxes and keys are the IDs of those checkboxes' corresponding blocks.

### Reason for Changes

Currently, the only API for setting a checkbox does so based on block ID, so it makes sense to index checkboxes based on block ID as well.

This should also be faster when looping over all checkboxes to update them, like in https://github.com/LLK/scratch-gui/pull/4147/.

### Test Coverage
Not sure if/where flyout tests exist

### DIFF
--- a/core/flyout_vertical.js
+++ b/core/flyout_vertical.js
@@ -59,13 +59,13 @@ Blockly.VerticalFlyout = function(workspaceOptions) {
   this.horizontalLayout_ = false;
 
   /**
-   * List of checkboxes next to variable blocks.
+   * Map of checkboxes that correspond to monitored blocks.
    * Each element is an object containing the SVG for the checkbox, a boolean
    * for its checked state, and the block the checkbox is associated with.
-   * @type {!Array.<!Object>}
+   * @type {!Object.<string, !Object>}
    * @private
    */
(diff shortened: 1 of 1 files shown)


fsih (approved):
LG, let's add this PR to the separate branch that we will test before merging with develop (along with https://github.com/LLK/scratch-blocks/pull/2072)

PR #2077, opened 2020-06-16
on Windows, use google-closure-compiler.cmd

### LINKED ISSUE #2060: WindowsError: [Error 2]  when trying to do npm install
### Expected Behavior

Forked and cloned this repo to my PC. I followed the guide in [this wiki](https://github.com/LLK/scratch-gui/wiki/Getting-Started#slightly-different-for-scratch-blocks). Expected to install 

### Actual Behavior

The following text appear in Git Bash when I typed in `npm i` in the directory of the clone
```bash
  File "build.py", line 574, in <module>
  File "C:\Python27\lib\subprocess.py", line 394, in __init__
    errread, errwrite)
  File "C:\Python27\lib\subprocess.py", line 644, in _execute_child
    startupinfo)
WindowsError: [Error 2] 
```
### Steps to Reproduce

clone the repo and follow the steps [here](https://github.com/LLK/scratch-gui/wiki/Getting-Started#slightly-different-for-scratch-blocks)
When you typed `npm i` in bash OR cmd, you should see the error above appear and the process fail.

Window 10, Python 2.7.17 64-bit

cwillisf:
### Resolves

Resolves #2060

### Proposed Changes

On Windows builds, when using the local compiler, run it as `google-closure-compiler.cmd` instead of `google-closure-compiler` with no extension.

### Reason for Changes

In Windows, trying to use `subprocess.Popen` to run `google-closure-compiler` with no extension doesn't work correctly. There are two ways to fix this: either use `shell=True` or add an extension. Unfortunately, using `shell=True` seems to break on Linux (and maybe macOS, but I haven't checked), so I went with adding an extension instead.

This fixes the specific complaint in #2060, which is a "file not found" error. Unfortunately this doesn't fix Windows builds: now some (but not all) of the calls to the Closure Compiler get a different error indicating that the command line is too long.

### DIFF
--- a/build.py
+++ b/build.py
@@ -51,7 +51,7 @@ CLOSURE_COMPILER = REMOTE_COMPILER
 CLOSURE_DIR_NPM = "node_modules"
 CLOSURE_ROOT_NPM = os.path.join("node_modules")
 CLOSURE_LIBRARY_NPM = "google-closure-library"
-CLOSURE_COMPILER_NPM = "google-closure-compiler"
+CLOSURE_COMPILER_NPM = ("google-closure-compiler.cmd" if os.name == "nt" else "google-closure-compiler")
 
 def import_path(fullpath):
   """Import a file with full path specification.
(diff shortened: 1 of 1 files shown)


picklesrus:
LGTM!

I can't really test it because I don't have windows... but I did take the python conditional and run it.  It resolves correctly based on what I set os.name to :)

PR #2090, opened 2020-07-16
Start a new event group when duplicating blocks

### LINKED ISSUE #2089: It takes two undo/redo steps to undo/redo the duplication of a block stack
### Expected Behavior

When you duplicate a block stack and then undo, you should only need to undo once to delete that block stack.

### Actual Behavior

The first time you undo after placing a duplicated block stack, it will move the duplicated stack directly on top of the original stack. The second undo actually deletes the duplicated stack.

Likewise for redoing: the first redo creates a duplicated block stack on top of the original stack, and the second redo moves it to the proper position.

### Steps to Reproduce

1. Right-click a block stack
2. Click "duplicate"
3. Place the new block stack somewhere (either attach it to another stack, or put it on the workspace)
4. Right-click the workspace and click "undo" (once)
5. Drag the stack that you duplicated, and observe that there are actually now two stacked copies of it
6. Repeat the process, but this time, click "undo" twice
7. Observe that the stack is now properly deleted

All

adroitwhiz:
### Resolves

Resolves #2089

### Proposed Changes

This PR starts a new event group inside `duplicateAndDragCallback` before creating the duplicated block stack.

### Reason for Changes

This will cause undo/redo to operate on both the creation and the placement of the duplicated stack in one step, rather than requiring the user to undo twice to undo both the movement of the duplicated stack and its creation.

### Test Coverage

Tested manually

### DIFF
--- a/core/scratch_blocks_utils.js
+++ b/core/scratch_blocks_utils.js
@@ -217,6 +217,14 @@ Blockly.scratchBlocksUtils.duplicateAndDragCallback = function(oldBlock, event) 
       } finally {
         Blockly.Events.enable();
       }
+
+      // Start a new event group for the block creation + drag, so that one undo
+      // step will undo both the creation of the duplicated stack and its
+      // placement. endBlockDrag will later end this event group.
+      if (isMouseEvent) {
+        Blockly.Events.setGroup(true);
+      }
+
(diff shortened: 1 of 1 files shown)


cwillisf:
This code will be replaced soon. Sorry we never merged this!

Issue #2092, opened 2020-07-20
"npm install" fails under nodeenv because "python" is python3


noamraph:
### Expected Behavior

Running "npm install" when using nodeenv should work.

### Actual Behavior

"npm install" fails. Reports that build.py only supports python2.

### Steps to Reproduce

```
cd /tmp
python3 -m venv venv
venv/bin/pip install nodeenv
venv/bin/nodeenv --python-virtualenv
source venv/bin/activate
git clone https://github.com/LLK/scratch-blocks.git
cd scratch-blocks/
npm install
```

It fails. I get:

```
> scratch-blocks@0.1.0 prepublish /tmp/scratch-blocks
> python build.py && webpack

Traceback (most recent call last):
  File "build.py", line 40, in <module>
    "You are using: " + sys.version)
Exception: Blockly build only compatible with Python 2.x.
You are using: 3.5.2 (default, Apr 16 2020, 17:47:17) 
```

### Operating System and Browser

Ubuntu 16.04

noamraph:
@BryceLTaylor why did you mark this as a question? I had to change the code in order to make scratch-blocks build. I also created pull request https://github.com/LLK/scratch-blocks/pull/2093 that fixes this.

cwillisf:
scratch-blocks v2 is a complete rewrite built on top of upstream [Blockly](https://github.com/RaspberryPiFoundation/blockly/). The old build system (Google Closure compiler, `python build.py`, the `core/` and `shim/` directories, etc.) no longer exists. This issue is no longer applicable. Thank you for reporting it!

If you believe this issue is still relevant to scratch-blocks v2, please let us know and we can reopen it.

PR #2093, opened 2020-07-20
Use "python2 build.py" instead of "python build.py"

noamraph:
I tried "npm install" and it failed. It turned out that it was because I was using nodeenv, and "python" was set to python3. Replacing "python build.py" with "python2 build.py" solved the problem. I don't think it should cause any trouble.

### Resolves

https://github.com/LLK/scratch-blocks/issues/2092

### Proposed Changes

Instead of running "python build.py", run "python2 build.py", so it will always be python2, instead of whatever "python" is.

### Reason for Changes

build.py only supports python2. When "python" is python3, running "python build.py" fails, while running "python2 build.py" succeeds.

### Test Coverage

None.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "browser": "./shim/vertical.js",
   "scripts": {
     "deploy": "rimraf gh-pages/closure-library/scripts/ci/CloseAdobeDialog.exe && gh-pages -t -d gh-pages -m \"Build for $(git log --pretty=format:%H -n1)\"",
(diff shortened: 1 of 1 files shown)


apple502j:
Reviewers: Please review #1964 which solves the problem way better than this

ericrosenbaum:
I'm going to close this, and we'll try to get the review for the earlier PR scheduled: https://github.com/LLK/scratch-blocks/pull/1964

PR #2103, opened 2020-09-03
Update scratch_msgs.js

caoyongxu:
Fix an translation error.
The meaning of the incorrect translation: "Insert the data BEFORE the specific index location", instead of "Insert the data AT the specific index location".     "前" is "before" in chinese.

### Proposed Changes
Just remove the chinese char "前" in the translated text.


apple502j:
scratch_msgs.js is a generated file. You should translate at Transifex.

caoyongxu:
> scratch_msgs.js is a generated file. You should translate at Transifex.

thanks, would u pls let me know which file I should change ? I tried to find it but didn't  find such a file.

apple502j:
Scratch Editor -> blocks

chrisgarrity:
@caoyongxu just confirming what @apple502j said - this file is generated from translations in transifex. See https://github.com/LLK/scratch-l10n/wiki/Guide-for-Scratch-Translators for more details on helping with translations.

caoyongxu:
@chrisgarrity, @apple502j thanks.

PR #2523, opened 2021-06-28
chore(renovate): automerge more, enable semantic commits

cwillisf:
### Proposed Changes

With this change, we'll automerge minor and patch updates from most dependencies. The only exception is dependencies with a pre-1.0.0 version number: automerge is disabled for those, since per the SemVer spec they may introduce breaking changes at any time.

Automerge should also be disabled for any dependencies which don't follow SemVer, but this repository currently doesn't have any.

Additional changes:
- rename "renovate.json" to "renovate.json5" so editors allow comments
- add ":semanticCommits" so Renovate will use semantic commits (default is "chore" for devDependencies and "fix" for dependencies)
- add ":separateMultipleMajor" so Renovate will open separate PRs if there are multiple available major updates.
- set the timezone to "America/New_York" to match Scratch time

### Reason for Changes

I think the only way we'll keep up with dependency updates is if we automerge when we can. Doing so for minor & patch updates, only when they pass tests, seems like a good compromise between safety (never automerge) and keeping up to date (always automerge). Over time we may want to add more tests to make these automerges even safer.

An earlier version of this PR set `automergeType=branch` to further reduce the notifications we'd receive due to dependency updates by skipping PRs for automerged updates. I've removed this after some discussion with the team. Skipping PRs would make these updates unlike other changes: for example, while the commit would show up in the commit history, nothing would come up in a search through GitHub's list of closed PRs. Also, Renovate PRs contain valuable information which we wouldn't have access to if we set `automergeType=branch`.

The other changes are nice-to-haves or experiments to see if we like Renovate's approach. For example: do we like Renovate's default method of determining when to use "chore" or "fix" for semantic commit prefixes? Let's find out!


rschamp (approved):
I'm personally OK with auto-merging non-breaking changes to dependencies. Just putting this review in as a 👍  from me.

cwillisf:
We should not merge this until after #2526 is merged.

Issue #1154, opened 2017-10-19
😭 Failing builds get published 😭


paulkaplan:
Our most recent merge into develop was broken on travis, but didn't actually fail the tests so it was declared "ok" and deployed to npm. That broke the GUI build, and is generally terrible.

Here is the log where it failed and didn't fail. https://travis-ci.org/LLK/scratch-blocks/jobs/289755894#L501 

If building the compressed library fails, travis should fail.

PR #2526, opened 2021-06-30
fix(test): fix unit tests

### LINKED ISSUE #2527: unit tests not running on CI due to Chrome / ChromeDriver problem
### Expected Behavior

Unit tests should run on CI using headless Chrome through Selenium's ChromeDriver.

### Actual Behavior

Unit tests fail to run on CI and instead print an error similar to:

```
WebDriverError: unknown error: Chrome failed to start: exited abnormally.
  (unknown error: DevToolsActivePort file doesn't exist)
  (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
    at Object.throwDecodedError (/home/travis/build/LLK/scratch-blocks/node_modules/selenium-webdriver/lib/error.js:517:15)
    at parseHttpResponse (/home/travis/build/LLK/scratch-blocks/node_modules/selenium-webdriver/lib/http.js:642:13)
    at Executor.execute (/home/travis/build/LLK/scratch-blocks/node_modules/selenium-webdriver/lib/http.js:568:28)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  remoteStacktrace: '#0 0x555a41cdd5b9 <unknown>\n'
}
```

Prior to #2526 the error will be prefixed with `UnhandledPromiseRejectionWarning:` and will not be treated by CI as a test failure.

### Steps to Reproduce

See recent CI build logs

### LINKED ISSUE #2528: failure to launch tests is not treated as a test failure
### Expected Behavior

If the unit test runner fails to run the unit tests for any reason, `npm test` should exit with an error code.

### Actual Behavior

If the unit test runner fails to run the unit tests because of a Selenium / ChromeDriver error, reported through promise rejection, that rejection is not caught. In this situation the test runner, and therefore `npm test`, both exit with code 0 (success).

### Steps to Reproduce

See recent CI build logs, or attempt to run unit tests locally in such a way that they will fail, such as if Chrome is not installed.

cwillisf:
### Resolves

Resolves #2528
Resolves #2527
Might be related to #1154

### Proposed Changes

If any of the calls to the Selenium web driver's `browser` object fail with a promise rejection, the test runner now reports the error and exits with code 1. To make this easier to read, I've converted the body of the test into an `async` function which uses `await` when making these calls.

UPDATE: I've added changes which actually make the tests pass, including:
- Update the `chromedriver` dependency
- Disable the Chrome sandbox on Linux CI
- Enable Chrome's headless mode on CI regardless of platform
- Quit the browser session after tests

### Reason for Changes

The previous formulation of the test runner was not catching promise rejections coming from the webdriver / browser. Node would print warnings to the console, but the test runner would exit with code 0 and CI would treat this as a passing build. Now the rejection is caught and reported, and the test runner exits with code 1.

Existing functionality to look for failed test results has not changed.

UPDATE:
- Updating the `chromedriver` dependency was needed to get the tests running at all. In fact, this was enough to get them working for me locally.
- I disabled the Chrome sandbox on Linux CI to get the tests working on Travis. See [Travis docs](https://docs.travis-ci.com/user/chrome#sandboxing)
- Chrome's headless mode is kind of a nice-to-have in general, but may be necessary on Travis since XVFB doesn't seem to be working.
- If we don't quit the browser session after tests, Chrome processes are left running each time. This isn't a big deal when testing on CI since the VM is short-lived anyway, but it's annoying when running tests locally.

### Test Coverage

The first commit in this PR made it so that test failures would be noticed on CI, so the fact that it failed with ❌ is a success. From there on, getting the existing tests to pass was the goal.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -489,28 +489,28 @@
       }
     },
     "@nodelib/fs.scandir": {
-      "version": "2.1.3",
-      "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.3.tgz",
-      "integrity": "sha512-eGmwYQn3gxo4r7jdQnkrrN6bY478C3P+a/y72IJukF8LjB6ZHeB3c+Ehacj3sYeSmUXGlnA67/PmbM9CVwL7Dw==",
+      "version": "2.1.5",
+      "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.5.tgz",
+      "integrity": "sha512-vq24Bq3ym5HEQm2NKCr3yXDwjc7vTsEThRDnkp2DK9p1uqLR+DHurm/NOTo0KG7HYHU7eppKZj3MyqYuMBf62g==",
       "dev": true,
       "requires": {
-        "@nodelib/fs.stat": "2.0.3",
+        "@nodelib/fs.stat": "2.0.5",
         "run-parallel": "^1.1.9"
       }
     },
     "@nodelib/fs.stat": {
-      "version": "2.0.3",
-      "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-2.0.3.tgz",
-      "integrity": "sha512-bQBFruR2TAwoevBEd/NWMoAAtNGzTRgdrqnYCc7dhzfoNvqPzLyqlEQnzZ3kVnNrSp25iyxE00/3h2fqGAGArA==",
+      "version": "2.0.5",
+      "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-2.0.5.tgz",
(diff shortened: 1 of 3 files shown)


rschamp (commented):
So, we should approve this because it's failing the tests? 🙃  L❌TM
Thanks for catching this, it does seem related to the problem in #1154.  If we merge this, then we won't be able to release scratch-blocks until we fix the tests... Maybe we should fix them in a separate branch based on this one?

cwillisf:
Yeah, I was thinking about this last night and came to the same conclusion: as bad as it is to ignore passing tests, it's probably worse to stop publishing altogether _in this particular case_ since the codebase is stable and it's just localization that's changing regularly. I'll try to figure out what's going wrong -- maybe just a Selenium update.

rschamp (approved):
Nice work!

Issue #2613, opened 2021-07-13
webpack output is not used for "browser"


cwillisf:
### Expected Behavior

We should use the build output we generate; we should not generate unused build output.

### Actual Behavior

We currently use `webpack` to build two targets: a set of "commonjs2" output under `dist/` and a set of UMD output under `dist/web/`. Our `package.json` lists:

```json
  "main": "./dist/vertical.js",
  "browser": "./shim/vertical.js",
```

This means we're instructing "browser"-style consumers of this module to use the build input, not the build output. The `webpack` output in `dist/web/` is not being used.

This also means that `exports-loader` and `imports-loader` need to be listed in our `dependencies` so that `scratch-gui` has access to those loaders at `scratch-gui`'s build time.

cwillisf:
scratch-blocks v2 is a complete rewrite built on top of upstream [Blockly](https://github.com/RaspberryPiFoundation/blockly/). The old build system (Google Closure compiler, `python build.py`, the `core/` and `shim/` directories, etc.) no longer exists. This issue is no longer applicable. Thank you for reporting it!

PR #2614, opened 2021-07-13
move {ex,im}ports-loader back to dependencies

cwillisf:
### Proposed Changes

Move `exports-loader` and `imports-loader` from `devDependencies` back into `dependencies`.

### Reason for Changes

This fixes the `scratch-gui` build. Since we specify the unprocessed `shim/vertical.js` rather than the build output in `dist/web/` as the "browser" field in `package.json`, consumers of this module in a browser context (such as `scratch-gui`) will need these loaders at runtime or build time.

See #2613 for more info on this.


BryceLTaylor (approved):
Looks good.

PR #2872, opened 2022-10-06
CE-180: Migrate to Transifex v3 API

hyperobject:
### Resolves

- CE-180
- https://github.com/LLK/scratch-l10n/issues/171

### Proposed Changes

This PR pulls in the `txPull` helper function from scratch-l10n and uses it to pull strings from transifex instead of the old API. (ideally there wouldn't be code duplication, but I'm not sure if we want to export the helper functions from scratch-l10n or not. maybe good to discuss!)

### Reason for Changes

The Transifex API we are currently using will be shut down on November 30th, 2022 and needs to be migrated off of before then.


i18n/sync_tx_translations.js
```diff
@@ -23,19 +23,16 @@ if (!process.env.TX_TOKEN || process.argv.length !== 2) {
...
-const transifex = require('transifex');
 const locales = require('scratch-l10n').default;
+const txPull = require('./transifex.js');
 
 // Globals
 const PATH_OUTPUT = path.resolve(__dirname, '../msg');
 const PROJECT = 'scratch-editor'
 const RESOURCE = 'blocks';
-const MODE = {mode: 'reviewed'};
+const MODE = 'reviewed';
```
cwillisf, line 33: 😎

i18n/sync_tx_translations.js
```diff
@@ -23,19 +23,16 @@ if (!process.env.TX_TOKEN || process.argv.length !== 2) {
 const fs = require('fs');
 const path = require('path');
 const assert = require('assert');
-const transifex = require('transifex');
 const locales = require('scratch-l10n').default;
+const txPull = require('./transifex.js');
```
cwillisf: Since `scratch-blocks` already depends on `scratch-l10n` it seems like, once LLK/scratch-l10n#175 is merged, this could be:
```suggestion
const {txPull} = require('scratch-l10n/lib/transifex.js');
```
...or you could re-export it from `scratch-l10n/src/index.js` and potentially merge this proposed line with the `locales` line above.

Anyway, either option would let you remove the duplicated code in this PR's `i18n/transifex.js`, and it looks like that might make it unnecessary to depend on the Transifex API or `download` packages directly as well.

cwillisf (changes requested):
I think the `transifex.js` part can be simplified, but the `sync_tx_translations.js` part looks good! I like how much simpler it is now that we can `await` ;)

i18n/sync_tx_translations.js
hyperobject: oh awesome! wasn't sure if there was a good way to import this stuff directly from scratch-l10n, I'd much rather do that. I'll change that once the l10n PR gets merged!

hyperobject:
@cwillisf looks like the `txPull` import doesn't depend on https://github.com/LLK/scratch-l10n/pull/175, so this should be good to go with those changes now!

cwillisf (approved):
Since this end result doesn't depend on any `package.json` changes, I'd recommend trying to make it so this PR doesn't modify `package-lock.json` either. Judging by the differences I see in there, though, it's not that big of a deal. Sorry for the difficulties caused by introducing `package-lock.json` with this PR hanging! :/

PR #2891, opened 2022-11-02
ci: replace Travis with Circle

### LINKED ISSUE #1154: 😭 Failing builds get published 😭

### LINKED ISSUE #1422: Update travis.yml to use node 8
We are trying to make all of our javascript code run using node 8. Make sure node 8 is one of the travis run options.

https://github.com/LLK/scratch-blocks/blob/develop/.travis.yml#L3-L4

### LINKED ISSUE #2042: Handle translation build with no updates
### Expected Behavior

The build for the automated update of translations should not fail when there are no updates. It should skip the commit step if there is nothing to commit.

### Actual Behavior

The build gets an error because it tries to commit when there are no changes to commit.

### Steps to Reproduce

https://travis-ci.org/LLK/scratch-blocks/jobs/654904556

cwillisf:
### Resolves

- Resolves #2042 
- Closes #1422
- I believe this also resolves #1154

### Proposed Changes

Migrate from TravisCI to CircleCI, keeping functionality as similar as reasonable.

Specific changes:
* add CircleCI config
* move contents of `i18n/sync_translations.sh` into the CircleCI config and adjust as needed
* remove `tests/scripts/setup_linux_env.sh`
* update various support files, like `.npmignore` and `README.md`, to reflect the migration

### Reason for Changes

As an organization, we're moving from Travis to Circle.

The contents of `i18n/sync_translations.sh` were specific to CircleCI so I felt it would make more sense to move them into that context. Another good option would be to move `i18n/sync_translations.sh` to `.circleci/sync_translations.sh`. Let me know if you have opinions on that.

Fun fact: installing Python 2 only takes about 10 seconds, and installing Chrome only takes about 20. Probably still slower than pulling extra layers for a custom Docker image, but not terrible :)

### Test Coverage

I enabled CircleCI and it has been building commits on this branch.


.circleci/config.yml
```diff
@@ -0,0 +1,133 @@
...
+jobs:
+  default:
+    <<: *defaults
+    environment:
+      # DETECT_CHROMEDRIVER_VERSION only works if Chrome is installed before installing NPM packages
+      DETECT_CHROMEDRIVER_VERSION: "true"
+    steps:
+      - checkout
+      - install_python2
+      - browser-tools/install-chrome
```
meyerhot95, line 83: Do we want to specify a node_version here to match 16.18 ? 

ie. 

```
- node/install:
          install-yarn: true
          node-version: '16.13'
   ```

meyerhot95: I am jw if there will be a mismatch of node versions used to install packages vs test.

cwillisf: The `node/install-packages` command only installs dependencies, like `node_modules` stuff. It'll just use the Node version that already exists in the image (the `cimg/node:16.18-browsers` from `defaults` above, like you said) so everything should stay consistent.

The `node/install` command could introduce inconsistency like you're suggesting, especially if we update one version number (like in `defaults`) and forget to update the other (here), so I'd prefer to avoid that

meyerhot95 (approved):
LGTM

PR #2962, opened 2023-02-15
feat: block text color with stylesheet reloading

aoneill01:
_The color theme epic will span multiple pull requests. This small piece will not be merged directly to `develop`. Instead all of the color theme work will be collected in a `feature/color-contast` branch until the epic is complete._

### Resolves

- Resolves [ENA-240](https://scratchfoundation.atlassian.net/browse/ENA-240)

### Proposed Changes

- Rename the `text` color to `textFieldText`
- Reuse `text` as the color of the text on the blocks
- Reload the css if it has already been injected

### Reason for Changes

Rather than always being white, the block text color should be configurable to support theming. 

The Scratch editor wants to apply themes based on user selection. The css needs to be reloaded when the user changes the color theme. 

### Test Coverage

Manual testing using LLK/scratch-gui#8755.

[ENA-240]: https://scratchfoundation.atlassian.net/browse/ENA-240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/css.js
+++ b/core/css.js
@@ -76,9 +76,9 @@ Blockly.Css.mediaPath_ = '';
  * @param {string} pathToMedia Path from page to the Blockly media directory.
  */
 Blockly.Css.inject = function(hasCss, pathToMedia) {
-  // Only inject the CSS once.
+  // Clear the CSS if it has already been injected.
   if (Blockly.Css.styleSheet_) {
-    return;
+    document.head.removeChild(Blockly.Css.styleSheet_.ownerNode);
   }
   // Placeholder for cursor rule.  Must be first rule (index 0).
   var text = '.blocklyDraggable {}\n';
@@ -459,7 +459,7 @@ Blockly.Css.CONTENT = [
   '}',
 
   '.blocklyText {',
-    'fill: #fff;',
+    'fill: $colour_blockText;',
     'font-family: "Helvetica Neue", Helvetica, sans-serif;',
     'font-size: 12pt;',
     'font-weight: 500;',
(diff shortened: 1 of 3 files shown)


core/css.js
```diff
@@ -93,14 +93,15 @@ Blockly.Css.inject = function(hasCss, pathToMedia) {
...
-    }
+  // Process longer colour properties first to handle common prefixes.
+  var compareByLength = function(a, b) { return b.length - a.length; };
+  var colourProperties = Object.keys(Blockly.Colours).sort(compareByLength);
+  for (var i = 0, colourProperty; colourProperty = colourProperties[i]; i++) {
+    // Replace all
+    text = text.replace(
+      new RegExp('\\$colour\\_' + colourProperty, 'g'),
+      Blockly.Colours[colourProperty]
+    );
```
aoneill01, line 104: Not really the style I would normally use (`var`, not using arrow functions or `for ...of` loops), but this was done intentionally to match the style of this library.

For more context, the sorting was done because there are both `textField` and `textFieldText` properties. If `textField` was processed first, it would incorrectly replace lines with `$colour_textFieldText`. In fact, I believe this was already an issue with `scrollbar` and `scrollbarHover` that just hadn't been caught. I can see in production that the fill for the hovered scrollbar is evaluating to the invalid `#CECDCEHover`.

delasare (approved):

PR #2963, opened 2023-02-16
chore: upgrade package-lock.json to lockfileVersion 2

cwillisf:
### Proposed Changes

Upgrade `package-lock.json` to `"lockfileVersion": 2` with no dependency changes.

### Reason for Changes

We now build this repository using Node 16, which uses `"lockfileVersion": 2` by default. Upgrading the file resolves a warning (shown below) and means less risk of a merge conflict for upcoming dependency changes.

```
npm WARN old lockfile 
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile 
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile 
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,9895 @@
 {
   "name": "scratch-blocks",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "scratch-blocks",
+      "version": "0.1.0",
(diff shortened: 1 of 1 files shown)


aoneill01 (approved):
There had been some discussion on waiting to merge until after the color theme work is done to avoid conflicts. I am fine either way. It seems unlikely that I will touch this file as part of the color theming, and if I do, it won't be the end of the world to resolve the differences.

PR #2976, opened 2023-03-02
feat: add quaternary color for selected dropdowns

aoneill01:
_Part of the color contrast epic. During development this will live on the `feature/color-contrast` branch._

_If you want to test this locally, there are also some changes in `scratch-gui` on the `feature/color-contrast` branch, commit 8533704. This would involve building `scratch-blocks` and using `npm link scratch-blocks` on `scratch-gui`._

### Resolves

- Resolves [ENA-244](https://scratchfoundation.atlassian.net/browse/ENA-244)

### Proposed Changes

- Add a quaternary color to blocks
- Use the quaternary color for styling selected dropdown menus (`core/field_dropdown.js`)
- Add a `menuHover` color property for hovered menu items

### Reason for Changes

The design plans for the high contrast mode have a lighter color for selected dropdown menus. This is a fourth color that does not match the primary, secondary, or tertiary colors. Perhaps this fourth color will be used for other theming in the future.

Screenshot 2023-03-03 at 11 22 21 AM

### Test Coverage

This is mostly a styling change which the existing automated tests are not testing. Used `scratch-gui` to see that the styles look correct for the high contrast theme.

- Added test to verify `jsonInit` sets all 4 colors and makes the source block's colors available to the dropdown.

[ENA-244]: https://scratchfoundation.atlassian.net/browse/ENA-244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/blocks_vertical/vertical_extensions.js
+++ b/blocks_vertical/vertical_extensions.js
@@ -35,25 +35,26 @@ goog.require('Blockly.constants');
 
 /**
  * Helper function that generates an extension based on a category name.
- * The generated function will set primary, secondary, and tertiary colours
- * based on the category name.
+ * The generated function will set primary, secondary, tertiary, and quaternary
+ * colours based on the category name.
  * @param {String} category The name of the category to set colours for.
  * @return {function} An extension function that sets colours based on the given
  *     category.
  */
 Blockly.ScratchBlocks.VerticalExtensions.colourHelper = function(category) {
   var colours = Blockly.Colours[category];
-  if (!(colours && colours.primary && colours.secondary && colours.tertiary)) {
+  if (!(colours && colours.primary && colours.secondary && colours.tertiary &&
+    colours.quaternary)) {
     throw new Error('Could not find colours for category "' + category + '"');
   }
   /**
-   * Set the primary, secondary, and tertiary colours on this block for the
-   * given category.
+   * Set the primary, secondary, tertiary, and quaternary colours on this block for
+   * the given category.
    * @this {Blockly.Block}
    */
   return function() {
     this.setColourFromRawValues_(colours.primary, colours.secondary,
-        colours.tertiary);
+        colours.tertiary, colours.quaternary);
   };
 };
 
--- a/blocks_vertical/vertical_extensions.js
+++ b/blocks_vertical/vertical_extensions.js
@@ -63,7 +63,8 @@ Blockly.ScratchBlocks.VerticalExtensions.colourHelper = function(category) {
  */
 Blockly.ScratchBlocks.VerticalExtensions.COLOUR_TEXTFIELD = function() {
   this.setColourFromRawValues_(Blockly.Colours.textField,
-      Blockly.Colours.textField, Blockly.Colours.textField);
+      Blockly.Colours.textField, Blockly.Colours.textField,
+      Blockly.Colours.textField);
 };
 
 /**
--- a/core/block.js
+++ b/core/block.js
@@ -239,6 +239,13 @@ Blockly.Block.prototype.colourSecondary_ = '#FF0000';
 Blockly.Block.prototype.colourTertiary_ = '#FF0000';
 
 /**
+ * Quaternary colour of the block in '#RRGGBB' format.
+ * @type {string}
+ * @private
+ */
+Blockly.Block.prototype.colourQuaternary = '#FF0000';
+
+/**
  * Fill colour used to override default shadow colour behaviour.
  * @type {string}
  * @private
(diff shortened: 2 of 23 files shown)


blocks_vertical/vertical_extensions.js
```diff
@@ -35,25 +35,26 @@ goog.require('Blockly.constants');
...
+ * colours based on the category name.
  * @param {String} category The name of the category to set colours for.
  * @return {function} An extension function that sets colours based on the given
  *     category.
  */
 Blockly.ScratchBlocks.VerticalExtensions.colourHelper = function(category) {
   var colours = Blockly.Colours[category];
-  if (!(colours && colours.primary && colours.secondary && colours.tertiary)) {
+  if (!(colours && colours.primary && colours.secondary && colours.tertiary &&
+    colours.quaternary)) {
```
cwillisf, line 47: Just double-checking: will it be guaranteed, going forward, that there's always a quaternary color? That is, is it reasonable for this to throw if there's not a quaternary color?

core/block.js
```diff
@@ -864,6 +881,11 @@ Blockly.Block.prototype.setColour = function(colour, colourSecondary, colourTert
     this.colourTertiary_ = goog.color.rgbArrayToHex(
         goog.color.darken(goog.color.hexToRgb(this.colour_), 0.2));
   }
+  if (colourQuaternary !== undefined) {
+    this.colourQuaternary_ = this.makeColour_(colourQuaternary);
+  } else {
+    this.colourQuaternary_ = this.colourTertiary_;
+  }
```
cwillisf, line 888: What's the relationship between this `else` clause and the `throw` at the start of `colourHelper`?

core/field_colour_slider.js
```diff
@@ -112,9 +112,10 @@ Blockly.FieldColourSlider.prototype.setValue = function(colour) {
   }
   this.colour_ = colour;
   if (this.sourceBlock_) {
-    // Set the primary, secondary and tertiary colour to this value.
+    // Set the colours to this value.
```
cwillisf, line 115: 😆

cwillisf (approved):
Looks good!
I left a couple questions inline, mostly for my understanding, but it's possible there's a corner case that's not covered or a tiny bit of redundant code. I'll admit I didn't trace through all the plumbing 😅

blocks_vertical/vertical_extensions.js
aoneill01: I did try to find every place that we were passing the three colors and added the fourth.

However, it looks like this is only called from the color extension though. So for example, the json for a block could contain:

```js
  "extensions": ["colours_looks"]
```

 ...and it would automatically apply the colors for the looks blocks. `core/colours.js` has been updated with default quaternaries for all of the categories used by this extension, so it should never be missing.

core/block.js
aoneill01: See my reply above about `colourHelper`. It would be called before this function as an extension to simplify defining the colors for block json.

`setColour` is used in many different places. I think I caught all of the places that were using 3 values and added the fourth. I do see at least one place (insertion marker in block.js) that was only using 1 value, so this conditional is used.

towerofnix:
Something interesting to note about this design is that previously, a selected dropdown's fill matched the stroke color of the block border (i.e. the tertiary color), which is also used for the input border:

Dropdown with shadow block (input), unselected vs selected<br>Dropdown without shadow block (field), unselected vs selected

In this implementation, since the quaternary color is introduced, that fill is different from the block border - and thus distinct from the border of the input as well! (Can't run locally ATM but see [main demo screenshot](https://user-images.githubusercontent.com/4893763/222773199-a2733f37-61af-4a36-8cf6-e3b89d119399.png).)

Folk viewing the repo from outside (👋) don't have access to the design documents, but it wouldn't surprise me if this is an intentional change. It *almost* reflects focus rings in standard browser/UI design. (It's not quite the same, because the stroke itself doesn't actually become bolder or darker when selected, but it does come close by increasing the contrast between input fill and stroke.)

Just thought I'd point out the parallel with long-standing accessible UI design outside of Scratch! The color contrast epic looks quite epic ^_^

blocks_vertical/vertical_extensions.js
```diff
@@ -35,25 +35,26 @@ goog.require('Blockly.constants');
...
-   * Set the primary, secondary, and tertiary colours on this block for the
-   * given category.
+   * Set the primary, secondary, tertiary, and quaternary colours on this block for
+   * the given category.
    * @this {Blockly.Block}
    */
   return function() {
     this.setColourFromRawValues_(colours.primary, colours.secondary,
-        colours.tertiary);
+        colours.tertiary, colours.quaternary);
```
towerofnix, line 57: The `COLOUR_TEXTFIELD` extension just below here doesn't set `colours.quaternary` - that's the only missed use of `setColourFromRawValues_` though!

```js
// Text fields transcend categories.

/**
 * Extension to set the colours of a text field, which are all the same.
 */
```

aoneill01: Good catch! I will update that line.

aoneill01:
Thanks for pointing out this color not matching the border color @towerofnix. This design is based off of the work in https://lab.scratch.mit.edu/contrast-blocks/. I believe it is intentional, but I will double-check with the designers next time I am talking to them.

PR #3084, opened 2023-08-08
REQ-71: add Hindi, update translations

hyperobject:
### Resolves
- REQ-71

### Proposed Changes
- Updates scratch-l10n dependency and pulls new translations from transifex to add Hindi language support

See also https://github.com/scratchfoundation/scratch-l10n/pull/182.


msg/scratch_msgs.js
```diff
@@ -21254,7 +21541,7 @@ Blockly.ScratchMsgs.locales["vi"] =
     "CONTROL_STOP": "dừng lại",
     "CONTROL_STOP_ALL": "tất cả",
     "CONTROL_STOP_THIS": "kịch bản này",
-    "CONTROL_STOP_OTHER": "các kịch bản khác cho nhân vật",
+    "CONTROL_STOP_OTHER": "other scripts in sprite",
```
cwillisf: Looks like this one lost its translation... do you know what's up with that?

cwillisf (approved):
Looks mostly reasonable! One question about one of the Vietnamese messages. If that's as it should be, then I think this PR is good :)

~I'm also poking at CI to see if I can resolve the build failure, but I don't think that's related to your changes.~
CircleCI just declared a major outage related to this, so I guess I'll let them handle it :)

msg/scratch_msgs.js
hyperobject: I'll take a look in transifex!

hyperobject: looks like that string got unreviewed in transifex, causing it to revert back to english. backed out that one line change, everything else should be okay now!

PR #3351, opened 2025-05-31
fix: show updated field value after editing in horizontal blocks

asdsteven:
Problem:
In horizontal blocks, let's say you created a control_repeat block, if you edit the number of iteration from 4 to 5, after pressing Enter, the number is still shown as 4.

Cause:
field.js setText function would call forceRerender.  Currently, forceRerender would set this.size_.width to 0 to force a render_() call later.  However, this would work only if this.getSize() is indeed called later to trigger this.render_(), which is the case for vertical blocks, but not for horizontal blocks, causing the updated value not being shown.

Fix:
Setting this.size_.width to 0 was a silly trick.  Since our purpose is to force render_() anyway, we could have simply called render_() directly.  One may argue that, render_() should not be called immediately; some setup from sourceBlock_.render() needs to be done before this.render_() maybe called.  Well, this is not true. this.render_() does not depend on sourceBlock_, or anything.  We can just call it immediately.


asdsteven:

PR #3355, opened 2025-06-11
feat: [UEPR-268] change canvas svg to rgenerate when containing container resizes

MiroslavDionisiev:
### Resolves

[UEPR-268](https://scratchfoundation.atlassian.net/browse/UEPR-268)

### Proposed Changes
Currently the canvas svg regenerates when window resizes. However new features require to resize the editor without changing the window size. This cause the canvas to be bigger or smaller than the `div` that contains it. This changes make it so the canvas regenerates when the containing `div` changes size.

[UEPR-268]: https://scratchfoundation.atlassian.net/browse/UEPR-268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/inject.js
+++ b/core/inject.js
@@ -368,6 +368,7 @@ Blockly.init_ = function(mainWorkspace, container) {
   });
 
   resizeObserver.observe(container);
+  mainWorkspace.setResizeObserver(resizeObserver);
 
   Blockly.inject.bindDocumentEvents_();
 
--- a/core/inject.js
+++ b/core/inject.js
@@ -71,7 +71,7 @@ Blockly.inject = function(container, opt_options) {
 
   var workspace = Blockly.createMainWorkspace_(svg, options, blockDragSurface,
       workspaceDragSurface);
-  Blockly.init_(workspace);
+  Blockly.init_(workspace, subContainer);
   Blockly.mainWorkspace = workspace;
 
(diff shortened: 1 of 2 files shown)


core/inject.js
```diff
@@ -361,13 +362,12 @@ Blockly.init_ = function(mainWorkspace) {
         }
       });
 
-  var workspaceResizeHandler = Blockly.bindEventWithChecks_(window, 'resize',
-      null,
-      function() {
-        Blockly.hideChaffOnResize(true);
-        Blockly.svgResize(mainWorkspace);
-      });
-  mainWorkspace.setResizeHandlerWrapper(workspaceResizeHandler);
```
cwillisf, line 370: The old code kept track of this handler and would remove it in the workspace's `dispose()` function. Do we need to do something equivalent in the new code?

cwillisf:
Also, the test failure appears to be due to the browser failing to load the JS file(s). It seems like `--allow-file-access-from-files` would be the right solution, but that isn't working for me. Any thoughts?

core/inject.js
MiroslavDionisiev: You are right. I added logic for disconnecting the observer

PR #3388, opened 2025-09-12
remove malicious package versions by reverting 5b6ca5b

derpygamer2142:
This reverts commit 5b6ca5b99352e78c7475dc07ba8d346c7bf635e0.

### Resolves

Trying to npm install will error because of the invalid package versions, which were removed due to having malicious payloads.
https://github.com/chalk/chalk/issues/656

### Proposed Changes

Reverts the commit that uses those versions

### Reason for Changes

Malicious code is bad and cringe and you shouldn't use it. Also I wanted to fork this but when I tried to install it didn't work(very sad).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -711,9 +711,9 @@
       }
     },
     "node_modules/@semantic-release/commit-analyzer/node_modules/debug": {
-      "version": "4.4.2",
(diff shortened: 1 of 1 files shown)


derpygamer2142:

PR #3431, opened 2026-02-12
power operator block `a^b`

AeroKoder:
### Resolves

https://github.com/scratchfoundation/scratch-editor/issues/426

### Proposed Changes

Add a new block to the Operator category to raise a given number to a power of another given number.

### Reason for Changes

This is a highly requested block by many Scratchers and Scratch Team members (I have with some [[1](https://scratch.mit.edu/users/algorithmar/#comments-402360195),[2](https://scratch.mit.edu/users/glowinday/#comments-402404181)], and they agree that it should exist in Scratch). There also aren't any accurate workarounds, besides ones that are crazy inefficient, making this block even more important to have.

It has been discussed in multiple topics and issues, and they always end up with tons of upvotes and supports.

The one thing I would like to discuss is whether, zero to the power of zero, should return `1` or `NaN`. Currently, it returns `1`, as that is what the native JavaScript operator returns. This might confuse younger Scratchers, but then again, I don't know if `NaN` is such a good idea either, since it acts as zero in scratch, making some math equations with the power operator return an incorrect result. Discuss this in the [scratch-editor PR](https://github.com/scratchfoundation/scratch-editor/pull/444)

### Test Coverage

Tests are in the [scratch-editor PR](https://github.com/scratchfoundation/scratch-editor/pull/444)


AeroKoder:

redspacecat:
@AeroKoder there are now "conflicts" from these changes, possibly due to the blockly updates.

AeroKoder:
> @AeroKoder there are now "conflicts" from these changes, possibly due to the blockly updates.

Thanks for the notice. I’m out all weekend, will be able to resolve them Monday.

PR #3433, opened 2026-02-26
fix: allow Scratch-style zoom controls as separate SVGs

cwillisf:
### Resolves

- Resolves gonfunko/scratch-blocks#94

### Proposed Changes

Implements a custom replacement for the Blockly zoom controls.

### Reason for Changes

This makes the zoom in/out/reset buttons look the way we want them to look in Scratch, while also allowing overrides based on the GUI color mode.

### DIFF
--- a/src/css.ts
+++ b/src/css.ts
@@ -682,6 +682,14 @@ const styles = `
     opacity: 1;
   }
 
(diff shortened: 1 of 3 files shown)


cwillisf:
@gonfunko, any concerns with this approach?

PR #3465, opened 2026-03-10
fix: convert prototypes from shadow to regular block

### LINKED ISSUE #3450: [Bug] cannot drag argument reporter from prototype
## Expected Behavior

Can drag a argument reporter from prototype

## Actual Behavior

We can only drag procedures_definiton.

## Steps to Reproduce

1. 'Make a Block' and create a block with argument
2. Try dragging a argument reporter from prototype

## Additional Context
RaspberryPiFoundation/blockly#9538 make shadow block not selectable, which prevents dragging events that needed for argument's custom dragger to work. No ideas any other about monkey-patches since Gesture's behavior is hardcoded.

## System Details

Chrome 138, Windows 11. Happens when blockly upgraded to 12.4.1 (see additional context for more details)

cwillisf:
### Resolves

- Resolves scratchfoundation/scratch-blocks#3450

### Proposed Changes

Stop using shadows for procedure prototype blocks or their argument reporters.

### Reason for Changes

This allows argument reporters to be draggable in the context of RaspberryPiFoundation/blockly#9538. The remaining changes deal with the fallout from this change: we still want the prototype to use the secondary (shadow-esque) color, we still want the arguments to be non-deletable, etc.

### DIFF
--- a/src/blocks/procedures.ts
+++ b/src/blocks/procedures.ts
@@ -44,6 +44,41 @@ enum ArgumentType {
 }
 
 /**
+ * A drag strategy for the procedures_prototype block that delegates all drag
+ * operations to its parent (the procedures_definition block). This lets the
+ * prototype act as a non-shadow block (so its argument-reporter children can
+ * also be non-shadow and therefore clickable/draggable) while still causing
+ * the entire definition to move when the prototype area is dragged.
+ */
+class DelegateToParentDraggable implements Blockly.IDraggable {
+  constructor(private block: Blockly.BlockSvg) {}
+
+  isMovable(): boolean {
+    return this.block.getParent()?.isMovable() ?? false
+  }
+
+  startDrag(e: PointerEvent) {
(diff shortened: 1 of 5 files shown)


cwillisf:
@gonfunko, what do you think? This got more complicated than I expected, but it seems to work in my local testing.

gonfunko:
This broadly looks reasonable, though I agree it's more complicated than would be ideal. I think if drags on a non-movable child block sort of bubbled through to its parent it might be possible to simplify a lot of this? Not sure how the rest of the team feels but that feels sort of reasonable to me, I'll check in with them. There's also some changes coming in v13 that will simplify the drag strategies (you can just return a different block to use in place of the one the drag started on).

cwillisf:
Sounds good! I'll go with this for now, and maybe we can remove `DelegateToParentDraggable` with Blockly v13, or maybe earlier if the drag bubbling happens.

PR #3469, opened 2026-03-11
fix: be more cautious about workspace scrolling on block deletion

### LINKED ISSUE #3464: Deleting blocks scrolls to the top left
## Expected Behavior

Deleting a block doesn't affect the scroll position

## Actual Behavior

It jumps to the top left corner

## Steps to Reproduce

1. Go to 'https://scratchfoundation.github.io/scratch-editor/develop/scratch-gui/index.html'
2. Open or create a project that's large enough so that you can scroll in the block workspace
3. Scroll to the bottom right corner
4. Drag and drop a block onto the palette to delete it
5. It jumps to the top left corner

## Screenshots

https://github.com/user-attachments/assets/519d142b-0647-419e-95f2-dc83d71e930f

## System Details

Windows 10, Chrome 132

cwillisf:
### Resolves

- Resolves scratchfoundation/scratch-blocks#3464

### Proposed Changes

Replace `getRestoredFocusableNode` with a version that defaults to returning `null`, rather than resetting to the first block, if there's no good answer.

### Reason for Changes

This was causing unwanted scrolling; see scratchfoundation/scratch-blocks#3464

@gonfunko, I feel like Blockly might want a more careful version of this, or (maybe more likely) there's something I'm missing about this being intentional or desirable, or (maybe most likely) there's some other set of things conspiring to make this a Scratch-specific issue.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -157,3 +157,18 @@ contextMenuItems.registerDuplicateBlock()
 Blockly.ContextMenuRegistry.registry.unregister('workspaceDelete')
 contextMenuItems.registerDeleteAll()
 Blockly.comments.CommentView.defaultCommentSize = new Blockly.utils.Size(200, 200)
+
+// When the focused block is deleted and has no parent or nearby neighbor,
+// Blockly falls back to focusing the first/topmost block in the workspace,
+// which triggers a scroll to that block. In Scratch, focus should fall back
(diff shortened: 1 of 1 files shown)


gonfunko:
I addressed this in https://github.com/RaspberryPiFoundation/blockly/pull/9599, which should be in the next Blockly point release.

PR #3549, opened 2026-04-16
fix: allow C-blocks without a next connection (e.g. forever) to wrap around mid-stack blocks

### LINKED ISSUE #3494: Cannot wrap a C-Block over a block inside an existing stack
## Expected Behavior
I should be able to wrap a C-Block over a block already in a stack.

## Actual Behavior
It doesn't work. I have to break the block from the stack, wrap it outside, and reconnect everything.

## Steps to Reproduce

1. [Go to the GitHub Pages build.](https://scratchfoundation.github.io/scratch-editor/scratch-gui/)
2. Make a stack of blocks.
3. Try to wrap a C-Block on a block in the stack that isn't top-level.
4. Either no wrapping will occur, or it might try to connect itself.

## Screenshots
Fork:
https://github.com/user-attachments/assets/a6ebf577-fe34-4af8-8011-00ab307c0b9c

Spork:
https://github.com/user-attachments/assets/b7273bb5-002d-47f9-89c4-c62ec7a85615

## System Details
Zen Browser 1.19.3b (based on Firefox 148.0.2), MacOS Tahoe 26.3.1

The bug also occurs on Google Chrome 146.0.7680.154

cwillisf:
### Resolves

- Fixes scratchfoundation/scratch-blocks#3494 more correctly :sweat_smile:

### Proposed Changes

Two fixes for wrapping a "forever" block (C-block with no bottom connector) around blocks mid-stack:

#### Enable the wrap insertion preview during drag

Blockly's base `ConnectionChecker.doDragChecks` rejects mid-stack insertion for blocks without a nextConnection, assuming displaced blocks have nowhere to go. This override in ScratchConnectionChecker checks whether `getConnectionForOrphanedConnection` can route the orphan into a statement input instead. Without this, dragging a forever block over the middle of a stack never offered a wrap preview.

#### Prevent data loss during insertion marker cleanup

The `hideInsertionMarker` patch from #3498 assumed the marker always has a `nextConnection`, so it could move the displaced block there for `unplug(true)` to heal. Forever blocks have no `nextConnection`, so the condition short-circuited and the displaced block was disposed with the marker. Now, when the marker has no `nextConnection`, we reconnect the displaced block directly to the connection the marker was plugged into (using `targetConnection`, which handles both next-connection chains and statement inputs).

### Test Coverage

- Connection checker allows mid-stack insertion for C-blocks with statement inputs
- Connection checker rejects insertion when target is an insertion marker
- Insertion marker cleanup restores displaced blocks for blocks with no bottom connector
- Insertion marker cleanup works when the marker is inside a statement input (not just a next-connection chain)

### DIFF
--- a/src/scratch_c_block_wrap.ts
+++ b/src/scratch_c_block_wrap.ts
@@ -81,9 +81,10 @@ Reflect.set(
     // Restore a real block from a statement input to nextConnection so that
     // unplug(true) can heal the stack correctly.  Conditions:
     //   - marker is mid-stack (has a predecessor)
-    //   - marker.nextConnection is empty (displaced block is NOT already there)
+    //   - marker.nextConnection is either empty or absent (cap blocks do not
+    //     have a bottom connector, so the displaced block is NOT already there)
     //   - a non-marker block is sitting in a statement input
-    if (markerPreviousConnection?.isConnected() && markerNextConnection && !markerNextConnection.isConnected()) {
+    if (markerPreviousConnection?.isConnected() && !markerNextConnection?.isConnected()) {
       for (const input of marker.inputList) {
         const conn = input.connection
         const connType = conn?.type as Blockly.ConnectionType | undefined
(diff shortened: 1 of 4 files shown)


cwillisf:
Please note: this doesn't seem to work correctly as it is. Will pick this up again later.

PR #3568, opened 2026-04-24
Fix: clear angle field event wrappers to prevent double-unbind

Joeclinton1:
### Resolves

- scratchfoundation/scratch-editor#537 — "Workspace Update Error" after using the angle picker, which then leaves blocks stuck and duplicated across sprites

### Proposed Changes

- In `src/fields/scratch_field_angle.ts`, null out `mouseDownWrapper_`, `mouseUpWrapper`, and `mouseMoveWrapper` right after `Blockly.browserEvents.unbind`, in both `onMouseUp` and `dispose`

### Reason for Changes

After dragging the angle picker, `onMouseUp` unbinds the move/up wrappers but leaves the references populated. When the field is later disposed (or mouseup fires again), `dispose` sees those references and unbinds them a second time — which blows up inside Blockly and bubbles out as a Workspace Update Error. Blockly then aborts the pending change and the workspace is left in the weird duplicated/stuck state from #537.

Setting the wrapper to `undefined` after unbind is what core Blockly does everywhere else for this pattern, and makes the handlers safe to call twice.

### Test Coverage

- Added `tests/unit/scratch_field_angle.test.ts`: after `onMouseUp`, calling `dispose` must not re-unbind the drag wrappers. Spies on `Blockly.browserEvents.unbind` and asserts it's called twice total, not four times.
- `npm test` and lint pass locally.

### DIFF
--- a/src/fields/scratch_field_angle.ts
+++ b/src/fields/scratch_field_angle.ts
@@ -140,12 +140,15 @@ class ScratchFieldAngle extends Blockly.FieldNumber {
     this.gauge = undefined
     if (this.mouseDownWrapper_) {
       Blockly.browserEvents.unbind(this.mouseDownWrapper_)
+      this.mouseDownWrapper_ = undefined
     }
     if (this.mouseUpWrapper) {
       Blockly.browserEvents.unbind(this.mouseUpWrapper)
+      this.mouseUpWrapper = undefined
     }
     if (this.mouseMoveWrapper) {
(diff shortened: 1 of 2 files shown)


Joeclinton1:

Joeclinton1:
recheck

cwillisf (approved):
Well, it looks obvious now that you've found it 😅
Thanks!

PR #3571, opened 2026-04-24
fix: preserve stack below when deleting a top stack block

Joeclinton1:
### Resolves

- scratchfoundation/scratch-blocks#3655 (reported by @Joeclinton1 / moved by @cwillisf)
  - From @cwillisf: I edited the line above when I created scratchfoundation/scratch-blocks#3655 as a sub-issue of scratchfoundation/scratch-editor#533

Deleting a hat block (or any top-of-stack block) via right-click → Delete Block also deletes every block connected below it. Keyboard Delete/Backspace has the same bug. Expected: only the top block is removed; the stack below stays.

### Proposed Changes

- Extract a shared `deleteBlock` helper in `context_menu_items.ts`. For a top stack block (no `previousConnection` target), disconnect `nextConnection` before `dispose` so the block below is left as a new top block.
- Route `BlockSvg.prototype.checkAndDelete` through the same helper in `index.ts` so keyboard delete matches the context-menu path.

### Reason for Changes

`dispose(healStack=true)` on a top block has nothing to heal to, so the next block gets disposed with it. Explicitly detaching fixes both paths through one helper.

### Test Coverage

Two unit tests in `tests/unit/context_menu_items.test.ts` covering both delete paths. Both fail without the fix.

Manually verified in scratch-editor (right-click + keyboard) that deleting a hat leaves its stack intact, mid-stack delete still heals above↔below, reporter delete is unaffected, and undo restores in one step.

### DIFF
--- a/src/context_menu_items.ts
+++ b/src/context_menu_items.ts
@@ -45,6 +45,7 @@ export function registerDeleteBlock() {
 
 export function deleteBlock(block: Blockly.Block) {
   if (block.workspace.isFlyout) return
+  if (!block.isDeletable() || block.isShadow()) return
 
   const priorGroup = Blockly.Events.getGroup()
   const shouldStartGroup = !priorGroup
--- a/src/context_menu_items.ts
+++ b/src/context_menu_items.ts
@@ -47,7 +47,10 @@ export function deleteBlock(block: Blockly.Block) {
   if (block.workspace.isFlyout) return
 
   const priorGroup = Blockly.Events.getGroup()
-  Blockly.Events.setGroup(true)
+  const shouldStartGroup = !priorGroup
+  if (shouldStartGroup) {
+    Blockly.Events.setGroup(true)
+  }
   try {
     if (!block.outputConnection && !block.previousConnection?.isConnected() && block.nextConnection?.isConnected()) {
       block.nextConnection.disconnect()
@@ -61,7 +64,9 @@ export function deleteBlock(block: Blockly.Block) {
       block.dispose(!block.outputConnection)
     }
   } finally {
-    Blockly.Events.setGroup(priorGroup)
+    if (shouldStartGroup) {
+      Blockly.Events.setGroup(false)
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


Joeclinton1:
This was another GPT 5.5 fast test in codex. It took 12 minutes from prompt to pr. This time I tested it better to be sure it actually fixed it, as the previous with the shadow block duplication I messed up the repro, and thought it had fixed it when it hadn't.

cwillisf (approved):
I like how careful `deleteBlock` is. Thanks for the help!

cwillisf (changes requested):
OK, the Copilot review has some good points...

cwillisf:
(I recommend rebasing on or merging from `develop` to get the changes from scratchfoundation/scratch-blocks#3574 here)

Joeclinton1:
> (I recommend rebasing on or merging from `develop` to get the changes from #3574 here)

Sure let me see what I can do, I will get a second version out soon.

Joeclinton1:
@cwillisf I tested the proposed bugs by the reviewer and they don't exist. So I can do these changes but you can't reach the proposed bugs. 

Also you can delete a custom block with delete blocks even though it has usages which is a new bug, and not caused by this change (i tested in production).

Joeclinton1:
I did the changes related to `block.nextConnection?.disconnect()` but the rest just isn't valid feedback. I cannot actually preproduce the lone stack block not being deletable bug, but to be safe I chose to implement it.

non-deletable blocks can't test because, the only non deletable is the procedure definition and that's currently deletable in production which is an upstream bug. Shadow block inputs weren't deletable with keyboard, maybe because you can never focus them in such a way that would let that be possible

The undo group thing can't replicate so didn't fix.

mxmou:
> The undo group thing can't replicate so didn't fix.

It can cause issues if someone writes code like this in the future:

```ts
Blockly.Events.setGroup(true)
// do things that create events ...
deleteBlock(block)
Blockly.Events.setGroup(false)
```
The other events should be grouped together with the deletion, but they won't be because `deleteBlock()` creates a new group unconditionally.

Joeclinton1:
I pushed changes to address the other two reviewer feedbacks. They are not accessible through the ui, but it is possible that by not guarding against them it leaves the possibility of a bug later on in the codebase development. So even though it seems a little overly defensive I've done those changes.

PR #3572, opened 2026-04-26
mobile?

rosics-code:
### Proposed Changes

Adds a sensing block for Scratchers to detect mobile for mobile-friendly games. This would make it easier.

### Reason for Changes

So mobile-friendly detectors/games have a block for it.

see: https://github.com/scratchfoundation/scratch-editor/pull/546


rosics-code:
@cwillisf what do you think?

